PR #154, opened 2015-06-23
Add Connector article: Create Test Active Directory Domain Controller

twistedstream:
These steps were originally for a customer, but seemed worth adding into docs for others to use.


ntotten:
Looks great! I'll let @sandrinodimattia accept the PR after he reviews the content

sandrinodimattia:
@twistedstream could you upload the images to the CDN first and change the links?

sandrinodimattia:
@twistedstream could you take the browser screenshots as described in our internal screenshot guideline.
As for the 'Install and configure the AD/LDAP Connector' part you should reference the current documentation for installing the connector to avoid duplicate content.

twistedstream:
@sandrinodimattia fixed the screen shots.

Regarding you comment:

> As for the 'Install and configure the AD/LDAP Connector' part you should reference the current documentation for installing the connector to avoid duplicate content.

So I do point them to the instructions for setting up the new AD/LDAP connection in Auth0 on [this line of the MD code](https://github.com/auth0/docs/pull/154/files#diff-3ceae2d158d890f3e5f4fde6c1441305R78).  But I don't instruct them to follow the remaining instructions on docs (for installing the Connector in Windows) because if they paste the Ticket URL into their browser instead (which is what I instruct them to do), they essentially get the same instructions right?  Plus those should really be opened in a browser on the target machine so they can just click the link to download and install the MSI.

twistedstream:
@sandrinodimattia FYI cleaned up the instructions in that section so they blend a little better with those coming from **Ticket URL**

PR #161, opened 2015-06-25
Update custom signup to API v2

rolodato:
Still needs an example for redirect mode.


sandrinodimattia:
OK if we merge this when we have the redirect mode example?

rolodato:
I think we can merge this now and add redirect mode later since they are quite similar. We should discourage customers from doing the "old" way of doing custom signups.

PR #177, opened 2015-07-13
Update rails.md

peterholcomb:
Was getting 

NoMethodError (undefined method `[]' for nil:NilClass):
  app/controllers/api/v2/secured_controller.rb:21:in`validate_token'

When changed decoded_token[0]["aud"] to decoded_token["aud"], the exception went away.


benschwarz:
Hey @peterholcomb, it looks like your token (`decoded_token`) was set to `nil`, which is a bit strange:

`JWT.decode` should return array, with two entries — `[payload, header]`, if the token is invalid, it'd throw an exception. 

Either way, if you'd like to contact [support](https://auth0.com/support), we'd be super happy to run through the troubles that you were having. Thanks for the PR.

PR #178, opened 2015-07-14
Extracted snippets from tutorials

ntotten:


articles/aws-api-setup.md
```diff
@@ -19,7 +19,7 @@ This is a one time setup
   ![](/media/articles/aws-api-setup/aws-api-setup-2.png)
 
 <% if (account.userName) { %>
-3. Download the metadata from <a href="https://@@account.namespace@@/samlp/metadata/@@account.clientId@@" target="_new">https://@@account.namespace@@/samlp/metadata/@@account.clientId@@</a> and upload it here.
+3. Download the metadata from <a href="https://@@account.namespace@@/samlp/metadata/@@account.clientId@@">https://@@account.namespace@@/samlp/metadata/@@account.clientId@@</a> and upload it here.
```
mgonto, line 22: Why removing the target=_new?

articles/client-platforms/angularjs.md
```diff
@@ -6,9 +6,9 @@ image: //auth0.com/lib/platforms-collection/img/angular.png
 tags:
   - quickstart
 snippets:
-  dependancies: angularjs/dependancies.html
-  setup: angularjs/app.js
-  use: angularjs/login.js
+  dependancies: client-platforms/angularjs/dependancies
```
mgonto, line 9: We agreed not moving them into `client-platforms`, but actually to technology name / snippet. Why the change @ntotten ?

articles/custom-authentication.md
```diff
@@ -15,21 +15,21 @@ In order to install the __Auth0 Custom Authentication Connector__ you need to fi
...
     <li>
-      <a href="http://nodejs.org/dist/v0.8.22/node-v0.8.22-x86.msi" target="_blank">
+      <a href="http://nodejs.org/dist/v0.8.22/node-v0.8.22-x86.msi">
         <img src="//cdn.auth0.com/docs/img/node-windows.png" alt="">
         Windows Installer
         <small>node-v0.8.22-x86.msi</small>
       </a>
     </li>
     <li>
-      <a href="http://nodejs.org/dist/v0.8.22/node-v0.8.22.pkg" target="_blank">
```
mgonto, line 25: why removing all of these?

articles/native-platforms/cordova.md
```diff
@@ -43,13 +47,13 @@ tags:
...
 <div class="setup-callback">
-<p>Go to the <a href="@@uiAppSettingsURL@@" target="_new">Application Settings</a> section in the Auth0 dashboard and make sure that <b>Allowed Callback URLs</b> contains the following value:</p>
+<p>Go to the <a href="@@uiAppSettingsURL@@">Application Settings</a> section in the Auth0 dashboard and make sure that <b>Allowed Callback URLs</b> contains the following value:</p>
 
 <pre><code>https://@@account.namespace@@/mobile</pre></code>
 
 <p>Also, if you are testing your application locally, make sure to add your local URL as an Allowed Callback URL and the following as an Allowed Origin (CORS):</p>
 
-<pre><code>file://*</code></pre>
+<pre><code>file://\*</code></pre>
```
mgonto, line 56: Why this escape?

articles/aws-api-setup.md
ntotten, line 22: I prefer to let users decide where they want to open links and think generally it is considered bad practice to use target=new. It is better to let the user decide. Additionally, it was applied highly inconsistently in docs. Many places don't have it, particularly since most links are written in markdown. 

Later I am going to go and convert all links that are in <a> format to markdown when possible.

articles/client-platforms/angularjs.md
ntotten, line 9: The issue was that I realized there was name collision in many cases. i.e. server-platforms and server-apis both have node, java, etc. In light of that I thought it best to be consistent with all platforms.

Alternatively, it could have been something like nodejs-api, but it seemed like it didnt make sense to create a new pattern when we already have one.

articles/custom-authentication.md
ntotten, line 25: same as _new

articles/native-platforms/cordova.md
ntotten, line 56: Because without it the atom markdown editor doesnt render styles correctly. It doesnt affect the output.

ntotten:
@mgonto in addition to the style notes, did you check the actual code files? I just want to make sure that they make sense for your pattern of dependencies, setup, and use.

articles/aws-api-setup.md
mgonto, line 22: Hmm. I prefer the other way :).

I'd like all links to ALWAYS open in another tab. What we want is for users to always continue reading Auth0 and then follow up on the other links and come back to Auth0. If we go to the other link in the same page, we might be loosing a lot of people.

What do you think?

articles/client-platforms/angularjs.md
mgonto, line 9: Makes sense.

articles/native-platforms/phonegap.md
```diff
@@ -6,6 +6,10 @@ hybrid: true
 image: //auth0.com/lib/platforms-collection/img/phonegap.png
 tags:
   - quickstart
+snippets:
+  dependancies: native-platforms/phonegap/dependancies
+  setup: native-platforms/jquery/setup
```
mgonto, line 11: I don't know if this was you or the elance person, but this is exactly right :).

articles/native-platforms/windows-store-csharp.md
```diff
@@ -39,22 +42,7 @@ There are three options to do the integration:
 To start with, we'd recommend using the __Login Widget__. Here is a snippet of code to copy & paste on your project.
 Since we are using `await` (.NET 4.5 or greater), your method needs to be `async`:
 
-```cs
-using Auth0.SDK;
-
-var auth0 = new Auth0Client(
-	"@@account.namespace@@",
```
mgonto, line 46: Let's separate this into 2 snippets, one on top of the other. The first 3 lines are the setup. Then it's the use. Makes sense? @ntotten

articles/native-platforms/wpf-winforms.md
```diff
@@ -39,22 +42,7 @@ There are three options to do the integration:
...
-var user = await auth0.LoginAsync(this);
-/*
-    Use this object to do wonderful things, e.g.:
-      - get user email => user.Profile["email"].ToString()
-      - get facebook/google/twitter/etc access token => user.Profile["identities"][0]["access_token"]
-      - get Windows Azure AD groups => user.Profile["groups"]
-      - etc.
-*/
-```
+@@snippet(meta.snippets.use)@@
```
mgonto: Let's separate this into 2 snippets, one on top of the other. The first 3 lines are the setup. Then it's the use. Makes sense? @ntotten

articles/native-platforms/xamarin.md
```diff
@@ -44,22 +43,7 @@ There are three options to do the integration:
...
-// 'this' could be a Context object (Android) or UIViewController, UIView, UIBarButtonItem (iOS)
-var user = await auth0.LoginAsync(this);
-/*
-- get user email => user.Profile["email"].ToString()
-- get facebook/google/twitter/etc access token => user.Profile["identities"][0]["access_token"]
-- get Windows Azure AD groups => user.Profile["groups"]
-- etc.
-*/
-```
+@@snippet(meta.snippets.use)@@
```
mgonto: Let's separate this into 2 snippets, one on top of the other. The first 3 lines are the setup. Then it's the use. Makes sense? @ntotten

mgonto:
The rest looks good.

articles/aws-api-setup.md
ntotten, line 22: Your preference is that they open in another tab, but that isn't universally agreed upon. The nice thing about NOT putting target=blank is that you can use CTRL+Click and get a new window and other people who don't like this can use regular click. I am strongly in favor of letting the user choose what happens in their browser.

You can't trick people to stay on a certain page. We aren't going lose people because we don't open in a new window. Everyone knows how the back button works. If somebody is invested in a tutorial they are going to return to it if they accidentally open a link in the same window (which by the way is the default expectation of any link you open on the internet).

You also have to consider the mobile browsing experience where tabs are not nearly as easy as back. If you go to news.google.com and open any article it opens in a new tab. If you do this on your mobile device you will shortly become infuriated because you read an article and cannot use the back button. The correct UX for google news would be to let the user chose what window the article opened in. I would choose to use the same page then click back when I am ready to read more on google news.

http://www.smashingmagazine.com/2008/07/01/should-links-open-in-new-windows/

articles/native-platforms/phonegap.md
ntotten, line 11: :+1:

articles/native-platforms/windows-store-csharp.md
ntotten, line 46: Will do.

articles/native-platforms/wpf-winforms.md
ntotten: Will do.

articles/native-platforms/xamarin.md
ntotten: Will do

articles/native-platforms/wpf-winforms.md
eugeniop: FWIW, I hate when I click on a link and opens somewhere else. If I want a new window/tab, I just use CMD+click. But I guess this is one of those things people disagree on like when judging Judas Priest vs. AC/DC. Or Picasso vs Modigliani.

mgonto:
Can we merge this?

mgonto:
@ntotten :point_up:

articles/native-platforms/windows-store-csharp.md
ntotten, line 46: done.

articles/native-platforms/wpf-winforms.md
ntotten: done

articles/native-platforms/xamarin.md
ntotten: done

PR #181, opened 2015-07-16
Create style_guide.md

nancymic2:
blank style guide doc


ntotten:
to keep with the convention, let's name this STYLEGUIDE.md

PR #184, opened 2015-07-22
Edited articles/api/v1/reference.md

Catografix:
Made edits, committed and pushed this one file (articles/api/v1/reference.md) but this pull request insists on including changes to 32 files from the articles/libraries folder that i made no changes to. Please advise.


ntotten:
@Catografix When you are pulling from the remote branch, could you do a `git pull --rebase upstream master` rather than `git pull upstream master` If you pull with rebase it will layer your changes on top of each other rather than doing all of those merge commits.

ntotten:
@Catografix Also, this looks like it has a bunch of deleted files.

And leave the .gitattributes out of this commit. If we need to add it we should do it separately.

PR #213, opened 2015-07-31
Issue 212: Updates to User Profile

### LINKED ISSUE #212: Updates to User Profile
This feature includes a pretty big facelift to all the content that discusses the **User Profile** as well as related topics like tokens.  At the moment, the best docs page we have that covers this topic is [Auth0 Normalized User Profile](https://auth0.com/docs/user-profile), which will also be enhanced in this feature and augmented with a handful of new pages.

twistedstream:
Starting a PR to collaborate on the updates to User Profile described in #212, with a special heads-up to @yvonnewilson and @eugeniop.

Closes #212


articles/tokens.md
```diff
@@ -0,0 +1,484 @@
...
+
+https://auth0.com/docs/libraries/auth0js#12
+
+###Best Practices
+For granularity of access control, good to set up different applications/APIs with different secrets, so that a delegation token can be issued uniquely for each application/API.
+
+Tokens should be issued with a short timeframe for expiration, where configurable.
+
+
+## The Auth0 APIv2 access token
```
twistedstream: @yvonnewilson, I found in [this section](https://auth0.com/docs/api/v2/changes#7) of the **API v1 vs v2** docs page that it calls this token the `api_jwt_token`.  What's nice about this term is that it gives the Auth0 APIv2 access token a similar moniker as the other tokens like `id_token` and `access_token`.  Should we consider using this in our docs?

yvonnewilson: @twistedstream Hi, I like the 'api' part of the name, but not sure we need the 'jwt' in it.  None of the other names allude to internal format details of the token.  I think it would be more helpful to clarify in the name which api it is for because there are several apis, each of which requires a different token.  It would eliminate a lot of confusion if
a) we use the same name (fully qualified) consistently throughout our docs for each type of token (e.g. don't just call it 'the token')
b) each token is named to indicate what it is and/or what you use it for. - auth0 api token, IDP api token, 3rd party app API token, etc.  
I'm not strongly against having jwt in the name, but I think the above 2 things are more important in helping people understand all these tokens and which ones they need.

ntotten:
@twistedstream Whats the status on this?

twistedstream:
@ntotten This PR is ready for review and merge. 

It includes new User Profile content as well as a new section that describes the various types of tokens (JWT, access_token, refresh_token, etc) used by Auth0, with a new entry in `config/navigation.yml` so Tokens shows up in the **How To's** section. It's not perfect, but it's a good start. @yvonnewilson and @eugeniop wanted to get this in sooner than later.

We'd like to leave the branch open so we can continue to revise content and add things like illustrations.

Issue #193, opened 2015-07-24
Windows Phone Native tutorial does not specify how to handle authentication cancell


iannyanes:
The tutorial on https://auth0.com/docs/quickstart/native-mobile/windowsphone/no-api does not specify how to handle the case when authentication is cancelled by the user. 

The following piece of code is recomended

try{
var user = await App.Auth0.LoginAsync();
}
catch (AuthenticationCancelException)
{
// Handle case wen user cancelled authentication
}

xperiandri:
Please, merge https://github.com/auth0/Auth0.WindowsPhone/pull/18 in order to close this issue and keep documentation in sync with sample in the repository.

PR #219, opened 2015-08-04
Techaid patch for Windows Phone tutorial

xperiandri:
Modified Windows Phone tutorial so user can simpler get started according to issue #193


ntotten:
Closes #193

PR #224, opened 2015-08-06
Appliance edits (3 articles)

Catografix:
Edits to checksum.md

Line 27 refers to an xxxx in the preceding code sample, which does not contain any such variable.

### DIFF
--- a/articles/appliance/checksum.md
+++ b/articles/appliance/checksum.md
@@ -1,4 +1,4 @@
-# Verify the integrity of an update package
+# Verify Update Package Integrity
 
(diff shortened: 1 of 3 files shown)


Catografix:
Note that this PR contains 2 articles, 
I forgot to create a new branch before beginning on the second file.

Catografix:
This PR now contains 3 articles. Even though I edited the new article in a new branch, when I pushed just the previous branch with new changes to the second article, the third article was included. I think my mistake was that I added the third article while on the first branch. I will know to avoid this in the future.

Catografix:
I have broken this PR into 3 separate requests, one for each file, and will close this one.

Issue #244, opened 2015-08-14
(iOS - Objective-C) JWTDecode dependency version must be set to 0.2


Division:
JWTDecode 0.3.1 is written in Swift - which cause cocoa pods to show error that asks adding use_frameworks! to the Podfile. XCode build fails until addition configuration is done for Lock-Facebook target.
The better and user friendly approach is to use JWTDecode version 0.2 which is written in Objective-C and doesn’t need any additional setup.

Issue #259, opened 2015-08-14
(iOS Swift) "Let's implement the login" sample code must be fixed


Division:
Sample code is not correct:
`controller.onAuthenticationBlock = {(profile:A0UserProfile!, token:A0Token!) -> () in`
must be changed to 
`controller.onAuthenticationBlock = { (profile, token) -> () in`

Issue #257, opened 2015-08-14
(iOS Swift) Add workaround to fix build error after Lock-Facebook added


Division:
After adding Lock-Facebook dependency project build starts failing:
![image](https://cloud.githubusercontent.com/assets/320958/9268942/bb04e1be-426c-11e5-9f4f-eb87bf74e58e.png)

Workaround to fix it: set property  “Allow non-modular includes in Framework Module” of the Lock-Facebook target to Yes:

Issue #256, opened 2015-08-14
(iOS Swift) Facebook key must be fixed


Division:
Key must be changed to FacebookAppID (last letter uppercase)

PR #260, opened 2015-08-14
Techaid patch iOS Objective-C and Swift

Division:
Fixed issues:
https://github.com/auth0/docs/issues/244
https://github.com/auth0/docs/issues/246
https://github.com/auth0/docs/issues/247
https://github.com/auth0/docs/issues/248
https://github.com/auth0/docs/issues/249
https://github.com/auth0/docs/issues/250
https://github.com/auth0/docs/issues/251
https://github.com/auth0/docs/issues/252
https://github.com/auth0/docs/issues/253
https://github.com/auth0/docs/issues/254
https://github.com/auth0/docs/issues/255
https://github.com/auth0/docs/issues/256
https://github.com/auth0/docs/issues/257
https://github.com/auth0/docs/issues/259

### DIFF
--- a/articles/server-apis/aspnet-webapi.md
+++ b/articles/server-apis/aspnet-webapi.md
@@ -43,9 +43,10 @@ ${snippet(meta.snippets.dependencies)}
 
 Open the **WebApiConfig.cs** file located in the **App_Start** folder and add the following `using` statements:
 ```cs
-using Api.App_Start;
+using YOUR_WEBAPI_PROJECT_NAME.App_Start;
 using System.Web.Configuration;
 ```
+Remember to replace the placeholder for your project name in the first `using` statement.
 
 Add the following code snippet inside the `Register` method.
 


snippets/native-platforms/ios-objc/dependencies.md
```diff
@@ -1,4 +1,4 @@
 ```ruby
-pod 'Lock', '~> 1.12'
-pod 'JWTDecode', '~> 0.2'
+pod 'Lock', '~> 1.16'
+pod 'JWTDecode', '0.2'
```
hzalaz: I believe this is because `0.3.0` drops support for iOS 7. In this case let's specify the full number of the version which is `0.2.2` or even use `~> 0.2.2` if we ever release a new patch for `0.2`

snippets/native-platforms/ios-swift/use.md
```diff
@@ -2,7 +2,7 @@
 let lock = MyApplication.sharedInstance.lock
 let controller = lock.newLockViewController()
 controller.closable = true
-controller.onAuthenticationBlock = {(profile:A0UserProfile!, token:A0Token!) -> () in
+controller.onAuthenticationBlock = { (profile, token) -> () in
```
hzalaz: For the sake of removing unnecessary code I think you can remove the `-> ()` :smile:

articles/native-platforms/ios-swift.md
```diff
@@ -177,6 +177,10 @@ Then add Lock Facebook's Pod
 pod 'Lock-Facebook', '~> 2.0'
 ```
 
+Click on the `Pods` project and select `Lock-Facebook` target and set `Allow Non-modular Includes In Framework Modules` to Yes:
```
hzalaz: Not sure this is something I'd rather recommend cause there are cases this will break everything and we shouldn't change the `Pods` project. I know there is an [issue](https://developers.facebook.com/bugs/850078471740674) with FB SDK but haven't tried the solution they proposed.

Division: I've tried to add Bolts dependency but it didn't help. Shouldn't it be done within Lock-Facebook?
By the way they also proposed to set this property to YES:
image

Finally, should we remove this setting from the tutorial?

snippets/native-platforms/ios-swift/use.md
Division: Done :)

snippets/native-platforms/ios-objc/dependencies.md
Division: 3.0 is written in Swift which requires use_frameworks! setting at the Podfile. It breaks Lock-Facebook target compilation unless `Allow Non-modular Includes In Framework Modules` set to YES 

Switched to ~> 0.2.2

articles/native-platforms/ios-swift.md
ntotten: @hzalaz Thoughts? This would be a good change to get committed as it has lots of fixes.

PR #286, opened 2015-08-20
Fixed aliases

tehsis:
Fixed/added aliases in different articles


articles/addons/mscrm.md
```diff
@@ -5,4 +5,7 @@ title: Dynamics CRM
 logo_class: MicrosoftDynamicsCRM
 configRoute: /add-ons/mscrm
 sitemap: false
+alias:
+  - Dynamics CRM
```
cristiandouce: This is wrong @tehsis

PR #287, opened 2015-08-21
Added addons logos and fixed aliases

tehsis:
This PR Adds logos to addons (as well as it corresponding `image` attribute) and fixes some aliases.


cristiandouce:
@tehsis ++

PR #301, opened 2015-08-26
Update servicestack.md

petemidge:


ntotten:
This doesn't seem to be right. The code is already in there as the setup step.

PR #306, opened 2015-08-27
Changed authorizer in ember.js configuration

thijsvdanker:
When using the authorizer that is described in the docs I get the following error mesage:

```
Error while processing route: tracking-facts.index ajaxPrefilter.authorizer.authorize is not a function TypeError: ajaxPrefilter.authorizer.authorize is not a function
```

I have changed the authorizer to `simple-auth-authorizer:jwt` which seems to resemble the authorizer that is available in the package and the error has gone away.


mgonto:
Thanks for that!

PR #505, opened 2015-10-22
fixed link-accounts images again

Catografix:
See if you like these


Catografix:
The originals were fine

PR #506, opened 2015-10-22
Small edit to refresh integrations-template doc

Catografix:
There are still problems with these pages. The live page (and the one on master) is looking for the images in the wrong location. Somehow these changes got lost. Refreshing this doc should fix the first two links. If so, I will then repush the other docs as well.


Catografix:
Oh I see. The images are now in a folder named third-party-apps. I see you had to fix my code. I will pull down your changes to get the code fix, then update the urls.

Catografix:
I guess theres a better way to fix it.

PR #511, opened 2015-10-23
add redirect for api v2 tokens link

solepano:


solepano:
will change the approach

Issue #422, opened 2015-09-25
Updates to AWS integration pages.


chriskeyser:
https://auth0.com/docs/integrations/aws

SSO with dashboard
- [x] step 1 - have a link that shows a user how to do this. (screenshot?)
- [x] step 3 - be specific => paste this into settings area on the Settings tab.
  -> on the settings tab press "Save" (this is way off screen and not obvious)
  f
- [x] step 4 - numbering restarts to 1? There is a lot to this that users may not get:
  Configure Auth0 as the IdP for AWS. AWS will require importing the IdP Metadata. Scroll down on the same configuration screen on the Auth0 dashboard. Look for the Identity Provider Metadata link.
- [x]  a. First at least point to the AWS documentation on setting up STS for SAML. Docs are here: http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_providers_create_saml.html
- [x]  b. The metadata download is not so obvious and the instructions are no longer correct. You need to do the following in auth0
  -> click on the "Usage" tab
  -> Under the "identity provider Metadata" link, click on "download"
- [x] step 5 -Send AWS roles or write a Rule to map values to it, like this example:
  I'm not sure what "Send AWS roles" means. The role needs to be injected into the SAML token, I don't think it will work otherwise. The only way I am aware of to get the roles into the token is as outlined. I suggest just say specify an AWS role with a rule as follows.  Also have skipped the step of creating a role in AWS. The role has to be created in a specific way to allow its use to gain access to AWS. Here is the doc on AWS. http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_saml.html The steps are:
  1. from the IAM console, select create roles in the left menu, then select "create new role"
  2. Enter a role name and click next
  3. Select "Role for Identity Provider Access" and click next.
  4. Select "Grant Web Single Sign-On (WebSSO) access to SAML providers"
  5. Accept the defaults for the next screen (SAML:aud value of https://signin.aws.amazon.com/saml)
  6. Accept the Role Trust proposed. (This policy tells IAM to trust the Auth0 SAML IDP)
  7. Choose an appropriate access policy for this role. This defines the permissions that the user being granted with this role will have within AWS. For example, to just let users read information in the console, select the ReadOnlyAccess policy.
  8. Select "Create Role"
  - [x] I suggest having a step (Step 5) to setup the role, and Step 6 is 
  - Write a Rule to map the role to a user.
    Point to our docs on rules.
[...]

PR #521, opened 2015-10-29
Fixes for first half of #422

Catografix:


chriskeyser:
1. In step one, can you finish the sentence with "and perform the below configuration". The screenshot is after it has been configured so it is slightly confusing.
2. there is a hanging comma here after the bracket. It works but correct is to not have the comma. I think this was a carry over i missed.
   `"nameIdentifierProbes": [
   "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress",
   ],
   }`
3. step 4 - there is no option, should say Write a rule to map...This was a hang over from previous, but it is a quick update, can you make it?

Catografix:
yes ill do these now

Catografix:
What is the 'it' in " Write a Rule to map values to it"?
The step one screenshot is just of the enabling of the SAML2, it doesnt
show any configuration. Would you prefer a different screenshot?

chriskeyser:
Write a rule to make the token issued for the user to an AWS role…

Does that sound good?

Catografix:
I had not begun worked on step 4, (which was step 5 in the issue #422),
theres seems to be alot of changes needed here including an entire step on
creating the role in the IAM console.

Ill work on this now.

Catografix:
Ive worked through the rest of the items in the first section:
https://github.com/auth0/docs/commit/fd0e799f7bd868c0bf56a2b35fe20793c56e9dde

let me know what you think

Il work on the Delegation section next.

ntotten:
@chriskeyser Please confirm if this is ready to be merged.

chriskeyser:
John, really looking great, two minor issues on wrapping (I'm in chrome). See step 5.iii it looks like the image is on the line instead of following. Also see 7, it looks like the trailing line is on the same line as the image. Otherwise ready to go.

Catografix:
Hi Chris

Im glad you're ok with the content.

I fixed the image alignment in step 7 by placing it after the text.
.
In step 5.iii I've tried several things. The code is identical to that in
step 5.i which displays perfectly. Often the live site formats the markup
differently than the preview, so it is best to see it live before trying
too many alternatives.

_In the Delegation section:_
  "The detail here is nice but needs to be updated" - what in this doc is
this referring to?
  "Screens have changed on AWS" and "Role interface in AWS has also
changed" - Ive been taking new screenshots
  "Step 5 says Principal ARN" - where to place this and the exact wording?
  "It'd be nice to then show the code" - Is this the Client-side sample
code alredy present?

Catografix:
In step 5.iii, the text was longer than the image. The solution was to make this its own step.

Catografix:
The image alignment is fixed.

Ive taken a shot at the Delegation section changes. Ive made the additions
requested but I am unclear where to put them.

"A brief description of the AWS add-on..." needs to be written. I will take
a screenshot when I know if the SAML2 addon should remain selected.

PR #556, opened 2015-11-10
Fix for issue 338

aguerere:
Advice the reader to avoid setting a value to window.location.href
while the user is logged in. Otherwise Lock will loss track of
window.location.hash and the token is lost.

### DIFF
--- a/articles/client-platforms/vanillajs.md
+++ b/articles/client-platforms/vanillajs.md
@@ -137,3 +137,10 @@ window.location.href = "/";
 ### 7. All done!
 
 You have completed the implementation of Login and Signup with Auth0 and VanillaJS.
+
+
+### Troubleshooting
+
+#### Window.location.hash contains an empty string after the user is logged in
+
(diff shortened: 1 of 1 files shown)


articles/client-platforms/vanillajs.md
```diff
@@ -137,3 +137,10 @@ window.location.href = "/";
 ### 7. All done!
 
 You have completed the implementation of Login and Signup with Auth0 and VanillaJS.
+
+
+### Troubleshooting
+
+#### Window.location.hash contains an empty string after the user is logged in
```
ntotten: lowercase `window.location.hash` as it is code.

ntotten:
Rather than add this warning, is there a way we can just fix the code to not ever cause this?

ntotten:
ping

aguerere:
Hey @ntotten . Sorry for the delay.
A permanent fix for this would be to revise the tutorial and make it compatible with the latest seed project that does not make use of window.location.hash . Looks like that was an old technique implemented by the older login widget.

Should I close this PR then?

Issue #105, opened 2015-06-07
Windows Store (C#) tutorial links to deprecated version of login widget


jose-e-rodriguez:
Step 3 of Windows Store (C#) tutorial links to an old version of the login widget:
https://auth0.com/docs/libraries/login-widget2
It should be using the newer "Lock" widget.

![2015-06-07_1745](https://cloud.githubusercontent.com/assets/12707130/8026248/32c2168c-0d3d-11e5-95bb-8b8334934ccb.png)

![2015-06-07_1747](https://cloud.githubusercontent.com/assets/12707130/8026249/3d7215a0-0d3d-11e5-88c0-6cb7a342684b.png)

PR #559, opened 2015-11-11
https://github.com/auth0/docs/issues/105

rafaelnogueraTechAid:
Corrected the relative route with the absolute and actual route for the
login widget


articles/native-platforms/windows-store-csharp.md
```diff
@@ -40,7 +40,7 @@ ${snippet(meta.snippets.dependencies)}
 ### 3. Integration
 There are three options to do the integration:
 
-1. Using the [Auth0 Login Widget](/login-widget2) inside a Web View (this is the simplest with only a few lines of code required).
+1. Using the [Auth0 Login Widget](https://auth0.com/docs/libraries/lock) inside a Web View (this is the simplest with only a few lines of code required).
```
ntotten: Don't use absolute URLs. should be `/libraries/lock`

PR #560, opened 2015-11-11
Update index.md

Annyv2:
Updated the line Create authorization rules based on complex logic (anything that can be written with node.js), to be Create authorization rules based on complex logic (anything that can be written with JavaScript) taking this issue into consideration https://github.com/auth0/rules/issues/17

### DIFF
--- a/articles/rules/index.md
+++ b/articles/rules/index.md
@@ -13,7 +13,7 @@ An App initiates an authentication request to Auth0 (__Step 1__), Auth0 routes t
 Here are a few examples. You could:
 
 * Bring information from your own databases and add it to the user profile object.
-* Create authorization rules based on complex logic (anything that can be written with node.js).
+* Create authorization rules based on complex logic (anything that can be written with JavaScript).
 * Normalize attributes from different providers besides to what we provide out of the box.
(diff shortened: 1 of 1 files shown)


articles/rules/index.md
```diff
@@ -13,7 +13,7 @@ An App initiates an authentication request to Auth0 (__Step 1__), Auth0 routes t
 Here are a few examples. You could:
 
 * Bring information from your own databases and add it to the user profile object.
-* Create authorization rules based on complex logic (anything that can be written with node.js).
+* Create authorization rules based on complex logic (anything that can be written with JavaScript).
```
ntotten, line 16: Is this better? It seems that node.js is more accurate as rules do run on node.

Annyv2, line 16: I took this discussion into account https://github.com/auth0/rules/issues/17 it seem to be the conclussion they arrived to (considering @jfromaniello comment "I think there is an issue with that part of the documentation, because it is not node."), but if it's better as it was I can close this PR. Let me know

ntotten, line 16: ah yeah, good point.

PR #562, opened 2015-11-12
Overall edits, all new screenshots, new policy steps, fixes 179

### LINKED ISSUE #179: AWS API Setup does not match with AWS Portal
This doc is not updated since there is a new step asking for the role type.

https://auth0.com/docs/aws-api-setup it should be based on http://docs.aws.amazon.com/IAM/latest/UserGuide/roles-creatingrole-identityprovider.html

also, the steps are wrong numbered

Catografix:
fixes #179


articles/aws-api-setup.md
```diff
@@ -1,70 +1,104 @@
...
 
   ![](/media/articles/aws-api-setup/aws-api-setup-5.png)
 
-2. Click on **Create New Role** and give it an arbitrary name.
+2. Provide a name for your new role and click **Next Step**:
 
   ![](/media/articles/aws-api-setup/aws-api-setup-6.png)
 
-3. Select or enter the following values.
+3. Select **Role for Identity Provider Access**. Then select **Grant API access to SAML providers** and click **Next Step**:
```
Catografix, line 39: Is this the correct Role Type?

articles/aws-api-setup.md
```diff
@@ -1,70 +1,104 @@
...
 
   ![](/media/articles/aws-api-setup/aws-api-setup-12.png)
+
+## Copy the ARN values
+
+1. From the summary page of the role you just created, copy the **Role ARN** value. You will use this value later when calling the `/delegation` endpoint in Auth0.
+
+  ![](/media/articles/aws-api-setup/aws-api-setup-13.png)
+
+2. From the summary page of the identity provider you created previously, copy the **Provider ARN** value. You will use this value later when calling the `/delegation` endpoint in Auth0.
```
Catografix, line 100: Its called Provider ARN here on AWS, is it referred to as Principal ARN at the /delegation endpoint ?

articles/aws-api-setup.md
```diff
@@ -1,70 +1,104 @@
...
+
+  ![](/media/articles/aws-api-setup/aws-api-setup-14.png)
+
+**NOTE:** For more information on supported AWS APIs, see: [AWS Services That Work with IAM](http://docs.aws.amazon.com/IAM/latest/UserGuide/Using_SpecificProducts.html).
```
Catografix, line 104: Better place for this note earlier in the doc?

articles/aws-api-setup.md
```diff
@@ -1,70 +1,104 @@
...
+In this example, you will create a policy that grants full access to the S3 resource: `YOUR_BUCKET/<%= '${saml:sub}' %>`. This will be evaluated at run-time and replaced with the `user_id` of the logged-in user.
+
+1. On the **Roles** page, select the role you just created.
+
+  On the **Permissions** tab under **Inline Policies**, click the link to create an inline policy:
 
   ![](/media/articles/aws-api-setup/aws-api-setup-11.png)
 
-8. Copy the **Principal ARN** value. You will use this value later when calling the `/delegation` endpoint in Auth0.
+3. Name your policy and enter the following code in the **Policy Document** field:
```
Catografix: Is this code correct? I copied it from the old screenshot.

articles/aws-api-setup.md
ntotten, line 39: @chriskeyser can you verify this?

articles/aws-api-setup.md
ntotten, line 100: @chriskeyser check

articles/aws-api-setup.md
ntotten, line 104: @chriskeyser feedback?

articles/aws-api-setup.md
chriskeyser, line 39: First really sorry for taking so long to get to this.

This is fine.

articles/aws-api-setup.md
```diff
@@ -1,70 +1,104 @@
...
 
   ![](/media/articles/aws-api-setup/aws-api-setup-8.png)
 
-5. This is where you assign permissions to this role. In this example, we are creating a policy that give full access to the S3 resource `YOUR_BUCKET/<%= '${saml:sub}' %>`. This will be evaluated on runtime and replaced with the `user_id` of the logged in user.
+5. On the Verify Role Trust page, accept the Policy Document as provided. Click **Next Step**:
 
   ![](/media/articles/aws-api-setup/aws-api-setup-9.png)
 
-  > For more information about what AWS APIs are supported: <http://docs.aws.amazon.com/IAM/latest/UserGuide/Using_SpecificProducts.html>
+6.  Skip the **Attach Policy** step for now. Click **Next Step**.
```
chriskeyser, line 57: This is the point that I was talking about for the attach policy. Wasn't sure if this was a specific example, I didn't think it was, so I'd instead say....

At **Attach Policy** either select a pre-defined policy to attach or define a custom policy as defined in the next section. Click **Next Step**

articles/aws-api-setup.md
chriskeyser, line 100: That is the case. Ideally we would have called this Provider vs. Principal for naming, but since it is already in place, I was just going to fix this by documenting them as equivalent rather than changing the delegation code.

articles/aws-api-setup.md
chriskeyser, line 104: I wouldn't worry about it, I think its fine at the end. The coverage now is almost 100%, it was more important a year ago when it was less complete.

chriskeyser, line 104: We probably need a code example on usage, I'll put adding that on my backlog, I think I need to do that part.

Catografix:
Since this is a total rework, impossible to merge small edit (736e6da481e7cf231d6208a95eccb8a0249efee0) on old doc by suffic.
New PR #712

articles/aws-api-setup.md
chriskeyser, line 39: I previously went through all of the updates, did something change? It
looks correct to me but want to be sure if there is something specific that
I look at it.

articles/aws-api-setup.md
Catografix, line 104: @chriskeyser The comment above about a code example on usage. Is this no longer needed?
I had to move this to a new PR #712

chriskeyser, line 104: I see I think you want to close the PR. Don't worry about the example then
I think. We could use one, but I don't know when I'll get to it.

PR #585, opened 2015-11-16
Renamed files to match conventions

dafortune:
Renamed email wall images
Renamed email wall case-studies


ntotten:
Rename files to `jet-privilege.md`, `safari.md`, and `schneider-electric.md`, no need for dates or case study to be in the title as it is already in the url.

vctrfrnndz:
I added these changes and others with @ntotten 's suggestions on https://github.com/auth0/docs/pull/590

PR #599, opened 2015-11-19
Add a note about configuring spring-security-auth0 using Gradle.

whittle:
Official Spring docs typically include configurations for Maven and Gradle. For an example, see the Maven/Gradle switch dingus under Quick Start at http://projects.spring.io/spring-security/


ntotten:
Thanks for the contribution!

PR #638, opened 2015-12-11
Added Angular Landing Page

diegopoza:


mgonto:
For the next one please change commit message per @ntotten request.

PR #658, opened 2015-12-16
Angular 2 beta 0 support update

chenkie:


chenkie:
Updated angular2-jwt and samples for Angular 2 beta.0 support

Issue #669, opened 2015-12-21
Missing step in Ruby Web App


Annyv2:
Ruby on rails tutorial https://auth0.com/docs/quickstart/webapp/rails/ is missing a step, since it doesn't work if the file /session_store.rb, doesn't have this set up properly `Rails.application.config.session_store :cache_store`

PR #673, opened 2015-12-23
[update] Added clarification in rails.md

Annyv2:
[update] Updated tutorial to explain what to do if the user is unable to access the user profile and gets a NoMethodError, which is that this `Rails.application.config.session_store :cache_store` has to be added to the session_store file.


Annyv2:
Issue https://github.com/auth0/docs/issues/669

ntotten:
@Annyv2 @iannyanes Commit message is not in correct format.

Issue #20, opened 2015-04-29
Missing docs for "Improved brute force protection"


glena:
Right now, there is no info about this topic.

To answer the questions:
- how does it work? 
- what is the impact to the user whose "account" is trying to be breached?

> Auth0 implements algorithms to detect suspicious login attempts. When a suspicious activity is detected, Auth0 will block subsequent login attempts originating from the same computer. It will also send an email to the user notifying him containing link to re-enable the origin of the request. Notice that Auth0 never blocks the user itself, just the attempts from the suspicious origin.
- are all access from a suspicious ip blocked or is it the combination of username/ip that is blocked?

> It depends on the algorithm and might change based on the context and evolution of the algorithm

This is an example of the email that is sent to the user

```
Blocked Access Attempt

We detected suspicious activity from the ip: 190.18.158.85 from Buenos Aires, Argentina.

If that is your IP address click here to unblock it. Otherwise disregard this email.

Clarification: Your account is not blocked, only access to it from the aforementioned IP.


Thanks, 
Default App
```

It can be edited in the dashboad under Emails -> templates -> Blocked Account Email **We need to change the tempate name to other than _Blocked Account_, maybe _Blocked Login Attempt_?**

PR #684, opened 2015-12-29
Added docs for brute force protection

aguerere:
Fixes https://github.com/auth0/docs/issues/20


aguerere:
This new content should be linked from somewhere else in the docs so that it gets into context.

ntotten:
@aguerere Can you create a PR with the link? It should probably be on https://github.com/auth0/docs/blob/master/articles/connections/database/index.md

aguerere:
Sure, I will do that.

PR #790, opened 2016-01-27
Proposed edits from George Gecewicz: Example #3

ggwicz:


ggwicz:
There's a tricky edit to be made under the "Colon" sub-section here, as shown in this screenshot:

Screenshot

That phrase is tricky and reveals a conflict with a number of rules.

> Colons are useful when you are directly referring to an example, such as "here is the relevant code:"

Ending with the colon is not ideal. Having both colon and period inside the quote is not ideal either. And having the colon inside the quote with the period outside the quote is a no-go, too.

I would posit a rewrite so that it's something like this:

> Colons are useful when you are directly referring to an example. You might say "here is the relevant code:" before a block of code, for example.

Although that feels a bit clunky, too.

So something even better might be to say this:

> Colons are useful when you are directly referring to an example. An example of this can be seen below in the _Punctuation and Quotes_ section.

In that _Punctuation and Quotes_ section is the following text, a great example of the use of colons that is trying to be described:

> For example: "You must end quotations with periods inside the quotation marks," said the editor, "with only one exception."

PR #793, opened 2016-01-27
Proposed edits from George Gecewicz: Example #4

ggwicz:


ggwicz:
:warning: All of the substantive content additions and deletions here are by @khorne3!

The only things I submitted here are edits on the level of copyediting/line-editing.

Issue #809, opened 2016-01-29
apache tutorial incomplete


vmartynets:
If you configure apache server like in tutorial it give you an error that module can't verify token signature: 

![screenshot_8](https://cloud.githubusercontent.com/assets/1114365/12676304/c02e5d12-c699-11e5-87f1-2c8fe1878d2c.png)

iannyanes:
Solved

PR #842, opened 2016-02-07
[update] apache tutorial

vmartynets:
Using RS256 (and work, see https://github.com/auth0/docs/issues/809). Windows instructions on a2enmod.

### DIFF
--- a/articles/server-platforms/apache.md
+++ b/articles/server-platforms/apache.md
@@ -27,7 +27,7 @@ ${snippet(meta.snippets.dependencies)}
 
 ### 2. Configure the module with your Auth0 Account information
 
-Now, you should get a new configuration file under the `/apache2/mods-available` folder, where Apache modules are normally installed.
+Now, you should get a new configuration file under the `/etc/apache2/mods-available` folder, where Apache modules are normally installed (On Windows you need to use `/apache/conf/httpd.conf` file).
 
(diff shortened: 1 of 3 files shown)


articles/server-platforms/apache.md
```diff
@@ -17,34 +17,28 @@ snippets:
...
 
-Once you've installed it, you just need to enable it for Apache
+Once you've installed it, you just need to enable it for Apache (If you are using Windows, you can use [this](https://github.com/enderandpeter/win-a2enmod#installation) to get `a2enmod` working on your system)
 
 ${snippet(meta.snippets.dependencies)}
 
 ### 2. Configure the module with your Auth0 Account information
 
-Now, you should get a new configuration file under the `/etc/apache2/mods-available` folder, where Apache modules are normally installed.
+Now, you should get a new configuration file under the `/apache2/mods-available` folder, where Apache modules are normally installed.
```
ntotten: Does this path maybe depend on which version of linux/apache you are using? Why the difference?

vmartynets: I'm using Windows. To make it more platform independent.

ntotten: But if they aren't using Windows this will be confusing, no? We should probably explain that the path can be `/etc/` if they are on a different platform.

Issue #808, opened 2016-01-29
Asp.net owin backend dependency


vmartynets:
In step 1 it suggests you to update package 'System.IdentityModel.Tokens.Jwt' which is not installed by default. Better suggest to install this package.

![image](https://cloud.githubusercontent.com/assets/1114365/12675401/a3c7d290-c692-11e5-9fd4-6a542e0f7a35.png)

iannyanes:
Solved

PR #844, opened 2016-02-07
[update] asp.net webapi owin dependency install instead of update

vmartynets:
related to https://github.com/auth0/docs/issues/808


ntotten:
@jerriep Can you verify and merged if OK.

PR #845, opened 2016-02-07
[bug] fix for twitter 

vmartynets:
(and other connections that don't send 'state') log in. See https://github.com/auth0/auth0-scala/pull/5

### DIFF
--- a/articles/server-platforms/scala.md
+++ b/articles/server-platforms/scala.md
@@ -53,10 +53,9 @@ import play.api.mvc.Controller
 class Callback extends Controller {
 
   // callback route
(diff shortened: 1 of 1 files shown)


ntotten:
What exactly is the issue here? Why not send the state?

vmartynets:
As you can see [here](https://github.com/auth0/auth0-scala/issues/4), when trying to log in using twitter it gives you exception `No parameters supplied` which related to response. Response when using twitter contains only `code`, not `state` although response from google, for example contains `state` and `code`. When state is not in response variable `stateOpt` equals to `None` and this rises exception.

Issue #745, opened 2016-01-18
Add android instructions Cordova


Annyv2:
Cordova has no android instructions but the seed project works on android

Issue #683, opened 2015-12-29
Add Android instructions for Phonegap


Annyv2:
Phonegap tutorial only has instructions for iOS, but the seed project also works for Android so instructions can be added to explain this

PR #848, opened 2016-02-09
[edit] Android instructions for Phonegap and Cordova

Annyv2:
https://github.com/auth0/docs/issues/683
https://github.com/auth0/docs/issues/745


ntotten:
I believe the reason is that it wont work on Android. Did you confirm that this runs and everything?

Annyv2:
Yes, I've tested it in Android.

PR #926, opened 2016-02-17
[edit] Overall edits, resized and tinypnged images

Catografix:


articles/step-up-authentication.md
```diff
@@ -78,30 +78,30 @@ router.get('/employees', requireMultiFactor, function(req, res) {
 });
 ```
 
-> Make sure to enable the **Use Auth0 instead of the IdP to do Single Sign On** option on your application in Auth0 or specify a connection when triggering the step-up authentication flow to make sure users don't have to authenticate all over again. By enabling the SSO setting or specifying a connection name you will trigger an SSO login which will only prompt the user for multifactor authentication.
+**NOTE:** Enable the **Use Auth0 instead of the IdP to do Single Sign On** option in the settings page of your app in Auth0 or specify a connection name when triggering step-up authentication to make sure users are not forced to authenticate when already logged in. By enabling the SSO setting or specifying a connection name, your app will trigger an SSO login which will only prompt the user for multifactor authentication.
```
Catografix, line 81: @sandrinodimattia Please check if my changes here are correct. Also, where should they specify a connection name?

sandrinodm, line 81: Looks good!

PR #944, opened 2016-02-22
[new] system requirements to NancyFx tutorial

vmartynets:


articles/server-platforms/nancyfx.md
```diff
@@ -12,6 +12,11 @@ snippets:
 
 ## NancyFX Tutorial
 
+::: panel-info System Requirements
+This tutorial and seed project have been tested with the following:
+* Microsoft Visual Studio 2015
+:::
```
ntotten, line 19: Does this use a specific version of .NET?

vmartynets, line 19: For example, [sample](https://github.com/auth0/Auth0.NancyFx.SelfHost/tree/master/sample) targets to .NET framework 4.5 (as you can see [here](https://github.com/auth0/Auth0.NancyFx.SelfHost/blob/master/sample/auth0-nancyfx-sample.csproj#L12))

ntotten, line 19: We should specify the .NET framework version as well

Issue #938, opened 2016-02-20
Missing Image in articles/saml/identity-providers/salesforce/


Catografix:
In `articles/saml/identity-providers/salesforce/`, the image `/media/articles/saml/identity-providers/salesforce/salesforceidp-8.png` is missing. The original link to the file in the CDN (https://cdn.auth0.com/docs/img/salesforceidp-8.png) was also broken.

mpaktiti:
Image reference removed.

PR #949, opened 2016-02-22
[edit] Overall edits to saml/identity-providers/salesforce

Catografix:
Deleted reference to missing screenshot. ( bug #938 ) Can't recreate without advanced Salesforce account. Should I leave the bug open?


ntotten:
Leave the bug open, we should get a screenshot.

PR #974, opened 2016-02-27
Update aspnet-mvc4-enterprise-providers.md

utahisnotastate:
I had some confusion reading through and editing this article, and was hoping for some clarification. 

I think my main source of confusion when editing/revising this tutorial stems from the previous tutorial this tutorial links. The ASP.NET tutorial does not have any reference to google 0Auth2. I went through every link in the previous tutorial, but they all show up as "Page not Found". I wasn't too sure if I didn't have permission or not to view them, but it was hard for me to ensure my understanding of this tutorial when the following links showed up as not found in the previous tutorial:

https://github.com/auth0/docs/blob/master/articles/server-platforms/$%7BuiAppSettingsURL%7D
https://github.com/auth0/docs/blob/master/server-apis/wcf-service
https://github.com/auth0/docs/blob/master/user-profile
https://github.com/auth0/docs/blob/master/link-accounts

I feel with additional clarification, I may have a better grasp on the technical aspects of the article. Sorry if I seem overly critical or harsh, but just want to make sure I give feedback on the viewpoint of a developer who is still trying to learn Auth0.


ntotten:
@jerriep Can you take a look at this?

jerriep:
Hey @utahisnotastate , the problem you have with not being able to view those links, is because you are viewing them locally, and the paths get adjusted when the documentation gets generated.

The best would be to view the documentation online:
https://auth0.com/docs/quickstart/webapp/aspnet/

If you do that, then all those links will open up just fine :)

One you have gone through that, please let me know what other clarification you need.

jerriep:
As far as the Google OAuth2 reference in this article, I am also not sure which tutorial that refers to. @ntotten: Any idea?

PR #986, opened 2016-02-29
[update] Updated Ionic 2 doc to reflect current dep versions and adde…

chenkie:
Updated Ionic 2 doc to reflect current dep versions and added token refreshing.


snippets/native-platforms/ionic2/setup.md
```diff
@@ -1,6 +0,0 @@
-```ts
```
ntotten, line 1: Why did we remove these? We need these for the SEO landing pages.

snippets/native-platforms/ionic2/use.md
```diff
@@ -1,20 +0,0 @@
-```ts
```
ntotten, line 1: same as above

snippets/native-platforms/ionic2/setup.md
chenkie, line 1: Ah gotcha. I just needed to replace them with new content. I can rename the markdown files that contain these steps to `setup` and `use` if you like.

PR #994, opened 2016-03-02
[bug] fix link to lock script at electron tutorial

### LINKED ISSUE #854: electron link fix
Need to add `https:` to lock library link in this step
image

image

vmartynets:
fix https://github.com/auth0/docs/issues/854


ntotten:
I need to look into this. I think that link should be like it was, but it might be a bug in the docs site. Good catch though!

vmartynets:
@ntotten looks like, if you link lock script without `https:`, it considers that that file is local

PR #1004, opened 2016-03-03
[new] Add custom-error-pages doc

khorne3:


articles/custom-error-pages.md
```diff
@@ -0,0 +1,113 @@
...
+![](/media/articles/error-pages/redirect-error-page.png)
+
+## Customizing Error Pages via the Management API
+Instead of using the Management Portal, you may configure your error pages by making the appropriate PATCH call to the Management API.
+
+![](/media/articles/error-pages/patch-tenant-settings.png)
+
+Prior to beginning, please ensure that you are logged in to an account that is permitted to make changes to your Auth0 configuration.
+
+1. Navigate to the [APIv2 Explorer Page](/api/v2).
```
ntotten: So for this, i would say we shouldnt recommend using the API explorer as a UI interface. I would document this explaining how to use the API directly. Explaining the PATCH call to the api and what the request will look like. In the beginning of the section maybe mention that you can either use the API directly or if they want they can try this via the management explorer.

I would show how to send the command using CURL. That is kind of the standard way to do it.

Bonus points if you want to create a run in postman button: http://blog.getpostman.com/2016/01/28/the-run-in-postman-button-for-improved-documentation/

articles/custom-error-pages.md
```diff
@@ -0,0 +1,113 @@
...
+
+-	Friendly Name: a user-friendly version of your company's name;
+-	Logo URL: the path to the logo you want to show to users;
+-	Support Email: the email address for your Support team;
+-	Support URL: the URL of your Support team's webpage.
+
+The following JSON will set the appropriate fields when included in your call to the API:
+
+```
+"friendly_name": "My Company",
```
ntotten: use a valid json object by wrapping this in `{ }`

articles/custom-error-pages.md
```diff
@@ -0,0 +1,113 @@
...
+"picture_url": "https://mycompany.org/logo.png",
+"support_email": "support@mycompany.org",
+"support_url": "https://mycompany.org/support"
+```
+
+### Customized Error Pages
+
+#### Redirecting Users to a Custom Error Page
+To redirect users to a custom error page, update the "url" field to point to the location of the error page:
```
ntotten: What is the API endpoint to send this to. This is a bit unclear, think about the cases where users are linked directly to a subsection of the doc. Each subsection should be usable independently or at least point to relevant documentation. It is a common use case to link users to a section of an article like: https://auth0.com/docs/connections/database/migrating#enable-automatic-migration

For any place that includes an API you should generally include the following:

METHOD === the HTTP Method (GET, POST, etc.)

```
METHOD https://url
```

Request Body:

```
{
   "foo": "baz"
}
```

Example cURL Command:

```
curl -X GET -H "Authorization: Bearer YOUR_TOKEN" "https://sdk-ruby-test.auth0.com/api/v2/users"
```

PR #1011, opened 2016-03-04
[edit] Added warning banner to email index

Catografix:
I think that the warning message is best placed on the index page as per my understanding of which service is being deprecated. I moved it up from a blockquote to the top of the page.


ntotten:
Lets use a panel-danger instead. The warning is really only for cases when something is deprecated or broken. https://github.com/auth0/docs#panels

Catografix:
fixed

PR #1049, opened 2016-03-16
[update] Add info on UI use to Logout doc

khorne3:
Add usage instructions regarding redirectTo parameter in management portal


ntotten:
@khorne3 Could you rebase, looks like there is a conflict.

PR #1080, opened 2016-03-23
[DO NOT MERGE] Add new docs on access token and id tokens

jerriep:
This is the starting point for the new access token and id token documents.

Lots of questions still, but it would be good to get feedback.

The contents of the old documents are in the __old_ files.

### DIFF
--- a/articles/tokens/access_token-vs-id_token.md
+++ b/articles/tokens/access_token-vs-id_token.md
@@ -1,0 +1,25 @@
+---
+title: Auth0 access_token vs id_token
+---
+
+# Auth0 access_token vs id_token
+
+## When should you use the `access_token` and the `id_token`?
+
+One of the common questions that gets asked is when you should use the `access_token` and when you should use the `id_token`? 
+
+The answer is as follows:
+
+* If you need to pass a user's credentials to your own API in order to authenticate the user against the API, then you should use the `access_token`
+* When you want to get information about a user from the [/tokeninfo endpoint](/auth-api#post--tokeninfo), then you should use the `id_token`
+
+The situation above is the ideally correct answer, but in reality things are not that simple. Due to historical reasons, `access_token` which is issued by Auth0 is not always a JWT. It used to be a proprietary format which was used to call some of the methods on the [Authentication API](/auth-api).
+
+The JWT format for the `access_token` was only introduced with the new [OAuth2 Authentication Flow](https://auth0.com/docs/api-auth), and you will therefore only be issued a JWT when the user authenticates using the new flow.
+
+If you are still using the older authentication flow, then you should be passing the `id_token` to your own APIs to authentication the user. **Please note** however that passing the `id_token` is strongly discouraged, and you should therefore only use it in this scenarion. If you are using the new OAuth2 authentication flow, then you should be passing the `access_token`.
+
+## Why is the `access_token` more secure?
+
+Can we get a short writeup here??
+
--- a/articles/tokens/access_token-vs-id_token.md
+++ b/articles/tokens/access_token-vs-id_token.md
@@ -4,22 +4,23 @@ title: Auth0 access_token vs id_token
 
 # Auth0 access_token vs id_token
 
-## When should you use the `access_token` and the `id_token`?
+One of the common questions we get asked is what the difference is between the `access_token` and the `id_token`, and what the use cases are for each. 
 
-One of the common questions that gets asked is when you should use the `access_token` and when you should use the `id_token`? 
+* The `access_token` should be sent as credentials when making API calls on behalf of a user.
+* The `id_token` is used by the client to obtain profile information about the user, by decoding and verifying the claims in the JWT. The `id_token` should never be sent to an API. 
 
-The answer is as follows:
+The `id_token` is an optimization to avoid network requests to the `/userinfo` endpoint. If the `/userinfo` API is called using the `access_token`, it should return the same claims that are encoded directly in the `id_token`. **Note** that in some cases, the `/userinfo` endpoint may return more information, due to size constraints with JWT id_tokens.
 
-* If you need to pass a user's credentials to your own API in order to authenticate the user against the API, then you should use the `access_token`
-* When you want to get information about a user from the [/tokeninfo endpoint](/auth-api#post--tokeninfo), then you should use the `id_token`
+## One caveat
 
-The situation above is the ideally correct answer, but in reality things are not that simple. Due to historical reasons, `access_token` which is issued by Auth0 is not always a JWT. It used to be a proprietary format which was used to call some of the methods on the [Authentication API](/auth-api).
+One caveat to this however is that due to historical reasons the `access_token` which was issued by Auth0 did not use to be a JSON Web Token. It used to be a proprietary format which was used to call some of the methods on the [Authentication API](/auth-api).
 
 The JWT format for the `access_token` was only introduced with the new [OAuth2 Authentication Flow](https://auth0.com/docs/api-auth), and you will therefore only be issued a JWT when the user authenticates using the new flow.
 
-If you are still using the older authentication flow, then you should be passing the `id_token` to your own APIs to authentication the user. **Please note** however that passing the `id_token` is strongly discouraged, and you should therefore only use it in this scenarion. If you are using the new OAuth2 authentication flow, then you should be passing the `access_token`.
+If you are still using the older authentication flow, then you should be passing the `id_token` to your own APIs to authentication the user. **Please note** however that passing the `id_token` is strongly discouraged because it is less secure than the  `access_token`. If you are using the new OAuth2 authentication flow, then you should be passing the `access_token`.
 
 ## Why is the `access_token` more secure?
 
-Can we get a short writeup here??
+The `access_token` is only intended to be consumed by the API (Resource Server), which is why its `aud` (audience) claim is set to the resource_server's identifier and why its signed by the Resource Server's signing secret. Therefore only the API (not the Client) can verify and trust the `access_token` JWT.
 
+In contrast, the `id_token` will only be consumable and trusted by the Client.
--- a/articles/tokens/access_token.md
+++ b/articles/tokens/access_token.md
@@ -1,14 +1,28 @@
 ---
 title: Auth0 access_token
+sitemap: false
 ---
 
 # Auth0 `access_token`
 
 ## Overview
 
-The Auth0 access token, usually called `access_token` in the code samples, is an opaque string which is returned by calls that invoke the Auth0 authentication process and is only used to call the Auth0 API.  The Auth0 access token does not contain any claims, cannot be decoded, does not conform with any standard and is not signed.
+The Auth0 access token, usually called `access_token` in the code samples, is a [JSON Web Token](/jwt) which can be used to authorize users of your APIs when using our [API Authentication](/api-auth)
+
+It conforms to an industry standard (IETF RFC 7519) and contains three parts: A header, a body and a signature. The header contains the type of token and the hash algorithm used on the contents of the token.  
+
+The body, also called the payload, contains identity claims about a user.  There are some non-mandatory claims with registered names, for things like the issuer of the token, the subject of the token (who the claims are about), and the time of issuance.  Any number of additional claims with other names can be added, though care must be taken to keep the JWT within the browser size limitations for URLs.  
+
+The third part of the JWT is the signature which is used by the recipient of a JWT to validate the integrity of the information conveyed in the JWT.
 
 The [Auth0.js documentation](/libraries/auth0js) shows a sample of how to get the `access_token`.
+
+::: warning-banner
+The `access_token` used to be an opaque string that was used to call the Auth0 API, and in some circumstances it still is. For more details on when the `access_token` is a JWT, and when it is an opaque string, please see [Access Token vs ID Token]()
+
+!! Must still do this doc !!
+:::
+
 
 ## How to get Auth0 access token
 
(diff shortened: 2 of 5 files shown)


jerriep:
/cc @jaredhanson @ntotten 

We need definite guidance on when to use which, i.e. what is the switch over procedure from the old to the new. Last guidance we had was this:
https://auth0.slack.com/archives/oauth2aas/p1456866256000648

jerriep:
The *_old files are to be removed once we are happy with the new content

 @jaredhanson  @twistedstream Can you guys please look at these docs as I would like to get them finalised and pulished. A lot of the upcoming docs are going to depend on us having these docs correct, especially the following ones:
- access_token.md
- id_token.md
- access_token-vs-id_token.md

articles/tokens/access_token-vs-id_token.md
```diff
@@ -0,0 +1,25 @@
...
+# Auth0 access_token vs id_token
+
+## When should you use the `access_token` and the `id_token`?
+
+One of the common questions that gets asked is when you should use the `access_token` and when you should use the `id_token`? 
+
+The answer is as follows:
+
+* If you need to pass a user's credentials to your own API in order to authenticate the user against the API, then you should use the `access_token`
+* When you want to get information about a user from the [/tokeninfo endpoint](/auth-api#post--tokeninfo), then you should use the `id_token`
```
twistedstream: @jerriep: I question whether someone would even call the `/tokeninfo` endpoint. If they have an `id_token` it should have been scoped to contain all of the user profile claims their app needs. If they need more than this, then they should use the `access_token` to call the `/userinfo` endpoint to get the _full_ profile.

What are your thoughts @jaredhanson ?

articles/tokens/access_token-vs-id_token.md
```diff
@@ -0,0 +1,25 @@
...
+
+The situation above is the ideally correct answer, but in reality things are not that simple. Due to historical reasons, `access_token` which is issued by Auth0 is not always a JWT. It used to be a proprietary format which was used to call some of the methods on the [Authentication API](/auth-api).
+
+The JWT format for the `access_token` was only introduced with the new [OAuth2 Authentication Flow](https://auth0.com/docs/api-auth), and you will therefore only be issued a JWT when the user authenticates using the new flow.
+
+If you are still using the older authentication flow, then you should be passing the `id_token` to your own APIs to authentication the user. **Please note** however that passing the `id_token` is strongly discouraged, and you should therefore only use it in this scenarion. If you are using the new OAuth2 authentication flow, then you should be passing the `access_token`.
+
+## Why is the `access_token` more secure?
+
+Can we get a short writeup here??
```
twistedstream: @jerriep: Some points to provide here:
- The `access_token` is only intended to be consumed by the API (`resource_server`), which is why its `aud` (audience) claim is set to the `resource_server`'s identifier and why its signed by the `resource_server`'s signing secret. Therefore only the API (not the Client) can verify and trust the `access_token` JWT.
- In contrast, the `id_token` in the old flows is typically signed by the same secret as the Client, which means there's an opportunity for the Client to generate their own JWT and spoof the API. In the new OAuth2 flows, the `id_token` will _only_ be consumable and trusted by the Client.
- The new `access_token` uses the more standard `scope` claim to represent the access rights of the Client vs. custom claims that we can generate today in the `id_token`.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+
 ## How to get Auth0 access token
 
 The Auth0 access token can be obtained in several ways.
 Calls to Lock or library functions that invoke authentication will return the `access_token`.
 
-* Calls to the Lock widget will return `access_token` as shown in the [Lock documentation](/libraries/lock)
-* [Examples using auth0.js](https://github.com/auth0/auth0.js)
-* The [/authorize endpoint in the authentication API](/auth-api) will return `access_token`
+* Calls to the Lock widget will return `access_token` as shown in the [Lock documentation](/libraries/lock) ?? Is this true - can we still use lock ??
```
twistedstream: @jerriep: To my knowledge Lock (and really auth0.js) has _not_ been configured to call the new OAuth2 `/authorize` endpoint. We'll need to check with that crew to find out how they will support configuring which flow to use. The same will be true for the mobile SDK versions of Lock.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+* [Examples using auth0.js](https://github.com/auth0/auth0.js) ?? is this still true ??
+* The [/i/oauth2/authorize endpoint in the authentication API](/auth-api) when using the Authorization Code Grant flow or Implicit Grant flow will return an `access_token`
+* The [?? endpoint in the authentication API](??) when using Client Credentials Grant will issue an `access_token`
```
twistedstream: The `client_credentials` endpoint is `/oauth/token`. Example:

``` sh
curl -X POST -H 'Content-Type:application/json' https://TENANT.auth0.com/oauth/token/oauth/token -d '
  {
    "audience": "https://foo-api.com/api",
    "grant_type": "client_credentials",
    "client_id": "CLIENT_ID",
    "client_secret": "CLIENT_SECRET"
  }'
```

I don't think we have any docs in the Authentication API for this endpoint yet.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
 * Check the [List of tutorials](/tutorials) to see how to make calls to libraries for other languages/SDKs.
 
 ## How to control contents of access token
 
-It is not possible at this time to control the contents of the Auth0 access token.
+The contents of the `access_token`, specifically the claims contained within it, are controlled by the use of a parameter called `scope` which is passed to the authentication functions mentioned above.  For example, the call to the Lock widget’s `.show` function can specify optional authentication parameters as follows:
+
+?? Is this still true - can we use lock ??
```
twistedstream: @jerriep: of course this examples assumes Lock support for the new flows

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+lock.show({
+   responseType: ‘token’,
+   authParams: {
+      scope: ‘appointments contacts’
+    }
+ });
+```
+The above sample, specifying `appointments contacts` will result in a JWT with claims for accessing appointments and contacts. The responseType should be `token` for Implicit Grant flow and `code` for Authorization Code Grant flow.
+
+The scope of the `access_token` JWT can also be altered via Rules, through the context.jwtConfiguration.scopes object as documented [here](https://github.com/auth0/docs/blob/32033877180affa26233b8f65cb28bd532514eab/articles/rules/index.md#context) ?? Should be still suggest this. Also AFAIK rules won't work anymore with the new endpoint ??
```
twistedstream, line 46: @jaredhanson: Any ideas of what we can say here? The above looks like a hack and I thought rules didn't even run at all in the new flows. Any notion of what a rules-equivalent might look like for the new flows?

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+
+There is a [sample for altering scopes in a Rule](https://github.com/auth0/rules/blob/dff2a3e72f01d33af3086414be7cf115b19eea0c/rules/custom-scopes.md)
```
twistedstream, line 48: Similar to my previous comment, I doubt this will be the correct mechanism for modifying the scopes in an `access_token`.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+A valid JWT can be pasted into the [jwt.io website](https://jwt.io) to view the contents of the JWT.
+
+A [blog entry on JWT](https://auth0.com/blog/2015/07/21/jwt-json-webtoken-logo/) provides an explanation of why it is getting to be more popular.
+
+[Additional samples](https://github.com/auth0/auth0.js) show use of the auth0.js library.
 
 ## Validity
 
-The Auth0 access token remains valid for 24 hours.
+In the case of the [authorization code grant](/api-auth/grant/authorization-code) and the [implicit code grant](/api-auth/grant/implicit) flows the `access_token` is valid for ???
```
twistedstream: This is controlled by the `resource_server`'s Token Expiration property. Looks like the default currently is 86400 seconds.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+
+In the case of the [client credentials flow](/api-auth/grant/client-credentials), the `access_token` is valid for 8000 seconds (roughly 2.25 hours) by default.  !!The expiration of this token can be set in the `Apps/APIs` -> Settings screen using the `JWT expiration` field!!
+
+The validity period of the token can be configured per [Resource Server](??)
 
 ## Renewing the token
 
-There is currently no way to renew the Auth0 access token without triggering authentication by the user.  To get a new access token, simply follow the process for getting an access token as described above in the How to Get an Access Token section.
+?? How do we get a refresh token ??
```
twistedstream: These are both @jaredhanson questions, but my assumption is if the `code` flow is used and `scope` contains `offline_access`, a `refresh_token` will be granted.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+?? How do we use the refresh token to obtain a new access token ??
```
twistedstream: Per the spec, by calling the `/oauth/token` endpoint with the `refresh_token`

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
 
 ## Termination of the token
 
-Once issued, Auth0 access tokens can not be revoked.  
+Once issued, tokens can not be revoked in the same fashion as cookies with session id’s for server-side sessions.  As a result, tokens should be issued for relatively short periods, and then renewed periodically if the user remains active.  See the above section on renewing access tokens.
```
twistedstream, line 64: IMO, `access_token` revocation is something that should be controlled by the API (`resource_server`) itself using techniques like the one we use in our own Management API and that are described in [this blog post](https://auth0.com/blog/2015/03/10/blacklist-json-web-token-api-keys/).

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
 
 ## Uses
 
-The Auth0 access token is used to call the Auth0 Management APIv1.  This version of the Auth0 Management API has been deprecated and should no longer be used.
+The `access_token` is designed to be used to pass information about a user between websites,  web programs and APIs in an industry standard, URL-friendly fashion.  One advantage of using an `access_token` for this purpose is that the recipient can validate the token without having to make a call back to the issuer of the token.  The token is also designed to enable being passed from one web property to another, via an untrusted client, such that the client cannot alter the token without such tampering being evident to the recipient.
+
+It can also be used to call the /userinfo endpoint within the authentication API to get user profile information as documented [here](/auth-api#user-profile) (?? is this still valid ??)
```
twistedstream: My understanding is that this will still work. The new `access_token` will be accepted by the existing `/userinfo` endpoint and return the full user profile. @jaredhanson is that correct?

articles/tokens/id_token.md
```diff
@@ -81,24 +81,28 @@ Once issued, tokens can not be revoked in the same fashion as cookies with sessi
 
 ## Uses
 
-The `id_token` is designed to be used to pass information about a user between websites,  web programs and APIs in an industry standard, URL-friendly fashion.  One advantage of using an `id_token` for this purpose is that the recipient can validate the token without having to make a call back to the issuer of the token.  The token is also designed to enable being passed from one web property to another, via an untrusted client, such that the client cannot alter the token without such tampering being evident to the recipient.
-
 The `id_token` can be used to call the /tokeninfo endpoint within the Auth0 authentication API to retrieve the user’s complete profile.  See the [/tokeninfo endpoint documentation](/auth-api#user-profile) for more details.
 
 The `id_token` can also be used to call the /delegation endpoint within the Auth0 authentication API to obtain another token for another API.  See the [/delegation endpoint documentation](/auth-api#delegated) for more information.
```
twistedstream: I'm not sure this will be valid in the new flows. @jaredhanson have we worked out an equivalent to delegation with the new flows yet? My assumption is that they will use the `access_token` and not the `id_token`.

articles/tokens/access_token-vs-id_token.md
jaredhanson: Yes.  The `id_token` should never be supplied to any API, including `/tokeninfo` and `/userinfo`.

There are some cases involving delegation (true delegation, not our version of it) that may use an id_token.  We are a long way off from implementing those, so I don't think those are a concern for now.

I would rephrase this as:
- The access token should be sent as credentials when making API calls on behalf of a user.
- The id_token is used by the client to obtain profile information about the user, by decoding and verifying the claims in the JWT.  The id_token should never be sent to an API.  The id_token is an optimization to avoid network requests to the /userinfo endpoint.   If the /userinfo API is called using the access_token, it should return the same claims that are encoded directly in the id_token.   Note that in some cases, the /userinfo endpoint may return _more_ information, due to size constraints with JWT id_tokens.

Something like that...

articles/tokens/access_token-vs-id_token.md
```diff
@@ -0,0 +1,25 @@
...
+* When you want to get information about a user from the [/tokeninfo endpoint](/auth-api#post--tokeninfo), then you should use the `id_token`
+
+The situation above is the ideally correct answer, but in reality things are not that simple. Due to historical reasons, `access_token` which is issued by Auth0 is not always a JWT. It used to be a proprietary format which was used to call some of the methods on the [Authentication API](/auth-api).
+
+The JWT format for the `access_token` was only introduced with the new [OAuth2 Authentication Flow](https://auth0.com/docs/api-auth), and you will therefore only be issued a JWT when the user authenticates using the new flow.
+
+If you are still using the older authentication flow, then you should be passing the `id_token` to your own APIs to authentication the user. **Please note** however that passing the `id_token` is strongly discouraged, and you should therefore only use it in this scenarion. If you are using the new OAuth2 authentication flow, then you should be passing the `access_token`.
+
```
jaredhanson, line 21: I'm not sure how to clarify this, but I would omit any discussions about what format the access_token is in.  JWT or something else, it really doesn't matter and doesn't effect the advice around when to use an access_token.

From the client perspective, the access_token is entirely an opaque string of bytes.  It's job is to just send that access token to the resource server for authentication.   The resource server will have to be able to decode the token of course.

APIv1 understands opaque tokens encoded in a db.
APIv2 understands JWT tokens.

Clients don't care.  They just make API calls using whatever token was given to them.

articles/tokens/access_token.md
```diff
@@ -1,49 +1,115 @@
...
+
+It conforms to an industry standard (IETF RFC 7519) and contains three parts: A header, a body and a signature. The header contains the type of token and the hash algorithm used on the contents of the token.  
+
+The body, also called the payload, contains identity claims about a user.  There are some non-mandatory claims with registered names, for things like the issuer of the token, the subject of the token (who the claims are about), and the time of issuance.  Any number of additional claims with other names can be added, though care must be taken to keep the JWT within the browser size limitations for URLs.  
+
+The third part of the JWT is the signature which is used by the recipient of a JWT to validate the integrity of the information conveyed in the JWT.
 
 The [Auth0.js documentation](/libraries/auth0js) shows a sample of how to get the `access_token`.
 
+::: warning-banner
```
jaredhanson: Again, I think this is misleading.  Focusing too much on the encoding of the token leads to unnecessary confusion in my opinion.

articles/tokens/access_token.md
jaredhanson: @twistedstream For the purposes of documentation changes, we are documenting the "correct" thing.  Correct in this case is recinding our earlier recommendations that were insecure.  Also, we are finalizing on `/authorize` as the endpoint that will implement this functionality.   Most people should never be aware that `/i/oauth2/authorize` ever existed.

The caveat of this is we may have things (implementation, docs, SDKs) that are not entirely in sync for a while.  This is the consequence of having a hole to dig ourselves out of.

articles/tokens/access_token.md
jaredhanson, line 46: `context.jwtConfiguration.scopes` is most likely not going to effect the access_token, and will continue to add scope to the id_token.  This entire usage should be deprecated as it makes no sense.

It remains to be determined, but we will add rules support for modifying the access_token, but for backwards compatiblity this will be  a new property.  Probably something liek `context.accessToken.scopes`.

articles/tokens/access_token.md
jaredhanson: The validity period is the same, regardless of user or client flows.  Both are controlled by the resource server settings (API tab in dashboard)

articles/tokens/access_token.md
jaredhanson: Yes, by requesting offline_access, you will get a refresh token that can refresh an access token.  This is currently not implemented correctly, but it will be.   Issuing of refresh tokens will be at the discression of the resource server.  If the refresh server disables this, no refresh token will be issued, and the client will have to go through a /authorize flow again to obtain a new token.

articles/tokens/access_token.md
jaredhanson: Yes

articles/tokens/access_token.md
jaredhanson, line 64: That blog post is fine, but misleading.  What we should describe is not the revoking of tokens, but the revoking of _grants_.  This is what OAuth's approach is, and what actually happens with the GitHub screen shots in the blog post.

This is best discussed over video chat.

articles/tokens/id_token.md
```diff
@@ -81,24 +81,28 @@ Once issued, tokens can not be revoked in the same fashion as cookies with sessi
 
 ## Uses
 
-The `id_token` is designed to be used to pass information about a user between websites,  web programs and APIs in an industry standard, URL-friendly fashion.  One advantage of using an `id_token` for this purpose is that the recipient can validate the token without having to make a call back to the issuer of the token.  The token is also designed to enable being passed from one web property to another, via an untrusted client, such that the client cannot alter the token without such tampering being evident to the recipient.
-
 The `id_token` can be used to call the /tokeninfo endpoint within the Auth0 authentication API to retrieve the user’s complete profile.  See the [/tokeninfo endpoint documentation](/auth-api#user-profile) for more details.
```
jaredhanson: Nope, remove this.

articles/tokens/id_token.md
jaredhanson: I would remove mention of delegation flows from the id_token documentation.  Our whole `/delegation` endpoint is a disaster.  We should document its use separately, and then bury it as deep as possible.

jaredhanson: As a general rule, here is my recommendation for cases where we've done bad things that need to be deprecated:
1. Update conceptual documentation to be correct.
2. Do not mention or link to bad or deprecated advice
3. Move all bad/deprecated advice to documentation about a specific topic (ie /delegation)

In this particular instance, the only mention of delegation should be on delegation-focused documentation.  Let's not pollute the rest of the documentation with mistakes.

articles/tokens/id_token.md
```diff
@@ -81,24 +81,28 @@ Once issued, tokens can not be revoked in the same fashion as cookies with sessi
...
-
 The [Delegation token request sample](https://github.com/auth0/auth0.js#delegation-token-request) provides further examples of using the id_token for other APIs.
 
 This [Animated Sequence Diagram](/sequence-diagrams) shows the sequence of calls used to get an `id_token`.
 
+::: warning-banner
+It is not suggested that you use an `id_token` to secure an API. For this purpose, you should rather make use of [access tokens](/tokens/access_token).
+:::
+
+?? What is the uses of the id token going forward? Access token will now be used in many instances we used id token before ??
```
jaredhanson: The id token basically has no real use, in the general case.  At best, it is an optimization to avoid network traffic to the /userinfo endpoint.

Consider it this way: the biggest IdP in the world (Facebook) does not use an id_token.  90% of all use cases do not need it.  Keeping that ratio: the access_token should recieve 90% of our attention wrt documenation, and the id_token should recieve 10%.  This is a big shift in mindset.

jaredhanson: The id_token is an extension to OAuth, defined by OpenID Connect, that returns the user's profile in a signed and verifiable manner.  That is it.  It has no security properties with respect to API calls.

articles/tokens/access_token.md
jaredhanson: Yes.  I think this work today.

jerriep:
We also need to fix https://auth0.com/docs/sequence-diagrams as it states that the id_token is passed to the API

mpaktiti:
Hi @jerriep Just checking if this PR is still valid and doesn't need to be closed.

auth0build:

jerriep:
@mpaktiti I do not think it is valid anymore, but there is knowledge from Jared and Peter in those documents which I do not just want to throw away. Whenever the new documents are done for the new `access_token` and OAuth-aas, the information in these documents may come in handy

@ntotten What should we do with this?

ntotten:
@mpaktiti Jerrie wrote these a long time ago. Might be something in here you could use, but it is probably very outdated. I'm going to close this.

PR #1098, opened 2016-03-28
removed old content, added configurations, compressed details

chriskeyser:
[Update] Revised the main landing page, and added detail about appliance configuration options.
### Description

Revised the landing page for appliance to add details on supported configuration models and deployment options. Removed stale content for features we no longer support.


ntotten:
@chriskeyser one thing to note, since you deleted a few files you need to add redirects.  https://github.com/auth0/docs/blob/master/redirect-urls.json

chriskeyser:
fixed redirects and confirmed with eugenio.

PR #1143, opened 2016-04-07
[update] Update email provider docs with the new api key authentication for sendgrid

hugoArregui:
### Description

Update email providers docs with the new api  key authentication for sendgrid, this depends and it's related to https://github.com/auth0/manage/pull/1655


articles/email/providers.md
```diff
@@ -97,12 +97,12 @@ The [Outbound Activity](https://mandrillapp.com/activity) page in Mandrill will
 3. To configure SendGrid as your email provider in Auth0, post your previously obtained SendGrid credentials to the [Configure the email provider](/api/v2#!/Emails/post_provider) endpoint with this curl command:
 
   ```js
-curl -H "Authorization: Bearer YOUR_AUTH0_V2_API_TOKEN" -X POST  -H "Content-Type: application/json" -d '{"name":"sendgrid","credentials":{"api_user":"YOUR_SENDGRID_API_USER","api_key":"YOUR_SENDGRID_API_KEY"}}' https://${account.tenant}.auth0.com/api/v2/emails/provider
+curl -H "Authorization: Bearer YOUR_AUTH0_V2_API_TOKEN" -X POST  -H "Content-Type: application/json" -d '{"name":"sendgrid","credentials":{"api_key":"YOUR_SENDGRID_API_KEY"}}' https://${account.tenant}.auth0.com/api/v2/emails/provider
```
ntotten, line 100: This should be changed to HAR format so it automatically generates the code snippets in a variety of languages. See: https://github.com/auth0/docs#http-request-snippets

hugoArregui, line 100: @ntotten I can do that if you want, but all the examples in this document are using curl, so I think it would looks a bit weird if we use HAR only here. IMHO, we should left curl here and perhaps create another PR using HAR everywhere. What do you think?

ntotten, line 100: sure

PR #1147, opened 2016-04-07
[DO NOT MERGE] Angular 1 quickstart section and updates to existing Angular 1 tutorial

christiannwamba:
### Existing Angular 1 updates
- Little restructure
- Added password reset example
- Added role authorization with rules example
### AngularJS multi-step quickstart
- Beginning of Angular quick start  with Setup & Login (Step 1)


updates/2016-04-07.yml
```diff
@@ -1,7 +1,10 @@
 added:
   -
-    title: "Added doc for Anomaly Detection"
```
ntotten: i think this is a mistake. This file shouldn't be changed.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+  dependencies: /quickstarts/angular1/snippets/dependencies
+  init: /quickstarts/angular1/snippets/init
+  head: /quickstarts/angular1/snippets/head
+alias:
+  - angular
+---
+
+## Setup and Authentication
+
+One of the most aching aspects of an Angular App is authentication. It becomes worse when you have routes to protect, page refresh to deal with, tokens to manage, multi-factor authentication, and so on. This quickstart is planned to help you tackle these challenges.
```
ntotten: I think this part isnt needed. Our customers have come to our site specifically to solve the issue of authentication so I think this is a bit redundant.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+
+This is the very beginning of a simple, practical and multi-step quickstart that will guide you through managing authentication in your Angular JS apps with Auth0.
+
+
+### Create an Application
+
+Create an [Auth0 account](https://manage.auth0.com/login) and an authentication application from the [dashboard](https://manage.auth0.com/#/). Once you create an app, you'll be provided with credentials (Domain, Client ID, and Client Secret) which should be stored somewhere safe (do not commit this information to your git repo!).
```
ntotten: Keep in mind that many people who are running this tutorial will 1) already have an auth0 account and 2) will already have an app. One thing you can do is check if they are logged in. It might be good to have two variants of this paragraph. 1) for anonymous users and 2) for users who have accounts.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+<%= include('../_includes/_package', {
+  pkgRepo: 'auth0-angular',
+  pkgBranch: 'master',
+  pkgPath: 'examples/widget-redirect',
+  pkgFilePath: null,
+  pkgType: 'js'
+}) %>_
+
+::: panel-info Running the Sample
+At any point in time, you can run the sample with `http-server`. First install with `npm install -g http-server` then run `http-server` on the root directory of the sample.
```
ntotten: At any point in time you can run this sample with a simple HTTP server. One example is `http-server` which can be installed...

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+:::
+
+::: panel-info Default Configuration
+Every instance of the seed project comes configured with your `Default App` credentials. Awesome right?
```
ntotten: we are going to change the tutorial navigator to allow an application picker so the user will be able to pick which app is configured for the tutorial and seed.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+:::
+
+### Configure Callback URLs
+
+Callback URLs are URLs that Auth0 invokes after the authentication process. It is especially useful in OAuth. Callback URLs can be manipulated on the fly and that could be harmful.
+
```
ntotten: Explain a bit more, your callback URL receives the token... etc. etc.

ntotten: Remove this " It is especially useful in OAuth."

Move and rephrase this "Callback URLs can be manipulated on the fly and that could be harmful." to the following paragraph. Just a single paragraph about security.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+```
+
+Going through these following sub-steps, you'll be able to replicate this process on your own Angular applications.
+
+#### 1: Setup Scripts, Viewport and Basic Style
+Some JavaScript dependencies are required for Auth0 to work as expected in an Angular app. Include the dependencies' scripts in yout `index.html`:
+
+${snippet(meta.snippets.dependencies)}
+
+These may seem like a lot of dependencies, but each one has a very important function. Go through and figure out if each package works for your use case. Feel free to strip off the unnecessary packages.
```
ntotten: Great!

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+ - **angular** is Angular's main library which you are building the application on.
+ - **angular-cookies** is Angular's wrapper for managing client cookies.
+ - **angular-route** is used to mange SPA routes in Angular application.
+ - **auth0-angular**: Auth0's SDK for Angular. Exposes most of the useful methods for authentication
+ - **angular-storage**: A `localStorage` and `sessionStorage` wrapper create with love by Auth0 team.
+ - **angular-jwt**: Angular service that makes using JWT easy in Angular apps
+
+ ::: panel-info Note
+ Some of the includes like angular-route and angular-storage will not be used in this step but will be useful in subsequent steps
+ :::
```
ntotten: Should we just add these when they are needed? What if a user doesn't follow all steps? They will have extra dependencies.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+Right after including the scripts, add a viewport to make the lock widget fit in to device widths:
+
+```html
+<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no" />
+```
+
+One more thing you can do in setting up is add a simple style sheet so the app does not look so boring:
+
+```html
+<link href="https://cdn.auth0.com/styleguide/4.4.24/index.css" rel="stylesheet" />
```
ntotten: They should not use our styleguide. this is for internal use only. Use raw bootstrap, but really this is probably not a required step.

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+
+Next, go ahead and add the app name in your html:
+
+```html
+<html ng-app="YOUR-APP-NAME">
+```
+
+An Angular-Auth0 app requires basically 2 configurations:
+
+ - Auth0's credential config with `init()`
```
ntotten: "Auth0's application configuration...

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+
+```javascript
+//Called when login is successful
+authProvider.on('loginSuccess', function(idToken) {
+  console.log("Login successful with token " + idToken);
+});
+
+//Called when login fails
+authProvider.on('loginFailure', function() {
+  alert("Error");
```
ntotten: We can do better here. How should a user handle an error?

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+The `authProvider` dependency is Auth0's API that exposes some methods including the `init()` method for for configuration:
+
+${snippet(meta.snippets.init)}
+
+Event listeners are available to handle different status of authentication. They are also configured in the `config()` method with `authProvider`:
+
+```javascript
+//Called when login is successful
+authProvider.on('loginSuccess', function(idToken) {
+  console.log("Login successful with token " + idToken);
```
ntotten: What do they do with the token now?

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+});
+
+//Called when login fails
+authProvider.on('loginFailure', function() {
+  alert("Error");
+});
+
+//Called when user is authenticated (Authentication flag is true)
+authProvider.on('authenticated', function() {
+  console.log("Authenticated");
```
ntotten: What is the difference between `loginSuccess` and `authenticated`? Do we need both?

quickstarts/angular1/01-login.md
```diff
@@ -0,0 +1,222 @@
...
+
+### Recap and Round Up
+At this point, you have learned how to:
+- Setup Auth0 in an Angular project
+- Configure Angular with Auth0 credentials and authentication status events
+- Login
+
+As mentioned earlier, this is just the beginning of a multi-step quickstart. You can continue to the next step when you are done here.
+
+::: panel-info Next Step
```
ntotten: Remove this part, we will generate the links to each step automatically for each document.

quickstarts/angular1/snippets/head.md
```diff
@@ -0,0 +1,24 @@
+```html
+<!-- Auth0's lock widget library -->
+<script type="text/javascript" src="https://cdn.auth0.com/js/lock-7.12.min.js"></script>
+<!-- AngularJS core -->
+<script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/angularjs/1.5.0-rc.0/angular.js"></script>
```
ntotten: Do we want to use rc releases? If they are close, this is fine as we will probably be a few weeks before this is fully ready.

articles/client-platforms/angularjs.md
```diff
@@ -13,14 +13,16 @@ tags:
   - quickstart
 snippets:
   dependencies: client-platforms/angularjs/dependencies
-  setup: client-platforms/angularjs/setup
-  use: client-platforms/angularjs/use
```
ntotten: @christiannwamba So the `dependencies`, `setup`, and `use` scripts are required. They are used for some of our SEO pages on auth0.com. You can change the content of the script, but not the name. They are used to generate these pages: https://auth0.com/authenticate/angularjs/active-directory

articles/client-platforms/angularjs.md
```diff
@@ -29,124 +31,275 @@ alias:
...
-
+### Prerequisites
 ::: panel-info System Requirements
 This tutorial and seed project have been tested with the following:
 
-* NodeJS 4.3
+* NodeJS 4.4
 * Angular version 1.5.0-rc.0
 :::
-${include('./\_callback')}
```
ntotten: This should be in the callback section. It tells users what their current callback is set to.

PR #1172, opened 2016-04-14
Coarse -> Fine grained authorization

jstrutz:
Rules permit more finely-grained authorization; coarse-grained would be the opposite direction.


sandrinodimattia:
The application itself is the only one that can enforce fine grained authorization (eg: can I click this button, can I modify this user's address, hide a button if that's not the case, also enforce it on the server side, ...)

Auth0 rules can only decide if you can access an application or not, which is coarse grained. I would change it back to coarse grained. 

Note: rules can load fine grained permissions from somewhere and pass them along to the application, but not enforce them.

jstrutz:
Aha; I misunderstood the context.  Thanks Sandrino!

PR #1216, opened 2016-04-20
[edit] Change FAQ page to point to public forum

mauricioschneider:
@yvonnewilson thought it would be better to directly point users to the public forum's FAQ category instead of having to add a question to this file each time a new FAQ topic was created in the forum.

I also added a message to encourage users to use the search feature to prevent duplicated questions in case they don't find what they're looking for in the FAQ category.

### DIFF
--- a/articles/faq.md
+++ b/articles/faq.md
@@ -1,10 +1,5 @@
 # FAQ
 
-This page contains pointers to frequently asked questions.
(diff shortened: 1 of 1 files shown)


ntotten:
What's the point of this doc then if all it does is point there? Lets just delete it.

PR #1240, opened 2016-04-26
Update Lock (for web) homepage images

gnandretta:
Updates the images in the Lock index page with ones that show the new widget's UI. Both images were taken from the project's README.


ntotten:
@gnandretta In the future please add the media directly to the github repo. We are moving all images out of third party services like cloudup.com.

PR #1250, opened 2016-04-28
[edit] APIv2 search - Overall edits, added descriptions

Catografix:
### Description

Overall edits, added descriptions


Catografix:
@ntotten One of these files is live, and one isnt. Is that possible?

PR #1261, opened 2016-05-02
Clarify change password process

mpaktiti:
The process of changing a user's password is a bit confusing to the customer, given the changes between API version 1 and 2. Some clarifications have been added to the document:
- Add note on how to check whether you are using API version 1 or 2
- List ways to change password
- Describe the process of change password using Lock
- Add clarifications on the steps that differ when a user tries to change the password using Lock (version 8 vs version 9)
- Rearrange paragraphs

Document to be reviewed by support and copyeditor before merge.

Also pending potential move of the document to the Guides section.


ntotten:
Very nice!

PR #1305, opened 2016-05-11
Quickstart 2: user profile

christiannwamba:
Additions: Quickstart step 2 (user profile). Demo available https://github.com/auth0-samples/auth0-angularjs-sample


updates/2016-05-11.yml
```diff
@@ -0,0 +1,8 @@
+added:
```
ntotten: remove this for now, the doc isnt public so we dont need to push an update. we will broadcast and update once the doc is public.

PR #1419, opened 2016-06-02
Geo HA

khorne3:
Geo HA Overview doc


articles/appliance/geo-ha2.md
```diff
@@ -0,0 +1,48 @@
+# Auth0 Appliance: High Availability Geo Cluster (Geo HA)
+
+The high availability geo cluster is an Appliance implementation that provides data center redundancy and automatic failure handling. This is the highest form of Appliance availability offered by Auth0.
+
+## Overview
+
+Auth0 adds to the single data center high availability solution by extending the cluster with a geographically distributed data center with a **maximum latency between the two data centers of no more than 100ms (round trip)**. The result is the high availability geo cluster, which is an active hot standby configuration with automated failure handling that can survive a regional outage.
```
chriskeyser: Katie, I'd soften this a bit. I think we'd be willing to discuss a latency > 100ms if the load is pretty low, or if the customer can tolerate more lost data. Could this read something like `the maximum round trip latency should be no more than 100 ms`?

articles/appliance/geo-ha2.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Standard Configuration
+
+![](/media/articles/appliance/geo-ha.png)
+
+The standard configuration is a stretch cluster that consists of the following pieces:
+
```
chriskeyser: stretched vs. stretch

articles/appliance/geo-ha2.md
```diff
@@ -0,0 +1,48 @@
...
+### Application Tier
+
+Auth0 requires the use of a geographically-aware load balancer or DNS failover solution that prefers to serve application requests using the primary data center, despite the fact that the Appliance instances in the hot standby data center are active and able to serve the requests.
+
+The application tier remains unaware of the locality of the primary data node, whereas the data layer resolves the location of the primary node (the only node that receives application queries). Additionally, the current implementation does not perform sharing or reading from a secondary replica, so all read and write activities are served by the primary node. For example, if the data centers for the geo cluster are in California (CA) and Texas (TX) and the ones in TX are active, any request serviced by the CA nodes at the application level requires data to then be written to a node in TX. This generally results in poorer performance due to the requests (and resulting round-trips) required to obtain the necessary data. Using a geographically-aware load balancer or DNS failover solution mitigates this performance issue.
```
chriskeyser: Additionally, the current implementation does not perform sharing or reading from a secondary replica, so all read and write activities are served by the primary node. => I think this gives away a little too much. can we just say `All read and write activities are served by the primary node`.

Also let's use Washington and Colorado in the example because TX and CA relate specifically to the AAA configuration, I'd like to make more ambiguous.

PR #1526, opened 2016-06-14
[update] New change password widget translations

Catografix:
Added Password history and widget translation code, renamed file to reflect contents


Catografix:
Preview OK

PR #1551, opened 2016-06-17
Sample code for logout action

nicosabena:
Added a sample code snippet for clearing the session cookie and redirecting to `/v2/logout`.


ntotten:
@jerriep Can you accept this if it works for you.

PR #1608, opened 2016-06-23
Pablo.02 custom login

volbap:
### Your checklist for this pull request

Please review the [guidelines for contributing](https://github.com/auth0/docs#contributing) to this repository.
- [x] Content conforms to our [Contributing Guidelines](https://github.com/auth0/docs#contributing-guidelines)
- [x] If applicable, you have added details to the [update feed](https://github.com/auth0/docs/tree/master/updates) 

If you are an Auth0 employee, after you submit this PR create and test a [Review App](https://github.com/auth0/docs#review-apps) to test your changes.
### Description
- Added `02-Custom-Login.md` file, containing a quickstart on how to implement Login (and optionally User Profile retrieval and Sign Up) easily without using the Lock widget, but using the Auth0 toolkit.
- Updated `index.yml` file, including this addition.

Thank you!

PR #1634, opened 2016-06-27
Pablo.03 session handling

volbap:
### Your checklist for this pull request

Please review the [guidelines for contributing](https://github.com/auth0/docs#contributing) to this repository.
- [x] Content conforms to our [Contributing Guidelines](https://github.com/auth0/docs#contributing-guidelines)
- [x] If applicable, you have added details to the [update feed](https://github.com/auth0/docs/tree/master/updates) 

If you are an Auth0 employee, after you submit this PR create and test a [Review App](https://github.com/auth0/docs#review-apps) to test your changes.
### Description
- Added `03-Session-Handling.md` tutorial.
- Updated `index.yml` accordingly.
- ⚠️ **Ideally, #1608 should be merged before this one.**

Thank you!

### DIFF
--- a/articles/native-platforms/ios-swift/03-session-handling.md
+++ b/articles/native-platforms/ios-swift/03-session-handling.md
@@ -54,7 +54,7 @@ Then, run `pod install`.
 
 > For further reference on Cocoapods, check [their official documentation](http://guides.cocoapods.org/using/getting-started.html).
 
-### 1. Store the user's idToken
+### 1. On Login: Store the user's idToken
 
 > The `idToken` is a string representing, basically, the user's [JWT token](https://en.wikipedia.org/wiki/JSON_Web_Token).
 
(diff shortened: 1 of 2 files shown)


articles/native-platforms/ios-swift/03-session-handling.md
```diff
@@ -0,0 +1,180 @@
...
+Once the user has logged in, you get both an `A0Profile` and an `A0Token` object, coming as follows:
+
+```swift
+let controller = A0Lock.sharedLock().newLockViewController()
+controller.onAuthenticationBlock = { profile, token in
+	// do something with profile and token
+}
+```
+
+> Even though `profile` and `token` come as optional values, the [A0LockViewController](https://github.com/auth0/Lock.iOS-OSX/blob/master/Lock/UI/A0LockViewController.h) documentation guarantees that both of them will only be `nil` iff login is disabled after sign up. We assume that's not our case; therefore, we will perform force-unwrap to get their values.
```
sebacancinos, line 72: "guarantees that both of them will only be `nil` **_iff_** login"

volbap, line 72: [iff == if and only if](https://en.wikipedia.org/wiki/If_and_only_if)

sebacancinos, line 72: I guessed as much, but I don't think you should use that kind of 'code expressions' on a coloquial text. Especially in this case where it could be misunderstood as a typo. It leads to confusion. Better to say something along the lines of: 

"guarantees that both of them will only be nil when login is disabled"

sebacancinos:
+1 @ntotten

volbap:
⛔️ Do not merge yet. Gotta fix some stuff.

auth0build:

PR #1661, opened 2016-06-29
Updates ruby and gems versions - Fixes a title for clarification

fleko:


fleko:
Created in error. Closing it.

PR #1697, opened 2016-07-01
Doc for GitHub Deployments extension

mpaktiti:
- New doc for GitHub Deployments extensions
- Update extensions index with new doc


articles/extensions/github-deploy.md
```diff
@@ -0,0 +1,116 @@
...
+> You can find details on how to configure a webhook in [GitHub documentation](https://developer.github.com/webhooks/creating/).
+
+You are now ready to start deploying.
+
+
+## Start deploying
+
+Once you setup the webhook using the displayed information you are ready to start committing to your repository. With each commit you make in your configured GitHub repository, if changes were made in the `rules` or `database-connection` folders the webhook will call the extension to start the deployment. Note that we always do a full redeploy of all contents to make sure we keep a consistent state.
+
+The __Deploy__ button lets you manually deploy the rules and database connection scrips you already have in your GitHub repository. This is useful if you already have a repository filled with scripts and you want to deploy them once you setup the extension. It can also come in handy in case you accidentally delete some things in Auth0 and you want to redeploy the latest version of your repository again.
```
sandrinodm: scripts

articles/extensions/github-deploy.md
```diff
@@ -0,0 +1,116 @@
...
+__set-country.json__
+```javascript
+{
+  "enabled": false,
+  "order": 15,
+  "stage": "login_success"
+}
+```
+
+Having multiple rules with the same order is not allowed, so make sure you don't have any collisions. Also, keep in mind that for the `user_registration` and `login_failure` stage only a single rule is allowed.
```
sandrinodm: Maybe add a note with best practices? Like: "We suggest you to create a json file for every rule and explicitly assign an order to each rule. If then you leave enough space between rules it will be easy to move them around without causing collisions. Eg:

 Order 10 - log_user.js
 Order 20 - set_country.js
 Order 30 ...
"

sandrinodimattia:
Could we also add somewhere in the docs that it's not supported to change the rule stage once you deployed a rule? If you really need to change it, you can create a rule with a new name and the other stage then delete the original rule.

auth0build:

PR #1708, opened 2016-07-05
Clarify Web API Signature Verification

jerriep:
Clarify that the JWT middleware is automatically downloading the keys and verifying the signature.


auth0build:

MicahZoltu:
In my case (I was confused by the previous text) I think just having a comment in the code in step 5 around `Authority` indicating that the passed in value will be used to get the public keys used to validate the token's signature would have helped me tremendously.  ASP.NET is doing magic under the hood to automatically fetch the private keys and I didn't realize that so the code looked like it wasn't actually validating against any sort of shared secret or public/private key.

PR #1710, opened 2016-07-05
Update lock v10 docs to rc.2

gnandretta:


auth0build:

PR #1712, opened 2016-07-05
[update] Metadata-in-rules - Added client_metadata samples, overall edits

Catografix:
New client_metadata feature added to rules doc.

Need an engineer to review code samples, I simply copied existing samples and replaced "user" with "client".

Does the rules `auth0` object include a `clients` property, as in:
`auth0.clients.updateClientMetadata(client.client_id, client.client_metadata)`

Also, might help to have real-world key/value examples in the sample code instead of just `key1` and `valuefor Key` but I do not know of actual scenarios to pull these from.

### DIFF
--- a/articles/rules/metadata-in-rules.md
+++ b/articles/rules/metadata-in-rules.md
@@ -1,10 +1,39 @@
+---
+description: This article explains how to work with metadata in Rules code. 
+---
+
 # Metadata in Rules
 
-Auth0 allows you to store data related to each user that has not come from the identity provider. This is known as metadata. There are two kinds of metadata, **user_metadata** and **app_metadata**. You can read about both of these at [app_metadata and user_metadata](/api/v2/changes#app-_metadata-and-user-_metadata).
+Auth0 allows you to store data that has not come from the identity provider. This is known as metadata. 
 
 This article explains how to work with metadata in [Rules](/rules) code. 
 
-Each sample rule in this article assumes the user is represented by the following JSON:
+* [Metadata types](#metadata-types)
+* [User object metadata](#user-object-metadata)
+  * [Read](#read)
+  * [Update](#update)
+  * [Delete](#delete)
+* [Client object metadata](#client-object-metadata)
+* [Considerations](#considerations)
+
+
+## Metadata types
+
+There are two types of metadata in the user object related to each user: **user_metadata** and **app_metadata**. You can read about both of these at [app\_metadata and user\_metadata](/api/management/v2/changes#app_metadata-and-user_metadata).
+
+Also available is **client_metadata**, where you can store information that is related to your client application.
+
+| Object | Metadata | Use |
+| :--- | :--- | :--- |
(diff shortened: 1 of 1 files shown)


auth0build:

articles/rules/metadata-in-rules.md
```diff
@@ -311,42 +338,131 @@ The resulting user is:
 }
 ```
 
-### Considerations
+## Client object metadata
```
Catografix, line 341: It is the samples below that need a code review.

articles/rules/metadata-in-rules.md
```diff
@@ -311,42 +338,131 @@ The resulting user is:
...
+
+```js
+function(client, context, callback){
+  client.client_metadata = client.client_metadata || {};
+  // update the client_metadata that will be part of the response
+  client.client_metadata.key1 = client.client_metadata.key1 || [];
+  client.client_metadata.key1.push('admin');
+
+  // persist the client_metadata update
+  auth0.clients.updateClientMetadata(client.client_id, client.client_metadata)
```
Catografix: Does auth0.clients exist? I merely copied from auth0.users example.

articles/rules/metadata-in-rules.md
```diff
@@ -311,42 +338,131 @@ The resulting user is:
...
-The following is accepted, however.
+### Update
+
+To update a property, use the `push' method.
+
+```js
+function(client, context, callback){
+  client.client_metadata = client.client_metadata || {};
+  // update the client_metadata that will be part of the response
+  client.client_metadata.key1 = client.client_metadata.key1 || [];
```
ntotten: this is confusing. In the above example `key1` is a string, but now it is an array. Was that intended?

articles/rules/metadata-in-rules.md
```diff
@@ -311,42 +338,131 @@ The resulting user is:
...
+  "key1": "newValue"
+  }
+...
+}
+```
+
+### Delete
 
-"color" : "light.blue"
+To delete a property, set the value to `null`.
```
ntotten: what about setting it to `undefined`? Will that have the same result? We should specify the behavior just to be sure.

articles/rules/metadata-in-rules.md
```diff
@@ -311,42 +338,131 @@ The resulting user is:
...
 
-If you use a key field name with a dot in it you will get an Internal Server (500) error.
+The resulting client is:
 
-A few workarounds are listed below:
+```js
+{
+...
+  "client_metadata": {
+  "key1": ""
```
ntotten: is that really the result? You set it to `null` before, but now it is set to an empty string? Those are very different values.

Catografix: I agree I dont know much about code. Will need feedback from a coder to fix.

Catografix:
@alex Can you have someone review this doc?

Catografix:
Katie will incorporate these changes into a rework of all metadata docs.

PR #1736, opened 2016-07-07
Fixed python snippets for API quickstart

cabralmartin:
- Fixed python snippets for API quickstarts


auth0build:

PR #1746, opened 2016-07-08
[edit]: removed note per Chris Keyser

ndelgado:
Removed the following note per Chris Keyser:

> When using Auth0 on your own environment (a dedicated instance of Auth0 running on your data center or in your own cloud), it is possible to connect directly to AD without using the Connector.
> Chris said: we should remove that from external docs. I am worried about the complexity of the implementation and debugging it. Its hard to debug, and only works on the root tenant authority.


auth0build:

PR #1749, opened 2016-07-11
Added requirement for Connection string

mutablestudio:
The example will not work if the developer does not specify a Connection string - at least for email/pwd auth. This is not in the current doc and is also not in the generated, downloadable Android example. I could not get my app nor the example to connect without adding the setConnection auth parameter. BTW it's also not clear in the documentation link below what setConnection() actually does:

https://auth0.com/docs/libraries/lock-android/sending-authentication-parameters


auth0build:

PR #1752, opened 2016-07-11
Fix API URLs in user impersonation doc

thameera:
${account.namespace} already contains `.auth0.com` suffix.


auth0build:

PR #1766, opened 2016-07-11
Mention how to detect RO login

rolodato:
Otherwise it's not clear how to handle this situation with redirect rules.


auth0build:

PR #1768, opened 2016-07-12
Updated to clarify we mean custom social connections

ashishdasnurkar:
In option #3 lock customization to add new connection button, just clarified that by Custom we mean Custom Social connection. We had a ticket where user misunderstood it for custom database connection


auth0build:

PR #1774, opened 2016-07-12
Updated java hs256 snippet and minor bugfix on node hs256

cabralmartin:
- Updated java snippet for hs256
- Minor fix on node hs256 snippet


auth0build:

PR #1777, opened 2016-07-13
Updated migration page with latest migration

mcastany:
Updated migration page with latest migration


auth0build:

PR #1781, opened 2016-07-13
Added warning on Katana cookies issue [review]

nicosabena:
Microsoft's OWIN implementation for ASP.Net has a bug that clears authentication cookies on occasions (not always, which makes it difficult to troubleshoot). This change includes a workaround.

Please see the changes. It's currently a note at the end of the tutorial, but adding it as an additional step might be a better idea. Let me know if you want me to modify this.


auth0build:

PR #1784, opened 2016-07-13
Remove Notify Dashboard Owners action

nicosabena:
**Notify Dashboard Owners** is not available as an action for the 1st level Brute Force Protection. 
I'm removing it from the docs so that we don't confuse users.


auth0build:

PR #1785, opened 2016-07-13
Include a link in ID token page

lobo:


auth0build:

PR #1788, opened 2016-07-14
Correction of links

NeelBhatt:


auth0build:

mpaktiti:
Hello @NeelBhatt 

The links are actually correct the way they are, you can verify it by navigating to the docs:
https://auth0.com/docs/connector
https://auth0.com/docs/email

There are some transformations that take place during the build, so the URLs end up correct, even though they may not look like it. This PR would result in breaking them so I'm going to close it. 

Thank you for your time.

NeelBhatt:
Okay thanks for clarification :+1: :)

PR #1794, opened 2016-07-14
Add details on how to generate API token

angelaLazar:
https://trello.com/c/u3QGFdLl/15-management-api-explore-users-need-to-us
e-global-client-id-and-global-client-secret-to-generate-api-token


auth0build:

PR #1797, opened 2016-07-15
Update CONTRIBUTING.md

camilosampedro:
Date is more appropriate than just year in this case; formatting example shows year, month and day together.


auth0build:

PR #1824, opened 2016-07-15
Passwordless email - new parameter

Catografix:
If you call the api to change the email, the template will be executed and, in case the customer wants to send something different, they can use `operation == 'change_email'` to define a block that will be only shown for change email actions

Also added Copilot instructions and updated screenshot.


auth0build:

Catografix:
Preview OK:
https://auth0-docs-content-pr-1824.herokuapp.com/docs/connections/passwordless/spa-sms
https://auth0-docs-content-pr-1824.herokuapp.com/docs/connections/passwordless/spa-email-code

PR #1830, opened 2016-07-18
New ASP.NET Core MVC Quickstart

jerriep:
Adds ASP.NET Core MVC Quickstart in the new format for server applications.

Missing is "Linking Accounts". This is quite challenging so I need to spend more time on this, but I want to get this out there.


auth0build:

mpaktiti:
Hi @jerriep Would you mind resizing the images to max 750px wide? Thanks!

articles/server-platforms/aspnet-core/01-login.md
```diff
@@ -179,88 +93,71 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerF
...
 }
 ```
 
-The `Login` method will call the `GenerateLockContext` extension method which will create the correct `state` and `nonce` parameters and set the correct Cookies for the OIDC middleware to function correctly. It will return a `LockContext` parameter which you can pass along to the View.
-
-Also notice that the `OpenIdConnectOptions` are injected into the Controller, and that it is passed along to the `GenerateLockContext`. This method needs to retrieve the configuration settings for the OIDC middleware so that the correct `state` can be generated. It is for this reason why we had to register the `OpenIdConnectOptions` with the dependency injection.
-
-### 9. Setting up the View
-
-For the Login screen you can create a Razor view and embed the code for Lock. You can head over to our [Lock Library page](https://auth0.com/docs/libraries/lock#start-using-auth0lock) and confgure Lock the way you want, and then copy and paste the code it generated into your Login page.
```
mpaktiti, line 232: The link should be relative and `confgure` has a typo.

articles/server-platforms/aspnet-core/04-storing-tokens.md
```diff
@@ -0,0 +1,90 @@
...
+
+            return Task.FromResult(0);
+        }
+    }
+});
+```
+
+The `access_token` will now be stored as a claim called "access_token", so to retrieve it inside a controller you can simply use `User.Claims.FirstOrDefault("access_token").Value`
+
+::: warning-banner Beware of your cookie size
```
mpaktiti, line 88: This looks like that:

image

It looks like a bit too much, and the `ClaimsIdentity` is barely visible.

jerriep, line 88: Fixed

articles/server-platforms/aspnet-core/01-login.md
jerriep, line 232: The text you referenced here is the deleted text.

I fixed it in the file 02-login-embedded-lock.md though

jerriep:
@mpaktiti Changes made based on your feedback

PR #1839, opened 2016-07-19
Move Lock Obj-C API to a separate document

lobo:
See my previous PR


auth0build:

PR #1859, opened 2016-07-20
Erase old docs links

lobo:


auth0build:

lobo:
They are nowhere. Were outdated and removed. Hence now the links were broken.

PR #1862, opened 2016-07-20
Update SSO on Mobile Apps

lobo:


auth0build:

chenkie:
We should change the copy a bit here:

Lock allows you to easily implement singe sign-on (SSO). Single sign-on is a mechanism that allows a user to use the same credentials between multiple applications. If you have two different applications and you have a user that is already authenticated app A, single sign-on allows that user to automatically be authenticated in app B as well. The experience is similar to what happens with Facebook and it's companion Messenger app, as well as Foursquare and Swarm.

lobo:
Change applied, thanks :)

PR #1863, opened 2016-07-20
Move SSO in iOS to its own section

lobo:


auth0build:

lobo:
Makes sense with the previous PR applied

PR #1868, opened 2016-07-20
Move TouchID iOS docs to its own section

lobo:
It already existed, just moved the information. Which was repeated.


auth0build:

PR #1871, opened 2016-07-21
Fixes to Web API QS based on user feedback

jerriep:
Updates to ASP.NET Core Web API QS:
- Clarifies that people can use HS256
- Updates requirements

The `.vscode/spell.json` is just a file used by VS Code for spell checking. This does not affect anything in the website


auth0build:

PR #1888, opened 2016-07-22
Fixing SEO crawling issues and adding redirections

ggonzaleza:
Redirected indexed sites that were returning 404 errors.
Fixed some broken links.


auth0build:

PR #1889, opened 2016-07-25
Clarify IDP access token still exists in rules

twistedstream:
A lot of customers have asked about the IDP access token migration and whether or not it will still be available in rules, and the answer is yes.


auth0build:

PR #1900, opened 2016-07-26
Add Lock iOS: Password Reset

lobo:

### DIFF
--- a/articles/libraries/lock-ios/password-reset-ios.md
+++ b/articles/libraries/lock-ios/password-reset-ios.md
@@ -36,7 +36,7 @@ params[A0ParameterConnection] = @"Username-Password-Authentication"; // Or your 
 
 ### Swift version
 
-```
+```swift
 let client = A0Lock.sharedLock().apiClient()
 let params = A0AuthParameters.newDefaultParams();
 params[A0ParameterConnection] = "Username-Password-Authentication";// Or your configured DB connection
--- a/articles/libraries/lock-ios/password-reset-ios.md
+++ b/articles/libraries/lock-ios/password-reset-ios.md
@@ -1,0 +1,52 @@
+---
+toc_title: Password Reset 
+description: All you need to know about password reset in iOS
+---
+
+# Lock iOS: Password Reset
+
+## Introduction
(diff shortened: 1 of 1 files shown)


articles/libraries/lock-ios/password-reset-ios.md
```diff
@@ -0,0 +1,52 @@
+---
+toc_title: Password Reset 
+description: All you need to know about password reset in iOS
+---
+
+# Lock iOS: Password Reset
+
+## Introduction
+You can allow the user to reset the password for any database connections. 
+We strongly suggest reading [this article](/articles/connections/database/password-strength.md) before implementing this.
```
mpaktiti: @lobo The valid URL is `/connections/database/password-strength` (we omit `www.auth0.com/docs` and add the rest; we also omit the `md` extension)

lobo: Modified, thanks! 👍

articles/libraries/lock-ios/password-reset-ios.md
```diff
@@ -0,0 +1,52 @@
...
+If you implement your custom UI you need to send a password reset email to a user using `A0APIClient`.
+
+## Important considerations
+Please note that passwords can only be changed for users signing in using database connections. If a user is signing in with a social or enterprise connection, their password would need to be reset in those platforms.
+
+## Code snippets
+
+### Objective-C version
+
+```objective-c
```
mpaktiti: @lobo I can't see any highlighting https://auth0-docs-content-pr-1900.herokuapp.com/docs/libraries/lock-ios/password-reset-ios#objective-c-version I think the valid one is `objc` (not really sure but I've seen it used)

lobo: `objc` seems to do the trick.

hzalaz:
We should also add how to reset password with https://github.com/auth0/Auth0.swift, we can leave the old way too (the one that uses Lock v1).

lobo:
Maybe in a future PR? @chenkie what do you think?
That's all I have on my end at the moment

chenkie:
Future PR for the Swift version works

PR #1901, opened 2016-07-26
Add Lock iOS: Close Lock UI

lobo:


auth0build:

PR #1921, opened 2016-07-28
Update head.md

ValeriyRadchenko:
dependency has removed 
auth0-variables.js causes the 404 error.


auth0build:

PR #1922, opened 2016-07-28
Update 01-login.md

ValeriyRadchenko:
Popup example has removed.
We can't call Lock 10 "show" method with option "{popup: false}". Lock 10 is needed to init with option "auth : { redirect: false }".


auth0build:

PR #1930, opened 2016-07-28
Adding info on using the API to create a rule & changing a stage of a rule

angelaLazar:
Document different Rule signatures when using the API to create a rule.

https://trello.com/c/5e8EcvCO/14-rules-have-different-signatures


auth0build:

PR #1933, opened 2016-07-29
Changed quickstart redirects to avoid problems with multiple articles per platform

nkohari:
This patch adds a regex check to the routes used for legacy quickstart redirects to avoid accidentally redirecting article URLs back to the corresponding platform URLs, while simultaneously redirecting legacy routes correctly.

### DIFF
--- a/redirect-urls.json
+++ b/redirect-urls.json
@@ -640,19 +640,19 @@
     "to": "/quickstart/native/:platform"
   },
   {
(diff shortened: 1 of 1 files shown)


auth0build:

Catografix:
In my local tests this still redirects links like https://auth0.com/docs/quickstart/spa/angular2/07-authorization back to the start page https://auth0.com/docs/quickstart/spa/angular2.

PR #1934, opened 2016-07-29
Seed Project Path Update

mehreencs87:


auth0build:

PR #1939, opened 2016-07-29
Go through docs commit history and find the last commit to the quickstart

sildur:
Go through docs commit history and find the last commit to the Angular 1.x quickstart that was before the multistep got put in. Make sure it links to Lock v9 using auth0-angular < v5. The reversion might need to be manual (ie delete multistep stuff and put the old files back in place). The doc will be a non-multistep quickstart and that’s what we want for now


auth0build:

articles/client-platforms/angularjs.md
```diff
@@ -0,0 +1,329 @@
...
+  - quickstart
+snippets:
+  dependencies: client-platforms/angularjs/dependencies
+  head: client-platforms/angularjs/head
+  setup: client-platforms/angularjs/setup
+  use: client-platforms/angularjs/use
+alias:
+  - angular
+seo_alias: angularjs
+---
```
alemata: @sildur you don't need all this fields anymore, just `title` and `description`

alemata:
👍 @chenkie @mpaktiti, there is only one comment, I don't know if that apply to that doc

articles/client-platforms/angularjs/index.yml
```diff
@@ -0,0 +1,19 @@
...
+tags:
+  - quickstart
+snippets:
+  dependencies: client-platforms/angularjs/dependencies
+  head: client-platforms/angularjs/head
+  setup: client-platforms/angularjs/setup
+  use: client-platforms/angularjs/use
+seo_alias: angularjs
+articles:
+  - login
```
chenkie: this should be:

```
articles:
  - 01-login
```

PR #1940, opened 2016-07-29
Fix API v2 URL in Migrations doc

thameera:


auth0build:

PR #1942, opened 2016-07-29
Update SAML config with Hosted Graphite provider data

zachd:
Hi there,

This change adds an entry for [Hosted Graphite](https://www.hostedgraphite.com) and the appropriate SAML configuration and Callback URL.

Please let me know if there is a more appropriate place to put this as well. I can provide a set of temporary credentials for testing the integration.

Kind regards,
Zachary - Hosted Graphite.

### DIFF
--- a/articles/saml/saml-apps/index.md
+++ b/articles/saml/saml-apps/index.md
@@ -26,6 +26,22 @@ url: /saml-apps
 
 ---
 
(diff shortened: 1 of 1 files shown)


auth0build:

ntotten:
Looks good! Thanks for the contribution.

PR #1944, opened 2016-07-29
Screenshot was not working because .PNG instead of .png

angelaLazar:
Had to rename the file for git to pickup on the change


auth0build:

PR #1962, opened 2016-07-31
fixed casing and example for signUpLink

darkyen:
The docs mentioned the property `signUpLink` as `signupLink` and the example usage showed it being passed in `.show`.


auth0build:

PR #1965, opened 2016-08-01
Removing link to doc which isn't in v10 docs yet

jeffreylees:
Customers following a link to a document which no longer exists in the new Lock 10 documents, at least not yet.


auth0build:

PR #1968, opened 2016-08-01
Added Signing key rollover notice

nicosabena:
Microsoft notifies customers of signing certificate rollover in Azure AD. This section clarifies that Auth0 users shouldn't be concerned about this.


auth0build:

PR #1976, opened 2016-08-02
Fix "caching expensive resources" example

davidnortonjr:
The code example here is saying that if global.db is there, to use it, but the boolean logic was backwards.


auth0build:

mpaktiti:
Hi @davidnortonjr This fix has been applied. Thank you! 👍

PR #1980, opened 2016-08-02
Fixed title duplication and duplicated content

ggonzaleza:
Fixed duplicated titles.
Fixed duplicated content between articles that were generated dynamically on /authenticate:
https://auth0.com/authenticate/sap-api
https://auth0.com/authenticate/sap-odata


auth0build:

ggonzaleza:
@ntotten this PR solves the duplicated content I told you about.
@alejofernandez guided me on how to solve it. He also mentioned that a nicer solution would be to enable an exclusion list in the future, so we can filter the dynamic content from our end.

PR #1986, opened 2016-08-03
Adding info on twitter attributes and access tokens

angelaLazar:
Updating the twitter docs to reflect that twitter permissions are managed on the app level and cannot be updated by the Auth0. Adding an optional step about how to generate an access_token and access_token_secret from Twitter.

https://trello.com/c/cEvhCQWa/36-update-twitter-docs


auth0build:

PR #1987, opened 2016-08-03
Metadata - Overall edits

Catografix:


auth0build:

Catografix:
Preview OK: https://auth0-docs-content-pr-1987.herokuapp.com/docs/metadata
There shouldn't be any problems this time

articles/metadata/apiv2.md
```diff
@@ -91,7 +96,7 @@ The `PATCH` call to the API would therefore be:
 }
 ```
 
-The user's profile now appears as follows:
+The user's profile will now appears as follows:
```
DylanMeeus: There was a small typo in the original: `will now appear as follows` instead of "will now appears". Maybe you could fix that in this PR as well? :)

articles/metadata/apiv2.md
```diff
@@ -108,9 +113,11 @@ The user's profile now appears as follows:
 }
 ```
 
-**Note:** Only properties at the root level are merged in to the object. All others replace the existing value. The following example demonstrates an example of this merging.
+### Merging 
+
+Only properties at the root level are merged into the object. All lower-level properties will be replaced.
 
-If we were to add the user's work address as an additional inner property, we have to send over the entire address block. Because the `addresses` object is a root-level property, it will be merged into the final JSON object representing the user, but its *properties* will not.
+For example, to add a user's work address as an additional inner property, you would have to include the complete contents of address property. Since the `addresses` object is a root-level property, it will be merged into the final JSON object representing the user, but its sub-properties will not.
```
DylanMeeus: "..the complete contents of the `address` property" reads better than "..the complete contents of address property" in my opinion. Just a suggestion though.

articles/metadata/index.md
```diff
@@ -27,29 +28,29 @@ Suppose the following data is stored for a user with the email address "jane.doe
...
 console.log(user.app_metadata.hobby); // "full"
 ```
 
-> **Note**: With APIv1, all metadata was stored in the `metadata` field. Data stored in this field will now be available under `app_metadata`.
+**NOTE**: With APIv1, all metadata was stored in the `metadata` field. Data stored in this field is now available under `app_metadata`.
 
 ### Naming Metadata Fields
 
-Metadata field names must not contain a dot. For example, use of the following returns a Bad Request (400) error:
+Metadata field names must not contain a period. For example, use of the following returns a Bad Request (400) error:
```
DylanMeeus: This is nitpicking it a little bit, but `dot` is not wrong in this case. A period is, as far as I'm aware, a synonym for the full stop. Thus it stands for a _dot_ at the end of a sentence. In a metadata field, I would consider _dot_ to be correct, as we are not dealing with a sentence. <- period :-)

articles/metadata/rules.md
```diff
@@ -51,19 +47,17 @@ function(user, context, callback){
...
-
   ...
-
 }
 ```
 
 ## Updating Metadata
 
-All [rules](/rules) have available an `auth0` object (which is an instance of the [node-auth0 SDK](https://github.com/auth0/node-auth0) that is capable of calling the Auth0 Management API v2). The `auth0` object is preconfigured with the necessary permissions to update users.
+All rules have available an `auth0` object (which is an instance of the [node-auth0 SDK](https://github.com/auth0/node-auth0)) that is capable of calling the Auth0 Management API v2. The `auth0` object is preconfigured with the necessary permissions to update users.
```
DylanMeeus: I think you could drop the 'available' from the first sentence. `All rules have an ..`

DylanMeeus:
I'm not an admin, but the overal content of the PR looks good to me. I did leave a few small comments here and there - but admittedly I was nitpicking a bit.  👍

Catografix:
@DylanMeeus thanks for the feedback.

PR #1993, opened 2016-08-03
removed unnecessary requirement. Fix #1981

### LINKED ISSUE #1981: VanillaJS Quickstart - Unnecessary system requirement
I don't think JQuery is a necessary system requirement for the vanilla js quickstart as it is not used in the code. It might be confusing for users and in any case if it is needed we should refer to JQuery Quickstart.
vanilla-quickstart-1

chenkie:


auth0build:

PR #1999, opened 2016-08-03
Clear up about refresh token expiration

angelaLazar:
Make it clear that refresh_token’s never expire, just the resulting
id_token expire


auth0build:

PR #2003, opened 2016-08-03
Update angular version, check code, make it more similar to samples

sildur:
Tasks done:
- Raise angular to 1.5.8
- Check all code snippets to make sure they work
- Make some changes in the snippets to bring them closer to the samples


auth0build:

PR #2007, opened 2016-08-04
Provide example in Lock 10 usage

lobo:
About how to retrieve a token with the profile data


auth0build:

PR #2010, opened 2016-08-04
Updating info about current unlinking behavior

jeffreylees:


auth0build:

PR #2015, opened 2016-08-04
Note that Azure AD can cause large token with 'openid profile'

angelaLazar:
Add to warning with `openid profile` about Azure AD as this seems to
cause the most issues for customers.

https://trello.com/c/o5p7Ls6g/6-azure-ad-returns-an-access-token-that-is
-long-and-causes-issues-if-someone-uses-openid-profile
https://auth0.zendesk.com/agent/tickets/6818
https://auth0.com/forum/t/id-token-is-too-large/3116/2


auth0build:

PR #2016, opened 2016-08-05
Refresh token

ValeriyRadchenko:
refresh token example has been added.


auth0build:

PR #2020, opened 2016-08-06
Add a section to Lock general doc about multiple clients/connections

angelaLazar:
https://trello.com/c/xZ7X5wOL/26-documentation-for-custom-login-to-speci
fy-connections-on-a-per-app-basis


auth0build:

PR #2023, opened 2016-08-08
Update appliance differences

sandrinodimattia:
Update the differences overview with new features that were added in the cloud but are not available in the appliance yet


auth0build:

PR #2029, opened 2016-08-08
Adding pointer to Lock language dictionary options

jeffreylees:


auth0build:

PR #2031, opened 2016-08-08
Clarify which API is being referenced

angelaLazar:
Fixing found instances of the Management API being referred to as just
"API v1" or "API v2", this is confusing since we have multiple APIs
(Management and Auth)

https://trello.com/c/yrQUB5xN/71-management-api-should-never-be-refered-
to-as-just-api-v1-or-api-v2-this-is-confusing-since-we-have-multiple-api
s-management-and-


auth0build:

khorne3:
@angelaLazar Can you please rebase to handle the merge conflicts?

angelaLazar:
@khorne3 should be good to go now

PR #2035, opened 2016-08-09
Edit extension lifecycle doc

mpaktiti:
- Replace `/.webtask/on-install` with `/.extensions/on-install`
- Replace `/.webtask/on-uninstall` with `/.extensions/on-uninstall`
- Add `/.extensions/on-update`
- Add note on `createClient` and crons


auth0build:

PR #2037, opened 2016-08-09
Extensions typos

Annyv2:


auth0build:

chenkie:
I think we should actually go without any punctuation at the end for these :)

Annyv2:
Fixed, could you review? @chenkie

PR #2058, opened 2016-08-10
Adding security fundamentals docs

angelaLazar:
This is a work in progress for creating various docs on fundamental
security topics with Auth0. Not ready to be merged until more topics are completed and reviewed.

https://trello.com/c/ejATWJap/75-security-fundamental-docs


auth0build:

articles/security/common-threats.md
```diff
@@ -0,0 +1,28 @@
+# Common Threats and How to Prevent Them
+
+## Man-in-the-Middle(MiM or MitM) Attacks
```
radekk: 1. There is a missing space between `...-Middle(MiM...`.
2. I would also get rid off using `MiM` acronym. Among the web, articles, security reports etc. `MitM` is the most common one.

articles/security/common-threats.md
```diff
@@ -0,0 +1,28 @@
+# Common Threats and How to Prevent Them
+
+## Man-in-the-Middle(MiM or MitM) Attacks 
+
+One type of threat is a Man-in-the-middle attack, sometimes called a bucket brigade attack. With this attack the attacker gets between two parties, where each party thinks they are interacting over a private connection, but it is actually being controlled by the third-party attacker. For this type of attack to succeed the attacker must create mutual authentication between both parties. 
+
+Usually MiM attacks involve the attacker using a WiFi router to intercept a user's communication. The user connects to the attacker's router then visits a website and logs in with their confidential credentials. The attacker saves that user's login credentials for which they can then use to impersonate the user.
```
radekk: We can add an information about another risk vector which is not only the credentials stealing but also all kind of attacks which allow to modify the data transmitted between the client and the server. In the wort case scenario it could result in the remote code execution, backdooring the system, installing malware etc. It could also lead to the compromise of other network assets as well.
As an example:
- https://duo.com/blog/out-of-box-exploitation-a-security-analysis-of-oem-updaters
- https://www.coresecurity.com/advisories/intel-driver-update-utility-mitm
- http://www.theregister.co.uk/2016/03/11/samsung_software_update_mitm_flaw_patched/
- http://arstechnica.com/security/2016/02/huge-number-of-mac-apps-vulnerable-to-hijacking-and-a-fix-is-elusive/

articles/security/common-threats.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Preventing MiM attacks
```
radekk: Let's change it to `MitM`

articles/security/common-threats.md
```diff
@@ -0,0 +1,28 @@
...
+
+To help defend against this type of attack it is important to use strong encryption and authentication between the client and the server. Using encryption the server authenticates the client's request by presenting a digital certificate, and only then can the connection be established. 
+
+For example, HTTPS uses the secure sockets layer (SSL) capability the browser to mask web traffic. To decrypt HTTPS, a man-in-the-middle attacker would have to obtain the keys used to encrypt the network traffic.
```
radekk: We should mention here that keys used to encryption is not always the case to leak the data from the encrypted transmission. There were vulnerabilities in SSL/TLS that allowed to attack those protocols like:
- POODLE attack (https://blog.qualys.com/ssllabs/2014/10/15/ssl-3-is-dead-killed-by-the-poodle-attack)
- DROWN attack (https://blog.qualys.com/securitylabs/2016/03/01/drown-abuses-ssl-v2-to-attack-rsa-keys-and-tls)

Additionally let's add here some recommendations like ciphers and protocol versions which are safe to use, together with recommendations for SSL/TLS. @eugk what do you think about that?

articles/security/common-threats.md
```diff
@@ -0,0 +1,28 @@
...
+## Replay Attacks
+
+Another type of attacks is a replay attacks, also sometimes called a playback attack. With this type of network attack an attacker spies on a data exchange such as a login authentication, then the attacker takes this information to the receiver. This does not necessarily happen in real-time. With a replay attack the attacker gives the proof of identity by impersonating a user with stolen credentials.
+
+### Preventing Replay Attacks
+
+Replay attacks are usually avoided by using session tokens. But if these credentials are stolen from local storage (such as from an XSS attack), there are alternatives that you could use to avoid someone having infinite tokens:
+
+1. Set expirations for tokens
+2. Provide a way to blacklist tokens that have not been used (perhaps even a user)
```
radekk: Shouldn't we write about the compromised tokens instead of tokens that have not been used? It makes more sense to blacklist tokens that were compromised/leaked.

articles/security/store-tokens.md
```diff
@@ -0,0 +1,32 @@
+# Where to Store Tokens
+
+ This document explains how to securely store tokens used in token-based authentication. The following assumes a basic knowledge of JSON Web Tokens(JWTs). To learn more about JWTs see:
```
radekk: There is a missing space character `JSON Web Tokens(JWTs)`.

articles/security/store-tokens.md
```diff
@@ -0,0 +1,32 @@
...
+
+### Using Cookies
+
+You can also use cookies to store the JWT. The exact way to set a cookie depends on the client side language you are using.
+
+There are different options to control the lifetime of a cookie:
+
+* Cookies can be destroyed after the browser is closed (session cookies).
+* Implement a server side check (typically done for you by the web framework in use), and you could implement expiration or sliding window expiration.
+* Cookies can be persistent (not destroyed after the browser is closed) with an expiration.
```
radekk: We can add an additional point describing that:
- Cookies can be read by both the JavaScript and the server side code or only server side if `httpOnly` flag was set.

radekk (changes requested):

articles/security/common-threats.md
angelaLazar: Can you elaborate more on how to use ciphers and protocol versions and recommendations for SSL/TLS?

eugk: We can include some guidance about using the latest protocol (TLS 1.2) and the strongest ciphers your client base supports. It's well-documented here: https://wiki.mozilla.org/Security/Server_Side_TLS

angelaLazar:
@radekk I addressed your comments, can you re-review? Thanks

angelaLazar:
@radekk or @eugk Can you confirm if this is ready to be merged?

eugk:
Radek is out, but I'd say you should merge. We can always make updates later.

radekk (approved):

articles/security/token-exp.md
```diff
@@ -0,0 +1,41 @@
+# Token Expiration 
+
+This document states the expiration of different types of tokens that are referenced in Auth0 documentation. [Click here to learn more about the types of tokens.](/tokens)
+
+## Auth0 `access_token`
```
cabralmartin: If this refers to the Management APIv1 Access Token then let's make the explicit mention to it. `Auth0 access_token` is misleading as a title.

articles/security/token-exp.md
```diff
@@ -0,0 +1,41 @@
...
+
+For customer application APIs registered in Auth0, the validity of a delegation token issued for that target is governed by the JWT Expiration (seconds) value. This is set for each application in Applications > Settings.
+
+For APIs registered as Addons in Auth0, the validity period of the token will vary by individual Addon. The documentation available from the provider of any Addon API should be consulted for further information on tokens and expirations.
+
+[Learn more about Delegation Tokens](/tokens/delegation)
+
+## Auth0 Management APIv2 Token
+
+There is no specific expiration for a Management APIv2 token. A Management APIv2 token can be built programmatically, as desired, by a client.
```
cabralmartin: APIv2 tokens are good for 24hs (https://manage.auth0.com/#/apis/management/settings). Don't advice users to build the token themselves since it's an insecure practice and will be soon deprecated.

articles/security/token-exp.md
cabralmartin: `access_token`s issued for user defined APIs will have their expiration defined under their settings tab in the dashboard. Default is 24hs.

articles/security/token-exp.md
```diff
@@ -0,0 +1,41 @@
...
+
+[Learn more about Auth0 access tokens](tokens/access_token)
+
+## Identity Provider Access Tokens
+
+Since these tokens are created and managed by a third-party (such as Facebook, GitHub, etc) **the validity period for third-party tokens will vary by the issuing IdP**. If you believe these tokens have been compromised, you will need to revoke or reset them with the third-party that issued them. 
+
+[Learn more about IdP tokens](tokens/idp)
+
+## Auth0 `id_token`
```
cabralmartin: `Auth0 id_token` doesn't sound right. maybe `id_tokens issued by Auth0` ?

PR #2062, opened 2016-08-10
[edit]: added requirements per Chris

ndelgado:
Added requirements for appliance storage per Chris Keyser.


auth0build:

PR #2066, opened 2016-08-11
Lock index - overall edits - (edits PR 2020)

Catografix:
Reduced image size and moved to repo.
Fixed links to Browserify and webpack.
Created an include for the Version disclaimer - should be linked to from all the new v10 docs instead of editing each.
Moved disclaimer below image.
Changed first subhead to reflect page contents since it becomes the page title.


auth0build:

Catografix:
@jeffreylees Please review. If you like the version include, I can add it to the other docs, replacing the hard-coded in each.

articles/libraries/lock/index.md
```diff
@@ -143,37 +141,37 @@ var client = new Auth0({
 });
 ```
 
-## Start using Lock!
+## Start using Lock
 
 ${lockSDK}
```
Catografix, line 146: Where is the source for this ${lockSDK}? It needs editing!!

Catografix:
Preview OK: https://auth0-docs-content-pr-2066.herokuapp.com/docs/libraries/lock

jeffreylees:
Disclaimer include is good because at some point we'll change or remove it, might as well do that now. Could use the include method from here on in on version changes, and the same when v9 is EOL, on the v9 docs.

What needs edited in `${lockSDK}` ?

Catografix:
@jeffreylees 
There are many ways of integrating Auth0 into your site. 
Choose from the methods below to generate a preview and code snippets.
This is how it will look in a browser:
Copy this HTML/JS code into your website:

PR #2081, opened 2016-08-12
Impersonate endpoint call example mistake fixed

dmitriychernenko:
`response_type` and `state` parameters should be under `additionalParameters`


auth0build:

mpaktiti:
Hello @dmitriychernenko ! Thank you for the contribution! 👍

PR #2083, opened 2016-08-12
Fix android quickstart

hzalaz:


auth0build:

PR #2093, opened 2016-08-13
Rules stage

jeffreylees:
Removing references to stage, as stage has a default and other options are not available yet.


auth0build:

mpaktiti:
@jeffreylees Hey Jeff, can you rebase please? Thnx

PR #2103, opened 2016-08-16
Move postman images from cdn

jerriep:
Images for Postman was hosted on the CDN and broke for appliance customers. They have been moved to the docs media folder


auth0build:

PR #2105, opened 2016-08-16
Add new .NET SDK to update feed

jerriep:


auth0build:

PR #2126, opened 2016-08-18
Fix the authorization snippet for the ASP.NET Core Web API QS

jerriep:
Fixed a code snippet


auth0build:

PR #2131, opened 2016-08-18
Fixed migration mandatory date for idp access token

mcastany:


auth0build:

PR #2132, opened 2016-08-18
Update customization.md

ccorral:
While working with Auth0 Lock 10, noticed an issue with Lock not taking my `allowedConnections` preferences when storing it within the `options` object. After a little debugging, I placed the `allowedConnections` array into the `auth` object and it successfully rendered my connection preferences. I updated the customization.md file to reflect these changes.

### DIFF
--- a/articles/libraries/lock/v10/customization.md
+++ b/articles/libraries/lock/v10/customization.md
@@ -11,7 +11,6 @@ var lock = new Auth0Lock('clientID', 'account.auth0.com', options);
 ## Table of Contents
 
 **Display Options**:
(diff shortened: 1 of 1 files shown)


auth0build:

jeffreylees:
@ccorral the correct location for `allowedConnections` is top-level within the `options` object. What else can you tell us about what you were trying, about your project, or perhaps we could see how you were laying out `options`?

mpaktiti:
Hi @ccorral ! Since we cannot merge this PR I will go ahead and close it. However if this case still applies please register an [issue](https://github.com/auth0/docs/issues) so we can follow up properly. Thanks! 👍

PR #2135, opened 2016-08-18
Lock10 reorganization

jeffreylees:


auth0build:

articles/libraries/lock/_includes/_lock-toc.md
```diff
@@ -0,0 +1,21 @@
+## Lock: Table of Contents
+
+* [Getting Started With Lock][getting-started] - Methods of installation and an example app setup
```
mpaktiti: @jeffreylees Just out of curiosity, why do this:

```
[Getting Started With Lock][getting-started] - Methods of installation and an example app setup
[getting-started]: /libraries/lock#lock-10-installation
```

and not do this instead?

```
[Getting Started With Lock](#lock-10-installation) - Methods of installation and an example app setup
```

articles/libraries/lock/index.md
```diff
@@ -150,38 +114,25 @@ ${lockSDK}
...
-* [Customizing Lock to Your Needs][lock-customization]
-* [Embedding Lock (Instead of a Modal)][display-modes]
-* [i18n - Languages and Custom Text][i18n-notes]
-* [Popup Mode][popup-mode] - an advanced option (not recommended in most cases)
-* [Development notes][development-notes]
-* [Release process notes][release-process]
-* [Migration guide][legacy-migration-guide] for users of the legacy **Auth0 Widget**
-
-<!-- Variables-->
+<!--vars-->
```
mpaktiti, line 117: @jeffreylees Also (again out of curiosity), why have these vars here as well, since they already are at the `_lock-toc.md` file?

jeffreylees: Going to go ahead and use your suggestion of the non-variable link notation for the TOC include - since the links have to be redefined for use in the actual document, anyway, coming from an include. We probably wouldn't want them to share variables with the include, anyway, that could get messy later.

PR #2139, opened 2016-08-19
Make note of exceptions from LoginAsync

jerriep:


auth0build:

PR #2147, opened 2016-08-19
Java quickstarts updates

mpaktiti:
- Update seeds to replace configuration values
- Update configuration snippets
- Add not on pre-population of values and `issuer`


auth0build:

PR #2149, opened 2016-08-19
Update java spring security api doc

mpaktiti:
- Update setup snippet with latest properties
- Replace long list of bullets with matrix


auth0build:

PR #2162, opened 2016-08-23
Using 'none' for 'pkgType'.

volbap:
- Reverted to this setting in `00-introduction` document since the test from the last PR did not work out well.


auth0build:

PR #2169, opened 2016-08-24
Updating id_token doc for Lock 10 info

jeffreylees:
And correcting links


auth0build:

jeffreylees:
Will be adding a few more corrections

PR #2182, opened 2016-08-26
Instructions on adding additional claims mappings

nicosabena:


auth0build:

PR #2190, opened 2016-08-30
Fix Android snippets

mpaktiti:
The Custom Login page of the Android quickstart had a "funny" layout when you went to the page using the [link](https://auth0.com/docs/quickstart/native/android/02-custom-login) (if you navigated to the page using clicks from the docs home page, Custom Login was displayed ok).

Bug was due to some errors in code snippets markdown.

image


auth0build:

PR #2194, opened 2016-08-30
Merge rules metadata docs

mpaktiti:
We have two almost identical docs on manipulating metadata using rules:
- https://auth0.com/docs/metadata/rules
- https://auth0.com/docs/rules/metadata-in-rules

The content was merged, `/metadata/rules` was removed, and a redirect was created.


auth0build:

PR #2200, opened 2016-08-30
Remove extra single quotes from 05-linking-accounts guide

harenson:
Remove extra single quotes from `05-linking-accounts` guide.


auth0build:

PR #2206, opened 2016-08-31
Removing reference to Lock 10 preview release

jeffreylees:


auth0build:

PR #2208, opened 2016-09-01
Emi.lock update

tenmaster:
- Updated docs according to the requirements specified here. 
- This is not final yet but needs review. 
- Remaining work:
  
  > API reference
  > Logging

Also, Building Your Own UI was removed as Pablo suggested, but it can easily be ported from Custom Login tutorial.


auth0build:

sebacancinos:
+1

ntotten:
@chenkie Can you take a look at this.

PR #2213, opened 2016-09-01
updated php snippets for token authentication compliant with v4

glena:
This PR targets to update the PHP snippets for token decoding following the new API provided by the php-auth0 SDK v4.


auth0build:

PR #2214, opened 2016-09-01
Clarification as to the code snippet for MFA

topherMarie:
Clarifying that the generated code snippet is a 'rule' and how it can be updated.


auth0build:

PR #2216, opened 2016-09-02
Fix 404 error and duplicated title

ggonzaleza:
Fix 404 error


auth0build:

PR #2223, opened 2016-09-02
More screenshot fixes

angelaLazar:
Looks like GitHub can only handle about 25 images changes, the rest don't show up correctly. Will try to keep each PR to about 20 

https://trello.com/c/dleEYwrO/208-check-for-and-fix-screenshots-that-don-t-meet-requirements


auth0build:

PR #2226, opened 2016-09-05
 PHP api snippet formatting

mehreencs87:
removed the extra tabulation and extra slashes from 2nd step of php api tutorial snippet.


auth0build:

PR #2231, opened 2016-09-05
Fixed a broken link on the article about tokens

AmaanC:
Fixed the link to the article about "what to do once the user is logged in"  for social providers on the article about Tokens.


auth0build:

PR #2235, opened 2016-09-05
Make angularjs sample work with `npm`

alejofernandez:
This sample fails when following `npm` instructions because `lock` npm component is not packaged with the `dist` files nor has a `postinstall` script to build it, so the `/dist` folder doesn't exist and the sample fails.
With this change, `lock` library will be included from the CDN instead of including it from the `node_modules` folder.
Also added an `or` between `npm` and `bower` sections to make sure the user understands that just one of them is required.


auth0build:

PR #2236, opened 2016-09-05
Quick fix cdn link for auth0-lock

cristiandouce:
Fix link to cdn for lock v10


auth0build:

PR #2237, opened 2016-09-05
Normalize local storage item name for idToken

joseluisdiaz:
Founded try to pull the [`userProfile`](https://auth0.com/docs/libraries/lock#displaying-the-user-s-profile) in the example is using a different _localstorage_ _itemname_


auth0build:

PR #2238, opened 2016-09-06
Fixed typo in article about Protocols.

AmaanC:
Assuming that the `_` was unintentional, this fixes the typo.

If it was intentional, maybe it was meant to emphasize `JSON Web Tokens`, in which case it should be `_JSON Web Tokens_`.

Also removed what seems to be a broken link, and replaced it with the appropriate step number.


auth0build:

PR #2240, opened 2016-09-06
Replace broken link with correct numeric header

AmaanC:
The broken link was introduced in commit d3e9142.

The problem seems to have been caused by a mass replace for anchor links. There was one other broken link that I could find in `articles/protocols.md`, which I fixed in PR #2238.


auth0build:

PR #2246, opened 2016-09-06
deleted all the bullets from android custom ui

vmartynets:


auth0build:

PR #2251, opened 2016-09-07
updates to saas doc images

camimonse:
The initial problem was not reproducible because the images were already
located in the repo, however the images were not complying with the
contribution guidelines for screenshots. In these two:
- The browser was showing bookmarks.
- The browser was in incognito mode.

These two have been changed.


auth0build:

PR #2255, opened 2016-09-08
Update Azure Blob Storage ext doc

mpaktiti:
Add how-to-retrieve-info paragraph


auth0build:

Issue #1318, opened 2016-05-16
Images stored in external location (cloudup)


mpaktiti:
Some files reference images stored in external location (cloudup). Images should be linked in this repo using `![](/media/folder/image-name.png)`

Files and references in attached text file:
[image-cloudup.txt](https://github.com/auth0/docs/files/266072/image-cloudup.txt)

mpaktiti:
[jira](https://techaid.atlassian.net/browse/AUTH-3253)

PR #2260, opened 2016-09-08
Cleanup all Cloudup hosted images

angelaLazar:
This change moves all cloudup hosted images locally

Fixes https://github.com/auth0/docs/issues/1318


auth0build:

PR #2267, opened 2016-09-10
add note about auth0-extension-secret

tjanczuk:


auth0build:

PR #2275, opened 2016-09-12
Fix changePassword() wrong usage of username

cristiandouce:
- Fixes auth0/auth0.js#195


auth0build:

PR #2287, opened 2016-09-12
Fix node version reference

khorne3:


auth0build:

fredericmora:
We used node.js v 4 and we had several problems on Ubuntu. I don't think it's that easy. Can you please verify your connector and its installation script on node.js version 4? As far as we are concerned, your AD/LDAP connector is broken when trying to install on Ubuntu 16.04.

mpaktiti:
Hi @fredericmora ! Could you please register an Issue for this so we can follow up? Thanks! 👍

fredericmora:
I created an issue against the connector. The doc isn't the problem here, only a symptom. See https://github.com/auth0/ad-ldap-connector/issues/80.

PR #2290, opened 2016-09-12
Added endpoints doc

eugeniop:
Added documentation on all endpoints for cloud service. Some customers require a whitelist in their firewalls.


auth0build:

PR #2303, opened 2016-09-14
More doc images resizing

angelaLazar:
Some images just compressed not resized, trying to keep PRs to around 20 images


auth0build:

PR #2317, opened 2016-09-15
Fix domain and clientID values in vanillajs custom login

mauricioschneider:
Right now, the domain and clientID values in the vanillajs custom login example are in the wrong order. This PR fixes the issue.


auth0build:

PR #2325, opened 2016-09-15
List supported algorithms for XML-DSig

nicosabena:
Useful information when debugging SAML

### DIFF
--- a/articles/saml/saml-sp-generic.md
+++ b/articles/saml/saml-sp-generic.md
@@ -155,7 +155,6 @@ Check to make sure that the callback URL specified by your application in its au
 
 The **[http://samltool.io](http://samltool.io)** tool can decode a SAML assertion and is a useful debugging tool.
 
-# Additional information
 ## Supported algorithms for signatures
 
 We currently support the following algorithms for processing XML Digital Signatures:
--- a/articles/saml/saml-sp-generic.md
+++ b/articles/saml/saml-sp-generic.md
@@ -154,3 +154,25 @@ Be sure to check to make sure cookies and javascript are enabled for your browse
 Check to make sure that the callback URL specified by your application in its authentication request is listed in the **Allowed Callback URLs** field in the __"Settings"__ tab of the application registered in the Auth0 Dashboard.  (In dashboard, Click on __"Apps/APIs"__ link, then on the __"Settings"__ icon to the right of the application name.)
 
 The **[http://samltool.io](http://samltool.io)** tool can decode a SAML assertion and is a useful debugging tool.
+
+# Additional information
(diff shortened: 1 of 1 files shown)


auth0build:

articles/saml/saml-sp-generic.md
```diff
@@ -154,3 +154,25 @@ Be sure to check to make sure cookies and javascript are enabled for your browse
 Check to make sure that the callback URL specified by your application in its authentication request is listed in the **Allowed Callback URLs** field in the __"Settings"__ tab of the application registered in the Auth0 Dashboard.  (In dashboard, Click on __"Apps/APIs"__ link, then on the __"Settings"__ icon to the right of the application name.)
 
 The **[http://samltool.io](http://samltool.io)** tool can decode a SAML assertion and is a useful debugging tool.
+
+# Additional information
```
mpaktiti: This would look better without the h1 header. Just the `## Supported algorithms for signatures` looks fine, so can you please remove this line?

articles/saml/saml-sp-generic.md
```diff
@@ -154,3 +154,25 @@ Be sure to check to make sure cookies and javascript are enabled for your browse
...
+## Supported algorithms for signatures
+
+We currently support the following algorithms for processing XML Digital Signatures:
+
+* Canonicalization
+  * 'http://www.w3.org/2001/10/xml-exc-c14n#'
```
mpaktiti, line 163: The URLs should be like this: 
`[Canonical XML (XML-C14N)](http://www.w3.org/2001/10/xml-exc-c14n#)`
`[Canonical XML (XML-C14N) #WithComments](http://www.w3.org/2001/10/xml-exc-c14n#WithComments)`
and so forth

mpaktiti (changes requested):
Hello @nicosabena ! We need some updates on this, can you please have a look at the comments? Thanks!

articles/saml/saml-sp-generic.md
nicosabena: These are not really URLs, they are just URIs to identify an algorithm.
They are used directly in XML, like this:

```
<CanonicalizationMethod algorithm='http://www.w3.org/2000/09/xmldsig#enveloped-signature' ...
```

so it doesn't really make sense to format them as URL.

articles/saml/saml-sp-generic.md
nicosabena: ok.

articles/saml/saml-sp-generic.md
mpaktiti: ok then!

mpaktiti (approved):

PR #2326, opened 2016-09-15
Update 01-login.md

theRemix:
matches the snippet from here https://github.com/auth0-samples/auth0-angularjs-sample/blob/master/00-Starter-Seed/app.js#L30


auth0build:

PR #2339, opened 2016-09-19
Minor content updates to several docs for clarity

jeffreylees:

### DIFF
--- a/articles/libraries/lock/v10/i18n.md
+++ b/articles/libraries/lock/v10/i18n.md
@@ -15,7 +15,7 @@ var options = {
 };
 ```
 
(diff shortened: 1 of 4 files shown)


auth0build:

articles/libraries/lock/v10/i18n.md
```diff
@@ -12,9 +15,9 @@ var options = {
 };
 ```
 
-Where dict can be a string matching the name of the file in the `i18n` folder.
+Of course, `language` needs to be a string matching the name of the corresponding file in the `i18n` folder within Lock. You can look at existing language files to provide an example format to copy for any new languages you wish to add yourself, and if and when supported languages are added to Lock 10, they will be added to the `i18n` folder with new releases.
```
chenkie: I think we should drop "of course" here :)

PR #2349, opened 2016-09-19
Added mention to the now missing `$auth0` property

nicosabena:


auth0build:

PR #2350, opened 2016-09-20
Update URL

mpaktiti:
The ﻿⁠⁠⁠Implicit flow﻿⁠⁠⁠ hyperlink in this section is outdated https://auth0.com/docs/refresh-token#obtain-a-refresh-token
The link is https://auth0.com/docs/protocols#oauth-for-native-clients-and-javascript-in-the-browser when it should be https://auth0.com/docs/protocols#oauth2-implicit-flow

This PR updates the link.


auth0build:

PR #2355, opened 2016-09-21
Add ContentType header to metadata code snippets

thameera:
Lack of this header causes the code snippets to fail (eg: in PHP)


auth0build:

PR #2358, opened 2016-09-21
Correcting and clarifying ommittance of connection in OIDC protocol.

fiddur:
You don't get an error if you ommit the connection parameter in oidc, you get the hosted login page.

As specified in discussion: https://auth0.slack.com/archives/pipeline-unity/p1474453623001083


auth0build:

mpaktiti:
Thanks for the fix @fiddur ! 👍

PR #2363, opened 2016-09-21
Fix YubiKey docs

rolodato:
- wt command wasn't working because of the --output option
- Fixed exception when stringifying HTML form
- Fixed exception in rule because of outdated jsonwebtoken lib
- Styling improvements


auth0build:

PR #2389, opened 2016-09-27
Separate HRD docs for v9 and v10

rolodato:
Both docs are identical, but the v10 has the last section about adding
custom buttons removed, since it's not supported.


auth0build:

jeffreylees (approved):

PR #2394, opened 2016-09-28
Update /anomaly-detection

pushpabrol:
The description under Brute Force Protection makes it seem to the reader that the IP address is blocked and the whole IP address is blocked. In reality the 1 user's sign in from that IP is blocked. Updated documentation to show a small note explaining this.


auth0build:

PR #2395, opened 2016-09-28
Add instructions for checking on allowed modules when using Appliance Webtasks

khorne3:

### DIFF
--- a/articles/appliance/webtasks.md
+++ b/articles/appliance/webtasks.md
@@ -89,12 +89,8 @@ module.exports = cb => {
 Afterwards, create a webtask profile using `wt-cli`, if you don't already have one.
 
 ```bash
-wt init --container "YOUR_TENANT_NAME" --url "https://sandbox.it.auth0.com" --token "eyJhbGci..." -p "a``YOUR_TENANT_NAME-default"
+wt init --container "YOUR_TENANT_NAME" --url "WEBTASK_URL" --token "eyJhbGci..." -p "a``YOUR_TENANT_NAME-default"
 ```
-
-::: panel-info wt init
-This is the very first command you will run once you install `wt-cli`. It is used to create and update webtask profiles. It also provisions a webtask token if you don't have one already. Each webtask profile has a name, a URL and a container. In order to configure a webtask profile you need to provide an email or phone number. You will receive a verification code there which you will have to input to the terminal. For more details you can use the wt built-in help: `wt init -h`.
-:::
 
 Now you are ready to register your webtask, using the `wt create` command. This command receives as input a path or URL of the webtasks's code and provides as output the URL where the webtask is available.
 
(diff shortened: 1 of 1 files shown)


auth0build:

articles/appliance/webtasks.md
```diff
@@ -27,6 +34,70 @@ The `auth0-sandbox` is the recommended method for running your custom code.
...
+#### Running the `list_modules.js` Webtask
+
+To run the `List Modules` Webtask, which requires you to send an HTTP `GET` request, you must host the code to be executed so that the URL used returns the code. If you are in need of a place to host your code, consider using [GitHub Gist](https://gist.github.com/) or a similar service.
+
+For the purposes of this example, suppose that `http://example.com/webtask.js` leads to a JavaScript file containing your Webtask code.
+
+You would then execute the following command to `GET` a list of Webtask Modules available for your given Appliance instance:
+
+```
+curl https://webtask.<a0url>.com=http://example.com/webtask.js
```
tehsis: this needs to send an authorization token to it and also use the webtask_url paremeter.
I would recommend to go with wt-cli which is going to be easier to explain.

Instructions to use the CLI are in https://<manage_url>/#/account/webtasks

tehsis (changes requested):

articles/appliance/webtasks.md
```diff
@@ -27,6 +34,92 @@ The `auth0-sandbox` is the recommended method for running your custom code.
...
+        node_version: process.version,
+        modules: natives.concat(modules)
+    });
+};
+```
+
+Afterwards, create a webtask profile using `wt-cli`, if you don't already have one.
+
+```bash
+wt init --container "YOUR_TENANT_NAME" --url "https://sandbox.it.auth0.com" --token "eyJhbGci..." -p "a``YOUR_TENANT_NAME-default"
```
mpaktiti: @tehsis Not sure if I "translated" this right (I replaced `adcasdfa` with `YOUR_TENANT_NAME`).

Also, I was wondering whether `https://sandbox.it.auth0.com` stays the same for non-US customers. I've seen it as `https://sandbox-eu.it.auth0.com` for EU tenants but I'm not sure if it applies in this case.

articles/appliance/webtasks.md
```diff
@@ -27,6 +34,92 @@ The `auth0-sandbox` is the recommended method for running your custom code.
...
+```
+
+::: panel-info wt init
+This is the very first command you will run once you install `wt-cli`. It is used to create and update webtask profiles. It also provisions a webtask token if you don't have one already. Each webtask profile has a name, a URL and a container. In order to configure a webtask profile you need to provide an email or phone number. You will receive a verification code there which you will have to input to the terminal. For more details you can use the wt built-in help: `wt init -h`.
```
mpaktiti: @tehsis Does this process (`You will receive a verification code...`) stays the same for appliance users or should I remove this?

articles/appliance/webtasks.md
tehsis: That URL depends on the appliance installation, it should be WEBTASK_URL

articles/appliance/webtasks.md
tehsis: No, appliances customers won't do this

tehsis: You still need to configure the profile (with a name, container and token) this is done by following the instructions on the webtask section on the account settings

tehsis (changes requested):

PR #2416, opened 2016-10-02
[update] Clarify behavior of Authz extension

sandrinodimattia:
This PR clarifies that any changes made in terms of group memberships (eg: adding a user to a group, removing a user from a group) will only be reflected in the user's profile and token after the user logs in again.


auth0build:

PR #2424, opened 2016-10-03
Lock Android version warning

jeffreylees:
Temporary warning for Lock version (2.0 not yet supported)


auth0build:

PR #2441, opened 2016-10-04
More info on extra button in Lock

nicosabena:


auth0build:

PR #2456, opened 2016-10-05
Update Lock var to 10.4

jeffreylees:


auth0build:

PR #2461, opened 2016-10-06
Fix started seed download for Winforms/WPF

jerriep:
I cleaned up the folder structure of the Winforms/WPF seed project repository

This fixes the download links and also updates to Packager V2


auth0build:

PR #2464, opened 2016-10-06
Explain the three layers of sessions for logout

mpaktiti:
Explain the three different levels of logout (application, Auth0, IdP) and how they should be handled.


auth0build:

PR #2466, opened 2016-10-06
Replace screenshot to show tab

mpaktiti:
The screenshot did not display the tab and this was confusing to a user.


auth0build:

PR #2481, opened 2016-10-07
Replace ${account.tenant}.auth0.com

angelaLazar:
https://trello.com/c/F1RLSX28/123-replace-account-tenant-auth0-com


auth0build:

PR #2483, opened 2016-10-08
Corrected `user_name` to `family_name` in User Profile example

lcolman:
Fixed typo, changed `user_name` to `family_name` in user profile example.


auth0build:

mpaktiti:
Nice catch, thanks! 👍

PR #2495, opened 2016-10-10
Link Identity Glossaries

khorne3:

### DIFF
--- a/articles/topics/identity-glossary.md
+++ b/articles/topics/identity-glossary.md
@@ -6,6 +6,10 @@ articles:
   - tokens/jwt
 ---
 
+::: panel-info Auth0 Identity Glossary
+For addition definitions of Auth0-related terms, please see the [Auth0 Identity Glossary](https://auth0.com/identity-glossary).
+:::
+
 # Identity Glossary
 
 <%= include('./_topic') %>


auth0build:

articles/topics/identity-glossary.md
```diff
@@ -6,6 +6,10 @@ articles:
   - tokens/jwt
 ---
 
+::: panel-info Auth0 Identity Glossary
+For addition definitions of Auth0-related terms, please see the [Auth0 Identity Glossary](https://auth0.com/identity-glossary).
+:::
```
ntotten, line 11: This is fine for now, but I would rather we work on writing more full articles for each of these terms. The reason being is that the auth0.com/identity-glossery page is going to link to our pages for "More Info" so its a little weird to hae a circular reference.

ntotten: @mpaktiti I think a lot of the work you are doing on tokens and such probably can fall into this area and be part of our "identity glossary". We should also try to add articles for other topics if we dont have them covered.

khorne3: I was thinking the same, but for now, I think we should link since those docs are in the backlog and it would be nice to have a place to go to get that info. I see a lot of defs questions asked in support.

ntotten: Yeah, agreed. It's better to have something than nothing.

PR #2520, opened 2016-10-13
Change sidebar items configuration

arielger:
Changes for Articles, APIs and Appliance sections:
- Add expanded property to show all subitems.
- Add external property to show arrow with the item (if the item will refresh the page / go to another page).
- Add default sub-item for all parent items

image

image


auth0build:

PR #2526, opened 2016-10-14
Clarify that Logout URL should be URL encoded

### LINKED ISSUE #2474: ASP NET Regular Web App Logout Call endpoint error
In the Logout section of the tutorial is suggested to call the Logout Endpoint encoding the logout URL (see the image below)

image

Having as a result, this url:

```
https://tutorials.auth0.com/v2/logout?returnTo=http%3a%2f%2flocalhost%3a36283%2f
```

After checking in the documentation of the logout endpoint, must be called sending as a parameter the logout URL without being encoded:

image

https://auth0.com/docs/logout#redirecting-users-after-logout

To solve this problem, the URLEncode function should be removed.

jerriep:
Add clarification to the Logout URL docs that the `returnTo` parameter should be URL Encoded. Closes #2474

### DIFF
--- a/articles/logout/index.md
+++ b/articles/logout/index.md
@@ -43,29 +43,29 @@ The Auth0 logout endpoint logs you out from Auth0, and optionally from your iden
 
 ## Redirect Users After Logout
 
-To redirect a user after logout, add a `returnTo` querystring parameter with the target URL as the value:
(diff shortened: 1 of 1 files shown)


auth0build:

PR #2527, opened 2016-10-14
Auth0js update

jeffreylees:


auth0build:

articles/libraries/auth0js/index.md
```diff
@@ -7,22 +7,53 @@ url: /libraries/auth0js
...
+bower install auth0.js
+```
+
+Include via our CDN:
+
+```html
+<script src="${auth0js_url}"></script>
+```
+
+Or, just [download the project](https://github.com/auth0/auth0.js) and take `auth0.js` or `auth0.min.js` from the `/build` directory and import it to your page.
```
hzalaz: Build is only present on github tags (for bower) so I'd remove this mention and let the customers use the other options.

hzalaz (changes requested):

PR #2528, opened 2016-10-14
Add panel-info to enable APIs section

mpaktiti:
We got some negative feedback from a customer that `There is no "APIs" section on the left at all`.

Added an info-panel that says how to enable the section if it's not visible.


auth0build:

PR #2529, opened 2016-10-14
Add icons to all quickstart articles

nkohari:
In support of https://github.com/auth0/manage/pull/2382.

This adds a `budicon` property to the front-matter of all quickstart articles. This value is used in the `TutorialNextSteps` component of `auth0-tutorial-navigator` when it is set to single-article mode.

Since there are quite a few articles, I double-checked that they were all given an icon using:

``` bash
find . -iname '[0-9][0-9]-*' | xargs grep -L "budicon"
```

That should have caught them all. Also, sorry about the clutter in this patch; my IDE is set to remove trailing whitespace on save!


auth0build:

PR #2564, opened 2016-10-17
Remove Lock v9 instructions from custom social connection docs

rolodato:


auth0build:

PR #2565, opened 2016-10-17
Password Grant Fix in Protocols 

ghost:
Minor update to Auth0: /docs/protocols#oauth-resource-owner-password-credentials-grant

Password Grant does not receive a `client_secret`. You can verify this on the Authentication API: https://auth0.com/docs/api/authentication#!#post--oauth-ro


auth0build:

PR #2567, opened 2016-10-17
Include SingleLogout service URL when Auth0 is SP

nicosabena:


auth0build:

PR #2571, opened 2016-10-18
Update extensions support on appliance

mpaktiti:
- add note on Delegated Admin (not supported).


auth0build:

Issue #2550, opened 2016-10-17
react tutorial hardcoded lock library url 


vmartynets:
[React tutorial](https://auth0.com/docs/quickstart/spa/react/00-intro#2-dependencies) contains [hardcoded](https://github.com/auth0/docs/blob/master/articles/client-platforms/react/_includes/_dependencies.md) url to cdn instead of using [variable](https://github.com/auth0/docs/blob/master/CONTRIBUTING.md#common-variables) (AUTH-3455)
![image](https://cloud.githubusercontent.com/assets/1114365/19440502/d1159a9a-948a-11e6-9435-a0f58438b95e.png)

camimonse:
Sanity passed

PR #2572, opened 2016-10-18
Replace hardcoded lock url with lock_url var

mpaktiti:
Fixes https://github.com/auth0/docs/issues/2550


auth0build:

PR #2573, opened 2016-10-18
Remove mention of customizing Lock via JS

rolodato:


auth0build:

PR #2585, opened 2016-10-19
Update clients URL at sidebar menu

mpaktiti:
Clicking on the Clients menu of the sidebar was redirecting the user to `https://auth0.com/docs/applications`. The doc was displayed correctly, but I updated the URL in order to display the correct one at the browser's address bar.

image


auth0build:

PR #2586, opened 2016-10-19
[edit]: added port 443 for docker.it.auth0.com

ndelgado:
from the latest version of the appliance (8986)
https://auth0.com/changelog/appliance
The docker repository can now use port 443 instead of port 5000 for getting updates. In some environments using a non-standard port was problematic.


auth0build:

PR #2590, opened 2016-10-20
Update docs on scopes for ASP.NET Core quickstarts

jerriep:
I updated the ASP.NET Core quickstart to ensure that we are more explicit about the scopes being requested and not just request the `profile` scope by default.

This is in response to the following Slack request:
https://auth0.slack.com/archives/quickstarts/p1475847386000420


auth0build:

PR #2592, opened 2016-10-20
Updates to delegated admin 2.3

sandrinodimattia:
This update reflects updates in the delegated admin extension.


auth0build:

PR #2596, opened 2016-10-21
Java QuickStarts: Add content on scopes and bump library versions

mpaktiti:
Fixes https://github.com/auth0/auth0-spring-security-api/issues/14


auth0build:

PR #2605, opened 2016-10-21
[edit]: added requirement for cert to be public

ndelgado:


auth0build:

Issue #2600, opened 2016-10-21
java tutorial outdated lock


vmartynets:
(AUTH-3493)
Snippet in [java tutorial](https://auth0.com/docs/quickstart/webapp/java/01-login#display-lock-widget) contains outdated lock library.
![image](https://cloud.githubusercontent.com/assets/1114365/19603233/09e01d06-97b9-11e6-8169-973da45129fc.png)

camimonse:
Sanity Passed

PR #2614, opened 2016-10-24
Replace hardcoded lock with var

mpaktiti:
Fixes https://github.com/auth0/docs/issues/2600


auth0build:

Issue #2402, opened 2016-09-29
Update Clients document


cabralmartin:
[https://auth0.com/docs/applications](https://auth0.com/docs/applications)
- Explain client types and implications on authN/Z flows
  - Consider how settings change depending on the client
  - Consider how different flows apply to different clients (e.g. a SPA cannot do client credentials)
- Explain new settings
  - `token authentication endpoint`
  - `first party`
- Update client examples. Don't recommend sharing the client id & secret amongst different clients

PR #2615, opened 2016-10-24
Update Clients overview doc

mpaktiti:
Fixes https://github.com/auth0/docs/issues/2402


auth0build:

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
 
 # Core Concepts: Clients
 
-An Auth0 **client** allows your application to utilize Auth0 for authentication. A fully-configured authentication process will consist of an Auth0 client that possesses information about and is able to communicate with your application (which, conversely, possesses information about and is able to communicate with the Auth0 client.)
+## Overview
 
-A client is a core Auth0 concept, so it's important to know how they:
+An Auth0 **client** maps to your application and allows to use Auth0 for authentication. The term *client* does not imply any particular implementation characteristics. Your application can be a native app that executes on a mobile device, a single page app that executes on a browser, or a regular web app that executes on a server. The term is aligned with the [OAuth 2.0 specification](https://tools.ietf.org/html/rfc6749), where a client is defined as *an application making protected resource requests on behalf of the resource owner and with its authorization*.
```
cabralmartin: Not sure if we should scope to OAuth only since clients may operate in ways out of that spec.

cabralmartin: `protected resource requests on behalf of the resource owner and with its authorization` speaks of a client trying to reach an API. While we support this authZ scenario, it sounds like clients are no good for authN only, which would be incorrect.

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
 
-Addons are extensions associated with clients. They are typically third-party APIs used by the application(s) for which Auth0 generates access tokens. Some typical scenarios used include:
+- **Regular Web Applications**: Used for traditional web applications that run on a server, like ASP .NET, Java or Node.js. For a complete listing of the SDKs Auth0 offers for Web Apps refer to: [Web App SDKs](/quickstart/webapp).
 
-* **Accessing External APIs**: Using the Delegation endpoint, you can exchange a Client's access token for a third-party service's (such as Salesforce or Amazon) access token
-* **Integrating with Applications Using SAML2/WS-Federation**: Addons allow you to integrate with any custom or SSO integration that does not currently enjoy built-in Auth0 support, since they allow you to configure every aspect of the SAML2/WS-Federation integration.
+- **Non Interactive Clients**: Used for server to server applications like CLIs, daemons or services running on your backend. Typically you would use this option if you have a service that requires access to an API and you want it to request access tokens in order to do so.
```
cabralmartin: I would remove this piece `and you want it to request access tokens in order to do so.`

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
 
-![](/media/articles/applications/applications-sso-integrations-overview.png)
+After you set the name and client type, click **Create**.
 
-## Connections
+Each client has four available tabs:
 
-![](/media/articles/applications/applications-connections-example.png)
+- [Quick Start](${manage_url}/#/clients/${account.clientId}/quickstart): Lists all available Quick Starts, filtered by your client's type.
```
cabralmartin, line 35: is it customary to use this type of links? these will become broken links for non authenticated users and not sure which client will use for the rest.

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
 
-Within the context of a rule, we have access to the Client the user is authenticating to, which is useful if we want to apply coarse-grained authorization policies for our applications:
+- **Domain**: Your Auth0 account name. Note that the domain name is chosen when you create a new Auth0 account and cannot be changed. If you need a different one you have to register for a new account by selecting *New Account* at the top right menu.
 
-* Only HR officials can access Application X
-* Only US-based users can access Application Y
+- **Client ID**: The unique identifier for your client. This is the ID you will use with when configuring authentication with Auth0. It is generated by the system when you create a new client and it cannot be modified.
 
-Here's a [sample rule](https://github.com/auth0/rules/blob/master/rules/simple-user-whitelist-for-app.md) where only the users in the whitelist are allowed to access the application:
+- **Client Secret**: A base64 encoded string used to sign and validate tokens for authentication flows and to gain access to select Auth0 API endpoints. By default, the value is hidden, so check the **Reveal Client Secret** box to see this value.
```
cabralmartin: `sign and validate tokens` -> `sign and validate id_tokens`.

cabralmartin: we should add a note about being super careful about safe keeping the secret

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
-      return callback(null, user, context);
-    }
+- **Client Type**: The type of client you are implementing. Depending on which you choose, the available settings differ to show you only the settings applicable to your Client Type. You can change this value at any time by selecting one of the following: Native, Non Interactive Client, Regular Web Application, or Single Page Application.
 
-    var whitelist = [ 'user1@mail.com', 'user2@mail.com' ]; //authorized users
-    var userHasAccess = whitelist.some(
-      function (email) {
-        return email === user.email;
-      });
+- **First Party Client**: When a First Party client requests authorized against an API with the **Allow Skipping User Consent** flag set, the User Consent dialog will not be shown to the final user. Note that if the hostname of your callbackURL is `localhost` or `127.0.0.1` the consent dialog will always be displayed.
```
cabralmartin: `client requests authorized against` -> `client requests authorized access against`

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
 
-![](/media/articles/applications/applications-logs-auditing.png)
+- **Allowed Origins (CORS)**: Set of URLs that will be allowed to make requests from JavaScript to Auth0 API (typically used with CORS). This prevents same-origin policy errors when using Auth0 from within a web browser. By default, all your callback URLs will be allowed. This field allows you to enter other origins if you need to. You can specify multiple valid URLs by comma-separating them. You can use the star symbol as a wildcard for subdomains (`*.google.com`). Notice that querystrings and hash information are not taking into account when validating these URLs.
 
-The logs include many of the actions performed by the user:
+- **JWT Expiration (seconds)**: The amount of time (in seconds) before the Auth0 [JWT](/jwt) expires. The default value is `36000`, which maps to 10 hours.
```
cabralmartin: `before the Auth0 id_token expires`

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
-* **Client Type**: the type of client you are implementing (depending on which you choose, the settings available differ);
-* **Token Endpoint Authentication Method**: the requested authentication method for the token endpoint: `None` (public client without a client secret), `Post` (client uses HTTP POST parameters), or `Basic` (client uses HTTP Basic parameters);
-* **Allowed Callback URLs**: the URLs of your application(s) to which Auth0 can redirect the user after authentication.
-* **Allowed Logout URLs**: the URLs that Auth0 can redirect your users to after logging out;
-* **Allowed Origins (CORS)**: the URLs of the applications running your JavaScript code (prevents same-origin policy errors when using Auth0 from within a web browser)
-* **JWT Expiration (Seconds)**: the amount of time before the Auth0 access tokens expires
-* **Use Auth0 instead of the IdP to do Single Sign On**: if enabled, this setting prevents Auth0 from redirecting authenticated users with valid sessions to the identity provider (such as Facebook, ADFS, and so on)
+The client type should match your application:
```
cabralmartin: should we extract this as a separate section? (i.e. # Client Types) wdyt? I would also place it before speaking about how to create a client.

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
-Connections may be shared among multiple clients.
+- [Addons](${manage_url}/#/clients/${account.clientId}/addons): Add-ons are extensions associated with clients. They are typically third-party APIs used by the client(s) for which Auth0 generates access tokens. For more details refer to: [Addons](/clients/addons).
 
-## Rules
+- [Connections](${manage_url}/#/clients/${account.clientId}/connections): Connections are sources of users. They are categorized into Database, Social and Enterprise and can be shared among different clients. For more details refer to: [Connections](/clients/connections). For a detailed list on the supported Identity Providers refer to: [Identity Providers Supported by Auth0](/identityproviders).
 
-![](/media/articles/applications/rules-flow.png)
 
-[Rules](/rules) are code snippets written in JavaScript that are executed as part of the Auth0 authentication process. This happens every time a user authenticates. Rules enable very powerful customizations and extensions to be easily added to Auth0.
+Click on the [Settings](${manage_url}/#/clients/${account.clientId}/settings) tab of your client to review the available settings:
```
cabralmartin, line 45: Maybe this would be better presented as a separated sub section (i.e. ## Client Settings) wdyt?

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
-* Logging in to a Client;
-* Failing to log in to a Client;
-* Signing up;
-* Requesting a password change;
+- **Use Auth0 instead of the IdP to do Single Sign On**: If enabled, this setting prevents Auth0 from redirecting authenticated users with valid sessions to the identity provider (such as Facebook, ADFS, and so on).
 
-You can download the event logs using the [Management API](/api/management/v2#!/Logs/get_logs) or view them via the [Management Dashboard](${manage_url}/#/logs)
-
-While the [Management Dashboard](${manage_url}/#/logs) displays the log data in a neatly formatted manner, clicking on the row corresponding to a particular event displays the raw data, which looks something like this:
+::: panel-info Auth0 Rules
```
cabralmartin: Is this necessary in this document?

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
-  "connection_id": "con_T60D5poVozRmw77h",
-  "client_id": "UEsQCe4RVHYDSQ2zlLWIAHSqDhpsYyTG",
-  "client_name": "N/A",
-  "ip": "108.248.62.158",
-  "user_agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36",
-  "strategy": "samlp",
-  "strategy_type": "enterprise"
-}
-```
+## Sample Clients
```
cabralmartin: would it make sense to remove these samples from here and direct the user to the [Architecture Scenarios doc](https://auth0.com/docs/architecture-scenarios)? that way we can focus the effort on maintaining scenarios in a single place. wdyt?

articles/clients/index.md
```diff
@@ -4,186 +4,167 @@ description: Explains the basics of creating and using Auth0 Clients.
...
 
-This example is a timesheet application that utilizes:
+For more details on APIs and how to configure one refer to: [APIs](/apis).
 
-* A REST API;
-* A Single Page Application (SPA) hosted on a server different from the one hosting the REST API;
-* Mobile apps capable of running on several types of devices.
 
-From a technical standpoint, the above comprises at least three Clients, due to their differing language, deployment model, and so on. However, for Auth0, this is **one** Client application that shares an ID-secret pair.
+## Auditing
```
cabralmartin: should we leave this here? maybe we can point to a much detailed doc where we discuss logs? wdyt?

cabralmartin (changes requested):

articles/clients/index.md
mpaktiti: Removed the last sentence.

articles/clients/index.md
mpaktiti: Yes we use these variables at docs. If the user is not logged in then he is redirected to Lock to login. You can try it from incognito and see the experience. When you login you get the Clients page (there is an error displayed for some seconds at the top of the page but then it goes away on its own).

articles/clients/index.md
mpaktiti: Can you explain why? Is the statement incorrect or you don't like the syntax?

articles/clients/index.md
mpaktiti: Sure, better this way. I moved it on its own sub-paragraph, before how to configure a client.

articles/clients/index.md
mpaktiti: OK I will add the note. A question about this `sign and validate id_tokens`: this applies even if I used HS256 for my API? Isn't the secret used for HMAC algorithms? I understand that it is not recommended to use HS256 but it is feasible, no?

articles/clients/index.md
mpaktiti: Fixed 👍

articles/clients/index.md
mpaktiti: This isn't used for the access token as well? When does the access token expire?

articles/clients/index.md
mpaktiti: I think it should stay. The doc is a generic introduction into Clients and rules are important. There was a whole paragraph here, which was too much, so I removed it and added this panel instead. 

It's the same with the logging info. I can remove the content to another doc, but I should mention it in this one so if anyone is interested can easily find this information.

articles/clients/index.md
mpaktiti: I agree, I don't think it serves any purpose really.

articles/clients/index.md
mpaktiti: Agreed 👍

articles/clients/index.md
mpaktiti: Agree 👍

articles/clients/index.md
mpaktiti: On a second thought I removed it. It is closely related to clients but it seemed out of context.

articles/clients/index.md
cabralmartin: It is correct that a NI will require to request access_tokens in order to access a secured API. But it sounds funny to me how it's phrased; it's like "if I want to access and API AND I want to request access tokens then I will select this client". Maybe talking about access tokens at this point is more of an implementation detail.

articles/clients/index.md
cabralmartin: I'm getting this

image

articles/clients/index.md
cabralmartin: This secret we see in the client settings can serve multiple purposes. One one hand, it's going to be use to sign the `id_tokens` (not `access_tokens`) when the algorithm is HS256. On the other hand, it's also "the client password" and be used on the token endpoint for executing several operations (e.g. client grant exchange, code exchange, etc).

Again, this is different from the API secret. The API secret is used for signing `access_tokens` (if using HS256).

The funny thing about client secrets is that are some client types that have no use for it (i.e. SPA / Native). Further down the line we will be removing them from the dashboard for those app types.

articles/clients/index.md
cabralmartin: The `access_token` lifetime is managed under the API settings. I'm thinking we could use a sub section similar to this one in the API doc where we go setting by setting. wdyt?

cabralmartin:
Addressed comments // cc @mpaktiti

articles/clients/index.md
mpaktiti: That's because there is a bug in the heroku review apps. There is a missing `:` so the URL is `https//manage....`. Add this manually and see how it works.

articles/clients/index.md
mpaktiti: Got it, in the API case the Global Client Secret is used for signing, not the Client one. Thanks for clarifying!

articles/clients/index.md
mpaktiti: Good idea, I'll make update that doc next.

PR #2618, opened 2016-10-24
Fix minor issues in jQuery authz quickstart

jhiner:
I fixed a couple minor issues with the jQuery API Authz quickstart.


auth0build:

PR #2621, opened 2016-10-24
MFA via amr on id_token

topherMarie:

### DIFF
--- a/articles/multifactor-authentication/developer/mfa-from-id-token.md
+++ b/articles/multifactor-authentication/developer/mfa-from-id-token.md
@@ -1,0 +1,38 @@
+---
+description: Describes how to determine if a user has done multifactor authentication using their id_token and JWT
+---
+
+## Determine if a user has performed multifactor authentication
+
+At times, it's necessary to determine if a particular user has had the additional security of multifactor authentication applied to their session. For instance, a user may be allowed access to sensitive data or allowed to reset their password only after further confirming their identity using MFA.
+
+For a particular user session, developers can check the claims information available on the `id_token`, a [JSON Web Token](/jwt) provided by Auth0 that contains claims information relevant to the user's session. After [retrieving the id_token](/tokens/id_token), the value of the `amr` field can be evaluated to see if it contains `mfa` as a claim. 
+
+Note that `amr` can contain claims other than `mfa`, so its existence is not a sufficient test. Its contents must be examined for the `mfa` claim.
+
+This example is built on top of the [JSON Web Token Sample Code](https://github.com/auth0/node-jsonwebtoken).
+
+```js
+const AUTH0_CLIENT_SECRET = '${account.clientSecret}';
+const jwt = require('jsonwebtoken')
+
+jwt.verify(id_token, new Buffer(AUTH0_CLIENT_SECRET, 'base64'), { algorithms: ['HS256'] }, function(err, decoded) {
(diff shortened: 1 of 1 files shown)


auth0build:

articles/multifactor-authentication/mfa-from-id-token.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+## Determine if a user has performed multifactor authentication
+
+At times, it's necessary to determine if a particular user has had the additional security of multifactor authentication applied to their session. For instance, a user may be allowed access to sensitive data or allowed to reset their password only after further confirming their identity using MFA.
+
+To determine this for a particular user session, developers can check the claims information available on the `id_token`. 
+The `id_token` contains claims information relevant to the usage of multifactor authentication. After retrieving the `id_token`, [as shown here](tokens/id_token), the value of the `amr` claim can be evaluated.
+
+```
```
dafortune: I think you could specify the language (js) after the ``` so you get better highlighting. I'd also mention that the example is based on https://github.com/auth0/node-jsonwebtoken and maybe you could create a `const` on top with the secret, something like:

```
const AUTH0_CLIENT_SECRET = '${account.clientSecret}'
```

that'd be replaced on runtime https://github.com/auth0/docs/blob/master/CONTRIBUTING.md#document-variables

articles/multifactor-authentication/mfa-from-id-token.md
```diff
@@ -0,0 +1,28 @@
+---
+description: Describes how to determine if a user has done multifactor authentication using their id_token and JWT
+---
+
+## Determine if a user has performed multifactor authentication
+
+At times, it's necessary to determine if a particular user has had the additional security of multifactor authentication applied to their session. For instance, a user may be allowed access to sensitive data or allowed to reset their password only after further confirming their identity using MFA.
+
+To determine this for a particular user session, developers can check the claims information available on the `id_token`. 
+The `id_token` contains claims information relevant to the usage of multifactor authentication. After retrieving the `id_token`, [as shown here](tokens/id_token), the value of the `amr` claim can be evaluated.
```
dafortune: I'd say that the value might not be present or might contain a different values other than just `'mfa'`, so you should always check that the `'mfa'` string is in the array and not just the trusting that the existence of the array is enough to tell that. Right now the only value is mfa, but will change for sure in future.

dafortune: Also depending on the target audience might worth having a link to the RFC docs, but not sure in this case.

articles/multifactor-authentication/mfa-from-id-token.md
dafortune: I know that I sent you the example, but it has some small mistakes, here is an examples that actually works:

``` js
const AUTH0_CLIENT_SECRET = '${account.clientSecret}';
const jwt = require('jsonwebtoken')

jwt.verify(id_token, new Buffer(AUTH0_CLIENT_SECRET, 'base64'), { algorithms: ['HS256'] }, function(err, decoded) {
   if (err) {
     console.log('invalid token');
     return;
   }

   if (Array.isArray(decoded.amr) && decoded.amr.indexOf('mfa') >= 0) {
     console.log('You used mfa');
     return;
   }

   console.log('you are not using mfa');
 });
```

PR #2623, opened 2016-10-25
Add missing event codes

xoob:
This PR adds some event codes we noticed were missing from https://auth0.com/docs/logs.

It might be best if Auth0 can update this list from their authoritative source so it has the complete list of names used in https://manage.auth0.com/#/logs, as there might be others we haven't noticed.


auth0build:

mpaktiti:
Thanks for the contribution @xoob ! 👍

PR #2624, opened 2016-10-25
Update index of policies

lobo:
Includes the URL of the new policy.


auth0build:

PR #2630, opened 2016-10-25
[Rule] Specify that the callback should use `UnauthorizedError`

sorin-davidoi:
I think this has to be stated explicitly, since it is pretty common to just use the standard `Error` class in JavaScript.


auth0build:

mpaktiti:
Thanks for the contribution @sorin-davidoi ! 👍

PR #2636, opened 2016-10-25
Step up

topherMarie:


auth0build:

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,53 @@
...
+var sessionHasStepUp(id_token){
+  var decoded = jwt.verify(id_token, new Buffer(AUTH0_CLIENT_SECRET, 'base64'), { algorithms: ['HS256'] });
+
+  return (Array.isArray(decoded.acr) && decoded.acr.indexOf(mfaAcr) >= 0);
+};
+```
+
+Assuming that step-up authentication is necessary, how is it requested?
+
+Auth0 will know that a user is required to perform a multifactor authentication if the request contains `acr_values` that include that same mfaAcr as above. For instance, when rendering the [Lock widget](/libraries/lock) you can do the following.
```
fiddur: Requesting `acr_values` is no guarantee that mfa will be performed.  You must also check the returned `id_token` for `acr` or `amr`.  Today, they will both be present when `acr_values` is requested.  BUT:

Since `acr` is a single value, while `amr` is an array, it COULD be more future proof to validate `amr`.  If there would be a higher level of security at the same time as a multifacor, then the `acr` could point to that instead.

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,76 @@
...
+To evaluate if the current session has been authenticated using the desired level of authentication, the `acr` value and/or the `amr` value may be evaluated.
+```js
+const AUTH0_CLIENT_SECRET = '${account.clientSecret}';
+const mfaAcr = 'http://schemas.openid.net/pape/policies/2007/06/multi-factor';
+const jwt = require('jsonwebtoken')
+
+var sessionHasMFA(id_token){
+  var decoded = jwt.verify(id_token, new Buffer(AUTH0_CLIENT_SECRET, 'base64'), { algorithms: ['HS256'] });
+
+  return (Array.isArray(decoded.acr) && decoded.acr.indexOf(mfaAcr) >= 0);
```
dafortune: Is acr an array?

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,76 @@
+---
```
dafortune, line 1: I'd also say that if you want to request step up in place you should add a nonce and check it afterwards, or (less secure) at least check the token iat for it to be recent; otherwise id tokens last a day and could be re-played to you at any time if a third party captures it (many people put them in local storage, which is pretty easy to access)

dafortune: If trusting the current id token is ok, then that "transaction" belonging verification should have been done before (but the token still last a day or so)

articles/multifactor-authentication/developer/step-up-with-acr.md
fiddur: No, `acr` is just one string.  `amr` is an array.

woloski: Did you try this @topherMarie?

articles/multifactor-authentication/developer/step-up-with-acr.md
woloski: I would expect first to see something about "How to check if the user has logged in with a second factor" and explain that you need to check the acr claim in the id_token. You don't need to verify the token to do that, that should have already happened as part of the login process in your API or regular web app (assuming you are checking this on the server side).

woloski: Would be useful to add a note on the MFA documents of each quickstarts: https://auth0.com/docs/quickstart/spa/vanillajs/09-mfa.

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,76 @@
...
+```js
+const mfaAcr = 'http://schemas.openid.net/pape/policies/2007/06/multi-factor';
+
+if(pageIsSensitive() && !sessionHasMFA(id_token)){
+  const lockOptions = {
+    auth: {
+      params: {
+        state: this.props.stepup.transactionId,
+        nonce: this.props.stepup.nonce,
+        scope: 'openid nonce',
```
woloski: adding nonce sounds like a hack

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,76 @@
...
+## Example
+To request that Auth0 require a multifactor authentication, add the field `acr_values` to the authentication request along with the `acr` level desired. For instance, when rendering the [Lock widget](/libraries/lock), the following example will prompt for multifactor authentication.
+```js
+const mfaAcr = 'http://schemas.openid.net/pape/policies/2007/06/multi-factor';
+
+if(pageIsSensitive() && !sessionHasMFA(id_token)){
+  const lockOptions = {
+    auth: {
+      params: {
+        state: this.props.stepup.transactionId,
```
woloski: where this `this.props.stepup` thing is coming from?

articles/multifactor-authentication/developer/step-up-with-acr.md
```diff
@@ -0,0 +1,76 @@
...
+
+lock = new Auth0Lock('clientID', 'account.auth0.com', options);
+```
+
+To confirm that a session has had multifactor authentication, the id_token can be checked for its `acr` and `amr` claims.
+```js
+var decoded = jwt.verify(id_token, new Buffer(AUTH0_CLIENT_SECRET, 'base64'), { algorithms: ['HS256'] });
+
+// Confirm that the acr has the expected value
+if(!Array.isArray(decoded.acr) || decoded.acr.indexOf(mfaAcr) < 0){
```
dafortune: Seems that acr and amr checks are inverted, amr for sure is an array; and I'm not sure about acr, but I think it is just an string

articles/multifactor-authentication/developer/step-up-with-acr.md
dafortune: Agree in case of "logging in with a second factor", but I think that "logging in with a second factor" (which is usually handled by your framework) is not the same as "step-up" specially if you don't want the user to pass the whole authentication process as if it where re-logging in. For "checking if logged in with a second factor"  I think that what is stated here is ok.

dafortune:
Looks good to me

PR #2639, opened 2016-10-26
[edit]: add "Auth0 Appliance can only be deployed in 1 NIC."

ndelgado:
per Sandrino, we need to add this: https://auth0.slack.com/archives/appliance-projectmgmt/p1477422525000150


auth0build:

PR #2643, opened 2016-10-26
[edit]: removed note that is no longer true

ndelgado:
This is no longer true (CDN access is needed for web extensions): You do not need to allow access to cdn.auth0.com for Appliance versions after 6868. More info: https://auth0.slack.com/archives/appliance-projectmgmt/p1477338109000136


auth0build:

PR #2648, opened 2016-10-27
Revise Swift credentials section

cocojoe:


auth0build:

chenkie:
Why are we removing the generated client ID and domain (namespace) in favor of the generic `YOUR_CLIENT_ID` etc strings?

cocojoe:
Sorry my mistake, I was testing locally and had copied the data from the web when unauthenticated. Allow me to correct.

PR #2650, opened 2016-10-27
Java quickstarts update

mpaktiti:
Bump library versions and update to the latest seed downloader.


auth0build:

PR #2659, opened 2016-10-28
Added a link to user schema and examples for user import extension

AmaanC:


auth0build:

PR #2680, opened 2016-10-31
API Authorization updates

mpaktiti:
- Add new doc on which OAuth flow to use


auth0build:

articles/api-auth/which-oauth-flow-to-use.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Is the Client a native app or a SPA?
+
+If the Client is a Single Page Application (meaning an application running in a browser using a scripting language such as Javascript) then most of the times the [Implicit Grant](/api-auth/grant/implicit) is used. In this case, instead of getting an authorization code that needs to be exchanged for an access token, the Client retrieves directly an access token. On the plus side, this is more efficient since it reduces the number of round trips required to get an access token. However, a security consideration is that the access token is exposed on the client side. Since we are talking about applications running in a browser, viewing the access token is fairly easy. One counter measure is to use [the `state` authentication parameter](/protocols/oauth-state). Also, it should be noted that **Implicit Grant** does not return a refresh token because the browser cannot keep it private. You need to get a new access token, using the `/authorize` endpoint of the [Auth0 Authentication API](/api/authentication), each time the old expires.
```
cabralmartin:
```
then most of the times the [Implicit Grant](/api-auth/grant/implicit) is used.
```

The Implicit Grant is the only option for SPA. SPAs cannot execute a Code grant since they cannot interact with the token endpoint (this type of clients have `token_endpoint_authentication_method=none` set which turns them unable to use the endpoint altogether).

It's not clear to me what's implied by saying 

```
 Since we are talking about applications running in a browser, viewing the access token is fairly easy.
```

Does this mean that the token might get subtracted? This is a fair point when it comes to the response mode (fragment vs query) but I think this might be a lil too advanced into the topic to present it now, wdyt?

```
One counter measure is to use [the `state` authentication parameter](/protocols/oauth-state).
```

The `state` param is there solely to prevent CSRF attacks. The idea behind this param is that the client (when called with an access_token on the callbackUrl) can identified that the authZ flow was initiated by itself and not by a third party. Maybe this is also a topic that we can wait to present.

articles/api-auth/which-oauth-flow-to-use.md
```diff
@@ -0,0 +1,35 @@
...
+
+**NOTE:** A workaround to improve user experience is to use `prompt=none` when you invoke the `/authorize` endpoint. This will not display the login dialog or the consent dialog so the user is not aware what's happening behind the scenes.
```
cabralmartin: A minor note here,

```
is not aware what's happening behind the scenes.
```

the customer will see all the redirects happening, so the UI will have to reload. The workaround for this to keep in under the hood, is to call `/authorize` from a hidden `iframe` and extract the new access token from the parent frame.

Maybe what we can say is something down the lines of `while SPAs cannot use refresh tokens, they can take advantage of other mechanics that provide the same function`

cabralmartin (changes requested):

cabralmartin:
Small items listed, otherwise content looks great 👍 

Also, two other small things:
- When we go over each question expanding on it, we should link to the recommended flow. So if I read the explanation of the question and I'm like "ok this is my escenario" I know what I should read next.
- Let's add a p on the overview doc with a reference to this content. Something like "do you need help deciding? check this out"

articles/api-auth/which-oauth-flow-to-use.md
mpaktiti: Great input, thanks, I updated the content 👍

PR #2681, opened 2016-10-31
Migration state redirect after rule 

BRVNDXN:
+### State Parameter required on redirect from rule
 +
 +| Severity | Effective Date |
 +| --- | --- | --- | --- |
 +| High | 2016-11-01 |
 +
 +When a redirect is done from an Auth0 rule, Auth0 takes care of generating and sending a state parameter in HTTP and Auth0 will check for a valid state parameter when flow returns to the /continue endpoint.  The site to which the redirect goes has to capture the value of the state parameter and return it by adding it as a parameter when returning to the /continue endpoint. 
 +
 +This is documented [here](/rules/redirect#what-to-do-after-redirecting)
 +
 +#### Am I affected by the change? 
 +You are effected by the change only if you redirect from rules, and do not yet capture and return (to the /continue end point) the state parameter.
 +

### DIFF
--- a/articles/email/templates.md
+++ b/articles/email/templates.md
@@ -29,7 +29,7 @@ The **From Address** field supports the following macros:
 * `{application.name}`
 * `{connection.name}`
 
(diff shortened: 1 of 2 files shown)


auth0build:

Issue #2671, opened 2016-10-31
Ember.js Single Page App missing ellipses at code snippets


rafanog:
There are some missing ellipses in the code snippets of the [Ember.js Single Page App](https://auth0.com/docs/quickstart/spa/emberjs) tutorial.

1) At the "[Configure the add on](https://auth0.com/docs/quickstart/spa/emberjs#2-configure-the-add-on)" section, on the first snippet:
![missing ellipses](https://cloud.githubusercontent.com/assets/15782359/19843882/017ff528-9efd-11e6-97f0-b32f06ce25f3.JPG)

2) Second snippet, also at the "Configure the add on" section
![missing ellipses 2](https://cloud.githubusercontent.com/assets/15782359/19843858/c0815fbc-9efc-11e6-9c2c-59c884ab78f3.JPG)

TechAid ticket: AUTH-3357

vmartynets:
Same in the [ember.js 2 tutorial](https://auth0.com/docs/quickstart/spa/ember2js)
![image](https://cloud.githubusercontent.com/assets/1114365/19856645/0ec277e8-9f83-11e6-9e26-cb5c60b23965.png)

PR #2684, opened 2016-11-01
Added missing ellipses to ember2js tutorial

fiannolo:
Added missing ellipses to ember2js tutorial according to the github issue: https://github.com/auth0/docs/issues/2671


auth0build:

ntotten:
@chenkie There seems to be some inconsistency on this. I noticed you removed some of the `...` on some samples. Which method do you prefer?

chenkie:
Going to change the approach here to favor fewer ellipses. Will adjust in testing cycle instructions.

sheriffderek:
For the record, there are no ellipsis on the live quickstart as of this date

PR #2687, opened 2016-11-01
P&P Server+API: Addressing review comments

mpaktiti:
- Redacting text from the premise
- Remove references to mobile + SPA clients
- Move OAuth 2.0 reference to solution paragraph
- Add note on secured vs non secured endpoints
- Add conclusion


auth0build:

PR #2690, opened 2016-11-01
Simplify an expression for a datetime conversion to unix format

testfirstcoder:


auth0build:

chenkie:
@jerriep can you review? :)

jerriep:
Hi @testfirstcoder 

Thanks for the contribution. Unfortunately the `ToUnixTimeSeconds()` method is only available in .NET Framework 4.6, Windows 10 and .NET Core, so it will not work for all our users.

For now we will keep the existing code sample as-is. 

If you want to you can instead just add a comment indicating that users can use `DateTimeOffset.ToUnixTimeSeconds()` on later versions of the .NET Framework

PR #2691, opened 2016-11-01
Remove region support

mpaktiti:
Functionality is now available globally.


auth0build:

PR #2710, opened 2016-11-03
added logout section to auth0.js library reference

chenkie:


auth0build:

Issue #2502, opened 2016-10-11
Improve tutorial instruction for better user experience


camimonse:
in https://auth0.com/docs/quickstart/native/android/01-login in the third step, it says: "inside your activity" as indicating that the user has to search in a different place than before (second step).

For a more comfortable user experience is better to say: "**scroll down** and add the authentication callback" or "**scroll down** and add the authentication callback" **inside your activity**. Since the location of that code is right down there and the user doesn't have to take time looking somewhere else as it was for the first and second step. (photo shows where I mean)

![image](https://cloud.githubusercontent.com/assets/18585436/19282361/156cba80-8fbc-11e6-8d71-14b839e963f7.png)

camimonse:
reference to AUTH-3440 ticket

chenkie:
@camimonse I'm a bit unclear on where this new text should go. Can clarify where the ideal text should go?

camimonse:
@benschwarz it would go here, (look the red arrow):
![image](https://cloud.githubusercontent.com/assets/18585436/19326152/6000b110-9097-11e6-80d2-de380fb18abb.png)

iannyanes:
@chenkie my suggestion would be to just give those snippets with one instruction since all methods mentioned are in the activity.

Something like:

Inside the activity add the following code to the methods indicated below:

To the Method 1:
[Snnipet]

To the method 2:
[Snnipet]

Because if we mention the activity component in the last instruction the user might think that the previous ones were not supposed to be executed on that same component.

camimonse:
@iannyanes Helpful suggestion! thanks

camimonse:
Sanity passed

PR #2711, opened 2016-11-03
instructions improved

camimonse:
for a better user experience, since setting these snippets is simpler than the tutorial shows. This shows that is in the same activity and the user just have to scroll down to keep with the setting.


auth0build:

camimonse:
this fixes #2502

chenkie:
@camimonse can you rebase? Thanks

PR #2716, opened 2016-11-04
Using SSO with social IdP

jerriep:
Explains conditions under which SSO works with social IdP


auth0build:

PR #2718, opened 2016-11-04
Update using-auth0-to-secure-an-api.md

toshke:
Sample code was updated, so user does not need to copy/paste authorization code from browser, but it's rather picked up from express server. In order for this to work localhost:3000 needs to be registered as callback url for client.


auth0build:

khorne3:
Closing due to inactivity. Would be happy to continue work on this in the future.

PR #2719, opened 2016-11-04
Load lock widget from Auth0 CDN via HTTPS

christianliebel:
Client applications delivered via HTTPS (which always is a good idea in authentication contexts) won’t be able to load the lock widget from the Auth0 CDN via HTTP.


auth0build:

mpaktiti:
Hello @chliebel! Actually this hardcoded URL shouldn't be there at all. We are using [variables](https://github.com/auth0/docs/blob/master/CONTRIBUTING.md#document-variables) for the latest URL of Lock script CDN location, so with each release we only need to update the variable. I'm gonna close this PR and update the doc so it uses our `lock_url` variable instead. Thanks for bringing this to our attention! 👍

Issue #2601, opened 2016-10-21
Numbering is needed in the PHP Symfony quicikstart


camimonse:
**Context:**

In the description of the PHP Symfony tutorial:  https://auth0.com/docs/quickstart/webapp/symfony are "steps" referenced, such as this examples bellow: 

![image](https://cloud.githubusercontent.com/assets/18585436/19606994/2030f358-979d-11e6-98f3-f4b3b84bacb8.png)

 If you scroll down a little, you found another one, see photo bellow:

![image](https://cloud.githubusercontent.com/assets/18585436/19607058/c0cc3192-979d-11e6-8121-151aa99a504d.png)

In these examples, we see that the tutorial descriptions is making reference to some "steps" indicating a number.

**Problem:**
- The steps in the tutorial are not numbered. There is no easy way to know wich is the step 6, and the step 7.

**Request:**

Whether add a numbering system to the tutorial, (1, 2, 3,...) or change "step 7" to "next step", and "step 6" to "previous step".
Otherwise is difficult to understand which step is the tutorial talking about.

camimonse:
this refers to the Auth0 ticket number 3359

PR #2722, opened 2016-11-04
Made the step following clearer

camimonse:
In the instructions in the tutorial, it was talking about "step 7" and "step 6" where we didn't have any numbering.
So know it is been changed to: "next step" and "the step before".
A numbering system has been not implemented, because it is not the patron comparing with other tutorials.

Also, the order of these two words "firewall" and "providers" is been changed, for it to match with the order in the snippet.


auth0build:

camimonse:
this fixes #2601

PR #2725, opened 2016-11-04
[edit]: Geo HA is an option in cloud as well

ndelgado:
GEO HA is an option in Auth0's private cloud and in customer's cloud


auth0build:

PR #2730, opened 2016-11-05
Fix wrong example

jbeurel:
Fix Metadata usage example.

The plan is stocked in `user.app_metadata.plan`, not in `user.app_metadata.hobby`

PS: Thanks for your product ;-)


auth0build:

ntotten:
Thanks for the contribution!!

PR #2731, opened 2016-11-06
Add other missing event codes

xoob:
Updates the table with a list of events codes generated from the JS source code of https://manage.auth0.com/#/logs.

This is a follow-up to #2623 because I noticed there were still events missing from the list.

Generator code is available at https://runkit.com/xoob/581f28ab7e2dca0014c1af53


auth0build:

ntotten:
Thanks for the contribution! We are changing the format of this a bit, but we will add this content shortly via #2752.

PR #2733, opened 2016-11-07
OIDC compliant /userinfo response

rolodato:
See https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims for the standard OIDC claim set definition.


auth0build:

PR #2736, opened 2016-11-07
Updating links to extension docs to point to specific headings

jeffreylees:


auth0build:

PR #2758, opened 2016-11-09
Update billing.md

lobo:
Explain the policy better.


auth0build:

lobo:
There's a previous one already. Not needing this now.

PR #2759, opened 2016-11-10
Update billing.md

lobo:
Include special case of billing information.


auth0build:

PR #2766, opened 2016-11-14
Remove base64 decoding of client secrets

jerriep:


jerriep:
@chenkie Will merge this PR when the product team has merge the code for non-base64 encoded client secrets

PR #2774, opened 2016-11-15
ID Token docs updates

mpaktiti:
Update id_token doc and other references as per pipeline2.

Modified docs:
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/tokens/id-token
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/tokens
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/architecture-scenarios/application/spa-api
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/metadata/lock
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/protocols/oauth2/oauth-implicit-protocol
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/tutorials/web-apps-vs-web-apis-cookies-vs-tokens
- https://auth0-docs-content-pr-2774.herokuapp.com/docs/user-profile/user-profile-details


articles/protocols/oauth2/oauth-implicit-protocol.md
```diff
@@ -30,9 +30,7 @@ The steps are quite simple though:
         &state=VALUE_THAT_SURVIVES_REDIRECTS</code></pre>
 
 
-  Your code would then parse the hash segment of the URL and extract the parameters: `access_token` and `id_token`.
-
-  > The `access_token` can then be used to call Auth0's `userinfo` endpoint to get the attributes of the user.
+  Your code would then parse the hash segment of the URL and extract the parameters: `access_token` and `id_token`. The `access_token` can then be used to call Auth0's `userinfo` endpoint to get the attributes of the user.
```
mpaktiti: @cabralmartin The [endpoint](https://auth0.com/docs/api/authentication#!#get--userinfo) mentions access_token but I want to confirm that this won't change with p2.

cabralmartin: this won't change

articles/libraries/lock/v9/types-of-applications.md
```diff
@@ -10,7 +10,7 @@ You can use Auth0 Lock with both Single Page Apps and with Regular WebApps. In t
 
 ## Single Page App
 
-When you're using a Single Page App, you need that after a successful login through Auth0, you get the `id_token` ([JWT](/jwt)) with which you can then call your API. When you want to use [this workflow](/sequence-diagrams), the `responseType` is then a `token`.
+When you're using a Single Page App, you may need an `access_token`, with which you can call your API, and an `id_token`, which contains user profile attributes like the user's name, email, and so forth, typically used for UI display. When you want to use [this workflow](/sequence-diagrams), the `responseType` should be `id_token`.
```
cabralmartin: I'm thinking we should undo changes to lock9 docs since lock9 does not support api authorization (e.g. `response_type:id_token` is not supported)

articles/tokens/index.md
```diff
@@ -49,7 +49,7 @@ The Refresh token is a long-lived token that is used to obtain a new [`id_token`
 
 ## Delegation Tokens
 
-The Auth0 [`id_token`](#auth0-idtoken-jwt)'s can be exchanged for another token, called a Delegation Token, that can be used to call either other application APIs registered as clients in the same Auth0 tenant or APIs represented by some types of application Addons registered in the same Auth0 tenant.
+The Auth0 [`id_token`](#id-token)'s can be exchanged for another token, called a Delegation Token, that can be used to call either other application APIs registered as clients in the same Auth0 tenant or APIs represented by some types of application Addons registered in the same Auth0 tenant.
```
cabralmartin, line 52: we need to validate this still happens on p2

articles/tokens/id-token.md
```diff
@@ -0,0 +1,90 @@
...
+
+::: panel-warning OAuth2 as a Service
+If you are not using *OAuth2 as a Service*, then you should use `code` for server side flows and `token` for client side flows.
+:::
+
+The attributes included in the issued `id_token` are controlled by the use of a [parameter called `scope`](/scopes).
+- If `scope` is set to `openid`, then the `id_token` will contain only the `iss`, `sub`, `aud`, `exp` and `iat` claims.
+- If `scope` is set to `openid name email`, then the `id_token` will contain additionally the `name` and `email` claims.
+
+__NOTE:__ The `id_token` claims are attributes within the user profile. You can add any attribute as part of the `app_metadata` or `user_metadata` and get its value once you decode the `id_token` in your app.
```
cabralmartin: this is no longer valid. you cannot request user_metadata or app_metadata in the scope.

articles/tutorials/web-apps-vs-web-apis-cookies-vs-tokens.md
```diff
@@ -16,6 +16,6 @@ description: This page compares web applications to web APIs and cookies vs. Tok
 
 * For both approaches you can get the **same amount of information from the user**. That's controlled by the `scope` parameter sent in the login request (either using the [Auth0Lock](/lock), our [JavaScript library](https://github.com/auth0/auth0.js) or a plain link). The `scope` is a parameter of the `.signin({scope: 'openid name email'})` method which ends up being part of the querystring in the login request. You can get more details about this in the [Scopes Documentation](/scopes).
 
-* By default we use `scope=openid` in **token-based authentication** to avoid having a huge token (since the token will be traveling in the `Authorization` header for each request as mentioned above). You can control the attributes that you want to get in the token by doing `scope=openid name email another_attribute`.
+* By default we use `scope=openid` in **token-based authentication** to avoid having a huge token. You can control the attributes that you want to get in the token by doing `scope=openid name email another_attribute`.
```
cabralmartin: clients can only request [standard oidc claims](http://openid.net/specs/openid-connect-core-1_0.html#StandardClaims) as scope values. from the total set, we need to double check those that auth0 supports (i think all but i need to double check)

articles/libraries/lock/v9/types-of-applications.md
mpaktiti: OK I'll remove it from the PR

articles/tokens/index.md
rolodato: Anything related to delegation is a legacy doc. It doesn't make sense to use cross-client nor third-party delegation on the new pipeline.

articles/protocols/oauth2/oauth-implicit-protocol.md
rolodato: The access token AFAIK can be used to call `/userinfo` only if the `openid` scope was requested and approved, otherwise it will only be valid for calling the resource server (if any resource server scopes were granted).

articles/architecture-scenarios/application/spa-api.md
```diff
@@ -2,27 +2,31 @@
...
 # SPA + API
 
-![](/media/articles/architecture-scenarios/spa-api.png)
+![SPA + API Flow](/media/articles/architecture-scenarios/spa-api.png)
 
-In this scenario you have a Single Page Web Application “Client” which talks to an API (“Resource Server”). The application will use OpenID Connect with the Implicit Grant Flow to authenticate users with Auth0.
+In this scenario you have a Single Page Web Application ("Client") which talks to an API ("Resource Server"). The application will use **OpenID Connect** with the **Implicit Grant Flow** to authenticate users with Auth0. Note that this flow can only be used for Clients whose type is **Single Page Application** in the [Dashboard](${manage_url}).
 
-When a user logs in, Auth0 will return to the application an `access_token` as well as an `id_token`. The `id_token` is used to securely call the API on behalf of the user. Alternatively the user profile can be obtained by calling the `/userinfo` endpoint in the Auth0 Authentication API with the `access_token`.
+When a user logs in, Auth0 will return to the application an `access_token` and optionally an `id_token`:
```
rolodato: This is the other way around - an ID token will always be returned for the client. An access token always will be returned unless the response type was `id_token`: https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse

articles/architecture-scenarios/application/spa-api.md
```diff
@@ -2,27 +2,31 @@
...
 
-The application will usually store the information about the user’s session (i.e. whether they are logged in, their tokens, user profile data, etc) inside some sort of storage such a Local Storage.
+- The `access_token` is used to securely call the API on behalf of the user.
+
+- The `id_token` is consumed only by the client and contains user profile data. Alternatively the user profile can be obtained by calling the `/userinfo` endpoint in the Auth0 Authentication API with the `access_token`.
```
rolodato: Need to mention that `/userinfo` is only callable if `openid` was granted as a scope.

articles/metadata/lock.md
```diff
@@ -49,7 +49,7 @@ if (id_token) {
 
 You can [update the metadata properties](/metadata/apiv2#update-user-metadata) with calls to the Auth0 Management API.
 
-By including the user's `id_token` in the `Authorization` header, you can make the appropriate `PATCH` call to the [Update a user](/api/management/v2#!/Users/patch_users_by_id) endpoint. 
+By including the user's `id_token` in the `Authorization` header, you can make the appropriate `PATCH` call to the [Update a user](/api/management/v2#!/Users/patch_users_by_id) endpoint.
```
rolodato, line 52: I believe this will be removed in the new pipeline, making a note to confirm this.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
+---
+title: ID token
+description: How to obtain, use and renew an id_token.
+---
+
+# ID Token
+
+## Overview
+
+The ID token, usually referred to as `id_token` in code samples, is a [JSON Web Token (JWT)](/jwt) that contains user profile attributes represented in the form of _claims_. These claims are statements about the user, which can be trusted if the consumer of the token can verify its signature (which was generated with the Auth0 client's _Client Secret_). The `id_token` is consumed by the client and used to get user information like the user's name, email, and so forth, typically used for UI display. It was added to the OIDC specification as an optimization so the client can know the identity of the user, without having to make an additional network requests.
```
rolodato: The ID token can also be signed with a private key and verified with a public key - some clients such as SPAs are not capable of storing secrets so they don't have a client secret at all.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+
+The `id_token` conforms to an industry standard (IETF [RFC 7519](https://tools.ietf.org/html/rfc7519)) and contains three parts: a header, a body and a signature.
+- The header contains the type of token and the hash algorithm used on the contents of the token.  
+- The body, also called the payload, contains identity claims about a user.  There are some claims with registered names, for things like the issuer of the token, the subject of the token (who the claims are about), and the time of issuance.  Any number of additional claims with other names can be added, though care must be taken to keep the JWT within the browser size limitations for URLs.
```
rolodato: ID tokens are not always returned in URLs, for example when using `response_type=code` or `response_mode=form_post`. When using these modes it doesn't really matter how large the ID token becomes, because it won't be sent anywhere (the access token, if any, will be used to perform API requests).

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+- The body, also called the payload, contains identity claims about a user.  There are some claims with registered names, for things like the issuer of the token, the subject of the token (who the claims are about), and the time of issuance.  Any number of additional claims with other names can be added, though care must be taken to keep the JWT within the browser size limitations for URLs.  
+- The signature which is used by the recipient of a JWT to validate the integrity of the information conveyed in the JWT.
+
+## How to get an ID token
+
+The `id_token` can be returned when calling any of the Auth0 functions which invoke authentication.  This includes calls to the Lock widget, to the auth0.js library, or the libraries for other languages. You can view the implementation details for retrieving the `id_token` at the [Lock web library](/libraries/lock) and [Auth0.js library](/libraries/auth0js) documents.
+
+## How to control the contents of an ID token
+
+In order to retrieve an `id_token` the `responseType` should be set to `id_token`, both for client-side and server-side authentication flows.
```
rolodato, line 33: The response type needs to *include* `id_token`, since `token id_token` is a valid response type.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+
+::: panel-warning OAuth2 as a Service
+If you are not using *OAuth2 as a Service*, then you should use `code` for server side flows and `token` for client side flows.
+:::
+
+The attributes included in the issued `id_token` are controlled by the use of a [parameter called `scope`](/scopes).
+- If `scope` is set to `openid`, then the `id_token` will contain only the `iss`, `sub`, `aud`, `exp` and `iat` claims.
+- If `scope` is set to `openid name email`, then the `id_token` will contain additionally the `name` and `email` claims.
```
rolodato: * `name` is no longer a valid scope, `profile` must be used instead. In this case all these claims will be returned: https://openid.net/specs/openid-connect-core-1_0.html#ScopeClaims

* `email` scope will cause the `email_verified` scope to be included as well

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+
+If you are using Lock, the `options` object used in Lock’s instantiation can specify optional [authentication parameters](/libraries/lock/v10/customization#auth-object-) as follows:
+
+```js
+var options = {
+  auth: {
+    responseType: 'id_token',
+    params: {scope: 'openid name email'}
```
rolodato: See above comment about `name` scope

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+var lock = new Auth0Lock(
+  ${account.clientId},
+  ${account.namespace},
+  options
+);
+
+lock.show();
+```
+
+Again, the `responseType` must be set to `id_token` in order to get one back. The `id_token` will contain only the claims specified as the value of the `scope` parameter (in this example, `openid name email`).
```
rolodato: See above comments about `id_token` response type and `name` claim

articles/user-profile/user-profile-details.md
```diff
@@ -91,7 +91,7 @@ Finally, there is the Authentication API specifically used for authentication fl
 
 In the authentication flows described above, Auth0 returns a set of tokens in lieu of a full User Profile.
 
-One of the returned tokens is the `id_token`, which is a [JSON Web Token](/jwt) (or JWT) that contains User Profile attributes represented in the form of *claims*. These claims are statements about the user, which can be trusted if the consumer of the token can verify its signature (which was generated with the Auth0 app's Client Secret). The app can then use the JWT to securely call other APIs as long as those APIs can verify the JWT's signature and trust and use the contained claims.
+One of the returned tokens is the `id_token`, which is a [JSON Web Token](/jwt) (or JWT) that contains User Profile attributes represented in the form of *claims*. These claims are statements about the user, which can be trusted if the consumer of the token can verify its signature (which was generated with the Auth0 app's Client Secret). The app can then decode the JWT and get the user information contained in its payload, like the user's name, email, and so forth, typically used for UI display.
```
rolodato: The ID token can also be signed with a private key and verified with a public key - some clients such as SPAs are not capable of storing secrets so they don't have a client secret at all.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,88 @@
...
+
+**NOTE:** You can also add claims to the `id_token` using [Rules](/rules), with the following format: `context.idToken['http://my-custom/claim'] = 'some-value'`.
+
+## Lifetime
+
+The `id_token` is valid for 10 hours (36000 seconds) by default.  The value can be changed in the [Dashboard > Clients > Settings](${manage_url}/#/clients/${account.clientId}/settings) screen using the `JWT Expiration (seconds)` field.
+
+There are cases where you might want to renew your `id_token`. The most common case is that you want to refresh the claims contained in the `id_token` and see that they are still valid. For example, the `id_token` may contain the user's authorization claims and you want to check that their rights are valid before allowing sensitive operations.
+
+In order to renew the `id_token` you can either perform another authorization flow with Auth0 (using the `/authorize` endpoint) or use a [Refresh Token](/tokens/refresh-token).
```
rolodato, line 73: We should mention `prompt=none` and `silentAuthentication` from auth0.js here, since that will be desirable for SPAs.

articles/protocols/oauth2/oauth-implicit-protocol.md
mpaktiti: I'll add a note.

articles/tokens/index.md
mpaktiti: Should we remove delegation token docs or just remove the reference that you can use them to call other APIs? I know that Addons will be replaced by APIs (I assume at that point the delegation tokens will become obsolete) but I'm not clear on when this is going to happen.

articles/tokens/id-token.md
mpaktiti: Shouldn't I remove this altogether? From what I understood from an earlier comment the only effect that `responseType = id_token` will have is to exclude the `access_token` from the response.

articles/tokens/index.md
rolodato: Yes, this ultimately depends on the timeline. Delegation docs will be legacy documents, so any docs on the new pipeline should not reference them at all.

articles/tokens/id-token.md
mpaktiti: Thanks @rolodato , can you check if I wrote this right?

@cabralmartin We should also update this doc I assume https://auth0.com/docs/libraries/auth0js but I'm not sure if it should be done now or after the rollout.

rolodato (approved):
LGTM

articles/tokens/id-token.md
rolodato: > For example, the `id_token` may contain the user's authorization claims and you want to check that their rights are valid before allowing sensitive operations

This example might be misleading, since in many cases the ID token is not used for authorization (access token is used instead). I would remove it.

mpaktiti: Thanks @rolodato , I updated this here https://github.com/auth0/docs/pull/2953/files I'm going to close this PR (I messed up the last rebase) and continue working on the new one.

PR #2785, opened 2016-11-16
Removed double-interpolation in snippets

nkohari:
Fixes https://github.com/auth0/auth0-docs/issues/1225.

Before the new document compilation pipeline, the contents of snippets were processed as templates. Since the HS256/RS256 snippets contained interpolation statements that are replaced by the management site, these statements had to be double-interpolated in order to "escape" them, such as:

```
${'<%= api.signing_secret %>'}
```

The new pipeline doesn't treat the raw snippet content as a template, only attempting to replace variables if they are actually included in a document via a call to `snippet()`. This means the double-interpolation is no longer needed.

An example of this behavior can be seen in the API response here:
https://auth0.com/docs/meta/snippets/api-auth/hs256/csharp

### DIFF
--- a/snippets/api-auth/hs256/csharp.md
+++ b/snippets/api-auth/hs256/csharp.md
@@ -7,12 +7,12 @@ public class Startup
 {
   public void Configure(IApplicationBuilder app, IHostingEnvironment env, ILoggerFactory loggerFactory)
   {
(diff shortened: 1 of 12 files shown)


nkohari:
Disregard, we're going to re-implement template replacement on snippet content in the API.

Issue #2405, opened 2016-09-29
Update Access Token document


cabralmartin:
[https://auth0.com/docs/tokens/access_token](https://auth0.com/docs/tokens/access_token)
- Update content to reflect the new Access Tokens auth0-server can now issue.

PR #2800, opened 2016-11-18
Access token docs updates

mpaktiti:
Update access-token doc according to pipeline2 changes.


articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+
+By default, Auth0 generates access tokens in JSON Web Token (JWT) format, an industry standard. JWTs contain three parts: a header, a set of claims, and a signature:
+- The header contains metadata about the type of token and the cryptographic algorithms used to secure its contents.
+- The set of claims contains verifiable security statements such as the identity of the user and the permissions they are allowed.
+- The signature is used to validate that the token is trustworthy and has not been tampered with.
+
+It should be noted that other token formats are also available. For example, Auth0 Management API v1 (which has been deprecated) uses an opaque token format in which claims are referenced in a separate database, rather than directly in the token. Simple Web Token is an example of another token format, although it has not seen widespread adoption. Support for future tokens formats will be implemented if and when industry consensus develops around them.
```
rolodato: "Other token formats are available" seems to imply that it's possible to configure Auth0 APIs/clients to return tokens in a different format. Since this is not the case as of today, I would reword this paragraph.

The point that we need to try to get across is that access tokens are opaque to clients.

articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+
+In this case the user will be prompted to permit read and write access. If allowed, an access token will be issued to the client, which the client can then use when making requests to the Calendar API. If consent has already been granted by the user, no consent dialog will be displayed and the access token will be issued without additional prompts.
+
+Also, the consent dialog might be displayed again if the access level changes. For example, if the user has granted read access but the functionality changes so write access is required as well, the user will have to use the consent dialog to grant the additional access.
+
+In some cases, consent can also be pre-configured administratively. This typically occurs when the user is an employee, and there are a set of company-specific applications that are always allowed access to employee data.
+
+### Server-to-server interactions
+
+Access tokens can also be issued directly to applications, permitting access to APIs that don't provide user-specific data. Such scenarios involve server-to-server interactions.
```
rolodato: I would rephrase to something like this - "Access tokens can also be issued directly to applications without an actual user needing to authenticate". It's perfectly feasible to have a single API that is called in the same way by both users and headless applications.

articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+
+```
+HTTP/1.1 200 OK
+Content-Type: application/json
+
+{
+  "access_token":"ey...",
+  "token_type":"Bearer",
+  "expires_in":86400,
+  "example_parameter":"example_value"
```
rolodato: What is `example_parameter` here?

articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+}
+```
+
+In order to obtain this access token, the client must first have been granted permission to access the geocoding API. This is typically done by requesting access from the administrator of the geocoding API.
```
rolodato: We should describe what "requesting access" means here (creating a client, authorizing that client to make calls via client credentials, share client ID and secret).

articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+Host​: api.example.com
+
+Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwczovL2V4YW1wbGUuYXV0aDAuY29tLyIsImF1ZCI6Imh0dHBzOi8vYXBpLmV4YW1wbGUuY29tL2NhbGFuZGFyL3YxLyIsInN1YiI6InVzcl8xMjMiLCJpYXQiOjE0NTg3ODU3OTYsImV4cCI6MTQ1ODg3MjE5Nn0.CA7eaHjIHz5NxeIJoFK9krqaeZrPLwmMmgI_XiQiIkQ
+```
+
+The token in this example decodes to the following claims:
+
+```json
+{
+  "alg": "HS256",
```
rolodato: Any reason why we chose to use HS256 instead of RS256 here?

articles/tokens/access-token.md
```diff
@@ -0,0 +1,142 @@
...
+  "iss": "https://example.auth0.com/",
+  "aud": "https://api.example.com/calandar/v1/",
+  "sub": "usr_123",
+  "scope": "read write",
+  "iat": 1458785796,
+  "exp": 1458872196
+}
+```
+
+Before permitting access to the API using this token, the API must do the following:
```
rolodato, line 120: The API also needs to validate that the incoming token has the correct scopes to perform the requested operation.

articles/tokens/access-token.md
mpaktiti: I believe it refers to the [Defining New Endpoint Parameters](https://tools.ietf.org/html/rfc6749#section-8.2) section of the spec. I'm thinking of removing it however, it's confusing in this context.

articles/tokens/access-token.md
mpaktiti: Not really, I'll change that.

articles/tokens/access-token.md
mpaktiti: I'll add a link [to the doc](https://auth0.com/docs/api-auth/config/using-the-auth0-dashboard) that offers more details on this.

mpaktiti:
@rolodato Thanks for the review! I updated the PR.

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+By default, Auth0 generates access tokens in JSON Web Token (JWT) format, an industry standard. JWTs contain three parts: a header, a set of claims, and a signature:
+- The header contains metadata about the type of token and the cryptographic algorithms used to secure its contents.
+- The set of claims contains verifiable security statements such as the identity of the user and the permissions they are allowed.
+- The signature is used to validate that the token is trustworthy and has not been tampered with.
+
+## How to get an access token
+
+Access tokens are issued via [Auth0's OAuth 2.0 endpoints](/api/authentication): `/authorize` and `/oauth/token`. You can use any OAuth 2.0-compatible library to obtain access tokens. If you do not already have a preferred OAuth 2.0 library, Auth0 provides libraries for many language and frameworks that work seamlessly with our endpoints.
```
cabralmartin: s/language/languages

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+description: This page explains an overview about Auth0 access tokens.
+---
+
+# Access Token
+
+## Overview
+
+The Access Token, commonly referred to as `access_token` in code samples, is a credential that can be used by a client to access an API. The `access_token` should be used as a `Bearer` credential and transmitted in an HTTP `Authorization` header to the API. Auth0 uses access tokens to protect access to the Auth0 Management API.
+
+By default, Auth0 generates access tokens in JSON Web Token (JWT) format, an industry standard. JWTs contain three parts: a header, a set of claims, and a signature:
```
cabralmartin: Also note the other token format (opaque) that Auth0 can issue for use at `/userinfo` of the deprecated APIv1

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+* Calls to the Lock widget will return an `access_token` as shown in the [Lock documentation](/libraries/lock).
+* [Examples using auth0.js](https://github.com/auth0/auth0.js).
+* Check the [List of tutorials](/tutorials) to see how to make calls to libraries for other languages/SDKs.
+
+## How to use an access token
+
+Access tokens are typically obtained in order to access user-owned resources. For example, a Calendar application on a mobile phone needs access to a Calendar API in the cloud in order to read the user's scheduled events and create new events.
+
+Such access is requested by the application and granted by the user. For example, when using Lock, access to read and write the user's calendar can be requested by calling `show()`:
```
cabralmartin: s/by the application/by the client (or by the client application)

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+```js
+lock.show({
+  responseType: 'token',
+  authParams: {
+    scope: 'read write'
```
cabralmartin: This is missing the `audience` param with the API audience. Without this param, the request will go thru the legacy pipeline.

Maybe we can replace the usage of lock with a call to `/authorize`. This will be better since Lock still has a lot of caveats when it comes to using `audience` (e.g. support for 3rd party hosted lock is not yet completed)

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+In this case the user will be prompted to permit read and write access. If allowed, an access token will be issued to the client, which the client can then use when making requests to the Calendar API. If consent has already been granted by the user, no consent dialog will be displayed and the access token will be issued without additional prompts.
+
+Also, the consent dialog might be displayed again if the access level changes. For example, if the user has granted read access but the functionality changes so write access is required as well, the user will have to use the consent dialog to grant the additional access.
+
+In some cases, consent can also be pre-configured administratively. This typically occurs when the user is an employee, and there are a set of company-specific applications that are always allowed access to employee data.
+
+### Server-to-server interactions
+
+Access tokens can also be issued directly to applications. Such scenarios involve server-to-server interactions. In this case the user does not need to authenticate.
```
cabralmartin: s/applications/clients

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+For example, a reverse geocoding API that accepts a latitude, longitude coordinate and returns a readable place name does not access user-owned data. In such cases a backend server needs to call the geocoding API in order to perform the translation.
+
+Server-to-server access tokens can be obtained using the [Client Credentials flow](/api-auth/grant/client-credentials). In order to get a token using this flow, the Client has to provide its credentials (`client_id`, `client_secret`).
+
+```
```
cabralmartin: Replace with HAR so we get the HTTP Snippet selector.

articles/tokens/access-token.md
```diff
@@ -0,0 +1,144 @@
...
+
+In some situations, it is desirable to have permanent, ongoing access to an API without having to repeat an OAuth flow. This is often referred to as `offline access`, and is possible with the use of a [refresh token](/tokens/refresh-token).
+
+A refresh token is issued from the OAuth 2.0 endpoints along with the access token. When the access token expires, the refresh token can be used to obtain a fresh access token with the same permissions, without further involvement from a user. Note that offline access is enabled as a policy of the API the access token grants access to. If the API does not permit offline access, a refresh token will not be issued. In such circumstances, the OAuth flow must be repeated in order to obtain a new access token.
+
+For more information on refresh tokens and how to use them refer to: [Refresh Token](/tokens/refresh-token).
+
+## Revoke access
+
+Revoking access is not supported at the moment.
```
cabralmartin: Worth mentioning that the best mechanism to revoke access is the token expiration itself. Customer should choose a sensible expiration time according to their domain. e.g. a banking API might have shorter lived token vs a TODO list API.

mpaktiti:
Fixes https://github.com/auth0/docs/issues/2405

PR #2802, opened 2016-11-18
Refresh Tokens

mpaktiti:
Update refresh-token doc according to pipeline2 changes.


articles/tokens/index.md
```diff
@@ -41,11 +41,11 @@ When a user authenticates via Auth0 with another social provider's authenticatio
 
 [More information](/tokens/idp)
 
-## Auth0 `refresh_token`
+## Refresh Token
 
-The Refresh token is a long-lived token that is used to obtain a new [`id_token`](#auth0-idtoken-jwt) after a previous one has expired.  This is useful for applications running on mobile devices that call remote APIs and do not want to require the user to log in every time the user uses the mobile app.
+The Refresh Token, commonly referred to as `refresh_token`, is a long-lived [JWT](/jwt) that is used to obtain a new [id_token](#id-token) or a new [access token](#access-token), after a previous one has expired.
```
rolodato: Refresh tokens are not necessarily JWTs. Auth0 might return them as JWTs, but client applications should have no expectation about their format.

articles/tokens/refresh-token.md
```diff
@@ -1,22 +1,25 @@
...
-description: A refresh token allows an application to request Auth0 to issue a new id_token directly, without needing to re-authenticate the user.
+title: Refresh Token
+description: What is a Refresh Token and what you can do with it.
 ---
 
-# Refresh Tokens
+# Refresh Token
 
-A **refresh token** is a special kind of [JWT](/jwt) that is used to authenticate a user without them needing to re-authenticate. This is primarily useful for mobile applications that are installed on a device.  
+A **refresh token** is a special kind of [JWT](/jwt) that contains the information required to obtain a new `access_token` or `id_token`.
```
rolodato: Refresh tokens are not necessarily JWTs. Auth0 might return them as JWTs, but client applications should have no expectation about their format.

articles/tokens/refresh-token.md
```diff
@@ -1,22 +1,25 @@
...
 
-For more information on the types of access tokens used by Auth0, see [Tokens](/tokens).
+> An `id_token` is used by the client that receives it in order to identify the user. An `access_token` is used by a client to access an API. For more information refer to: [Tokens](/tokens).
 
-If you are new to refresh tokens, you can learn more about them in this blog post: [Refresh Tokens: When to Use Them and How They Interact with JWTs](https://auth0.com/blog/refresh-tokens-what-are-they-and-when-to-use-them/).
+Usually, a user will need a new access token only after the previous one expires, or when gaining access to a new resource for the first time.
 
-## Introduction
+Refresh tokens are subject to strict storage requirements to ensure that they are not leaked. Refresh tokens can also be blacklisted by the authorization server.
```
rolodato: This should link to the doc that describes how to revoke refresh tokens.

We should be consistent in our usage of "revoking" vs "blacklisting". The OIDC spec uses "revoke", so I would suggest we use that (see example use here: https://openid.net/specs/openid-connect-core-1_0.html#TokenLifetime)

articles/tokens/refresh-token.md
```diff
@@ -1,22 +1,25 @@
...
+## Overview
 
-A refresh token allows the application to request Auth0 to issue a new `id_token` directly, without needing to re-authenticate the user. This will work as long as the refresh token has not been revoked.
+The response of an [authentication request](/protocols) can result in an `access_token` and/or an `id_token` being issued by Auth0. The  `access_token` is used to make authenticated calls to a secured API, while the `id_token` contains user profile attributes represented in the form of _claims_. Both JWTs have an expiration date indicated by the `exp` claim (among other security measures, like signing).
 
-Refresh tokens can be issued for each combination of __app__, __user__ and __device__. Once the Auth0 refresh token is issued, the values of the client, user, and device set during its creation cannot be changed.
+A refresh token allows the application to request Auth0 to issue a new `access_token` or `id_token` directly, without needing to re-authenticate the user. This will work as long as the refresh token has not been revoked.
```
rolodato: We should be consistent in our usage of "revoking" vs "blacklisting". The OIDC spec uses "revoke", so I would suggest we use that (see example use here: https://openid.net/specs/openid-connect-core-1_0.html#TokenLifetime)

articles/tokens/refresh-token.md
```diff
@@ -1,22 +1,25 @@
...
+
+Refresh tokens can be issued for each combination of app, user and device. Once the Auth0 refresh token is issued, the values of the client, user, and device set during its creation cannot be changed.
```
rolodato: The OIDC spec doesn't mention this anywhere AFAIK. What are the practical implications of issuing a refresh token per app/user/device? I think we can remove this sentence altogether.

articles/tokens/refresh-token.md
mpaktiti: I will be using "revoke".

mpaktiti:
Thanks for the review @rolodato ! I updated the PR.

articles/tokens/refresh-token.md
```diff
@@ -61,45 +62,22 @@ The refresh token is returned as part of the URL, in the form of an opaque strin
...
-```json
-{
-  "token_type": "Bearer",
-  "expires_in": 36000,
-  "id_token": "eyJ..."
-}
-```
-
-The `expires_in` parameter indicates the lifetime of the new JWT in seconds. It can be calculated by the difference between the `exp` and `iat` claims of the JWT.
+To obtain a new `id_token`, call the `/oauth/token` endpoint in the Authentication API, using `grant_type=refresh_token`.
```
mpaktiti: Add sample request and response.

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
+### Revoke a refresh token using the dashboard
 
-To see if a user has existing devices with associated refresh tokens, go to the [Users section](${manage_url}/#/users) of the dashboard. Click the name of the user to view their **Details** page.
+To revoke a refresh token, go to the [Users section](${manage_url}/#/users) of the [dashboard](${manage_url}). Click the name of the user to view their *Details* page.
 
-Select the **Devices** tab. This page lists all device names and the number of refresh tokens associated with each. To revoke a refresh token, click the **X** to the right of the device name.
+Select the *Authorized Applications* tab. This page lists all the clients to which the user has authorized access. Revoking an authorized application revokes also its associated refresh tokens.
 
-![](/media/articles/tokens/dashboard-revoke-refresh-token.png)
+To revoke a refresh token, click **Revoke**.
```
cabralmartin: Subtle difference, we don't revoke the token in the dashboard. We delete the underlaying `user grant`. This is grant is used to validate that a refresh token should get a new access token / id token for that particular app. If the grant gets revoked, then the refresh token becomes useless.

Maybe we can switch the writing here to talk about "Revoking access to an Authorized Application will cause any refresh token issue to that application to become invalid."

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
 
-Click **UNLINK** to confirm.
+![Revoke a refresh token using the dashboard](/media/articles/tokens/dashboard-revoke-refresh-token.png)
 
 ## SDK Support
 
-The [Lock](/libraries/lock), [auth0.js](/libraries/auth0js), and [auth0-angular.js](https://github.com/auth0/auth0-angular) libraries include support to obtain and use refresh tokens.
+The [Lock](/libraries/lock), [auth0.js](/libraries/auth0js), and [auth0-angular.js](https://github.com/auth0/auth0-angular) libraries include support methods to obtain and use refresh tokens.
```
cabralmartin: Remove this p. The mantra is "no refresh tokens should be placed in the browser". If any of our front end SDK support this we need to remove it.

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
     access_token=2nF...WpA
     &id_token=eyJhb...
     &state=VALUE_THAT_SURVIVES_REDIRECTS
     &refresh_token=Cqp...Mwe
 ```
 
-The refresh token is returned as part of the URL, in the form of an opaque string.
+The refresh token (an opaque string) is part of the URL. You should store it securely and use it only when needed.
+
+**NOTE**: In this example, the token was returned to the client in the URL because the [implicit grant](/api-auth/grant/implicit) (`response_type=token`) was used.
```
cabralmartin: The implicit grant cannot issue refresh tokens. Only Authorization Code, PKCE and password grant.

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
 ---
-description: A refresh token allows an application to request Auth0 to issue a new id_token directly, without needing to re-authenticate the user.
+title: Refresh Token
+description: What is a Refresh Token and how you can use it.
 ---
 
-# Refresh Tokens
+# Refresh Token
```
cabralmartin: Let's add a p saying that refresh tokens are only for Authorization Code, PKCE and password/pwd realm grants. As refresh tokens are super sensitive (you can think of keeping a refresh token as keeping the user u&p) they cannot be placed in a browser.

cabralmartin: The corollary of this is that SPA cannot, under no conditions, use refresh token. SPA should use silentAuth to refresh session/access.

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
 :::
 
-Refresh tokens can be [obtained](#obtain-a-refresh-token) or [revoked](#revoke-a-refresh-token-using-the-management-api) programmatically through the Auth0 API.
+Refresh tokens can be [obtained](#obtain-a-refresh-token) or [revoked](#revoke-a-refresh-token-with-a-request) programmatically through the Auth0 API.
 
-They can also be viewed and revoked [from the dashboard](#revoke-a-refresh-token-in-the-dashboard).
+They can also be viewed and revoked [from the dashboard](#revoke-a-refresh-token-using-the-dashboard).
 
-## Obtain a Refresh Token
+## Get a Refresh Token
```
cabralmartin: Here things start to diverge a lot, expected, from the legacy flow. Maybe it's worth to keep the old doc under `/legacy/refresh-tokens` and link from this one.

"This article was rewritten to the new Refresh Token story introduced by the API Authorization Features. If you are looking for the legacy Refresh Token docs you can check here (link)"

Can you sync with Nathan to validate this? We agreed on doing this in the past but never got a doc worth doing it. This can be the first.

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
 ```
 
-A response from this request could be as follows:
+Where:
+- `grant_type`: The type of grant to execute (the `/token` endpoint is used for various grants, for more information refer to the [Authentication API](/api/authentication#get-token)). To refresh a token use `refresh_token`.
+- `client_id`: Your application's Client ID.
+- `client_secret`: Your application's Client Secret.
+- `refresh_token`: The refresh token to use.
+
+The response will include a new `access_token`, its type, its lifetime (in seconds), and the granted scopes. If the scope of the initial token included `openid`, then a new `id_token` will be in the response as well.
```
cabralmartin: 💯

articles/tokens/refresh-token.md
```diff
@@ -157,3 +178,12 @@ For more information about using refresh tokens with these libraries, see:
 * [Get a refresh token with Auth0.js](https://github.com/auth0/auth0.js#login) and [Use a refresh token to get new id_token](https://github.com/auth0/auth0.js#refresh-token)
```
cabralmartin: These links are no longer valid.

articles/tokens/refresh-token.md
cabralmartin: We can change this section to say something that our main SDKs support refresh tokens out of the box (maybe name a few like node, c#, java).

articles/tokens/refresh-token.md
cabralmartin: Let's add a section talking about the "refresh story" for SPA (The `renewAuth` method in the API, all documentation warnings apply https://github.com/auth0/auth0.js#api)

articles/tokens/refresh-token.md
```diff
@@ -1,152 +1,173 @@
...
 
-**NOTE**: In this case, the token was returned to the client directly in the URL because the [implicit flow](/protocols#oauth2-implicit-flow) (`response_type=token`) was used.
+::: panel-info Troubleshooting
+If the response did not include a refresh token, check the **Allow Offline Access** switch on your [API Settings](${manage_url}/#/apis). It must be enabled, otherwise Auth0 won't return a refresh token, even if you included the `offline_access` scope.
+:::
 
 ## Use a Refresh Token
 
-To obtain a new `id_token`, call the [delegation](/api/authentication/reference#delegation) endpoint in the Authentication API:
+To refresh your token, using the `refresh_token` you already got during authorization, make a `POST` request to the `/oauth/token` endpoint in the Authentication API, using `grant_type=refresh_token`.
```
cabralmartin: Let's add a warning in this section saying that refresh tokens won't run rules, that's something we will add on the future.

articles/tokens/refresh-token.md
mpaktiti: I added a Restrictions paragraph, addressing the first two comments.

articles/tokens/refresh-token.md
mpaktiti: Oops. Replaced example with authorization code grant, thanks.

articles/tokens/refresh-token.md
mpaktiti: I restructured the section.

articles/tokens/refresh-token.md
mpaktiti: I added info on the SPA refresh story at SDK Support > Single Page Apps.

articles/tokens/refresh-token.md
mpaktiti: I made some changes to the paragraph.

articles/tokens/legacy/refresh-token.md
```diff
@@ -0,0 +1,153 @@
+---
+description: A refresh token allows an application to request Auth0 to issue a new id_token directly, without needing to re-authenticate the user.
+---
+
```
mpaktiti: @cabralmartinThis is the legacy doc. I removed some of the content that I suppose we don't want there (for example referring to using id_tokens to invoke an API). I left the obtain/use/revoke paragraphs as they were. Let me know what you think.

saltukalakus:
Should we also add a note for refresh token revocation for mobile apps? Here is the [discussion link](https://github.com/auth0/auth0-server/issues/3045#issuecomment-278023506) for  it. 


For clients that can not keep the client_secret safe e.g. mobile apps, the `/oauth/revoke` endpoint supports passing `tokenEndpointAuthMethod: none` within the request body for token revocation.


    {
      "token" : ..., 
      "client_id" : ...,
      "tokenEndpointAuthMethod": "none"
    }

mpaktiti:
Hey @saltukalakus thanks for the info! I will update the docs.

PR #2822, opened 2016-11-22
Add API Auth docs for ASP.NET Quickstarts

jerriep:


articles/server-apis/aspnet-core-webapi/02-authorization.md
```diff
@@ -0,0 +1,136 @@
...
+}
+```
+
+Finally, to ensure that a scope is present in order to call a particular API endpoint, you simply need to decorate the action with the `Authorize` attribute, and pass the name of the Policy for that `scope` in the `policy` parameter:
+
+```csharp
+[Route("api/timesheets")]
+public class TimesheetsController : Controller
+{
+    [Authorize("read:timesheets")]
```
nicosabena, line 110: Shouldn't the `Authorize` attribute have the `Policy` parameter name? Like this

```
[Authorize(Policy="read:timesheets")]
```

articles/server-apis/webapi-owin/02-authorization.md
```diff
@@ -0,0 +1,120 @@
...
+    }
+    public override void OnAuthorization(HttpActionContext actionContext)
+    {
+        base.OnAuthorization(actionContext);
+
+        ClaimsPrincipal principal = actionContext.ControllerContext.RequestContext.Principal as ClaimsPrincipal;
+        if (principal != null)
+        {
+            // If user does not have the scope claim, get out of here
+            if (principal.HasClaim(c => c.Type == "scope"))
```
nicosabena, line 69: Should we check the `Issuer` as well? 
It would be a hassle to include it in the attribute's constructor. Maybe set it by default to the configured Auth0 domain? 

```
public ScopeAuthorizeAttribute(string scope)
  : this(scope, ConfigurationManager.AppSettings["Auth0Domain"])
{
}
public ScopeAuthorizeAttribute(string scope, string issuer)
{
    this.scope = scope;
    this.issuer = issuer;
}
```

jerriep: We should. Thing is however, how do you pass it in to the Attribute? 

You can only pass in a constant value as a parameter to an Attribute, which means the developers cannot configure the Issuer in the configuration file, and will instead have to hard code it at compile time.

I am open to suggestions to work around this problem. Source code is at https://github.com/auth0-samples/auth0-aspnet-owin-webapi-sample/tree/master/02-Authorization/WebApi if you want to play around with it :)

articles/server-apis/aspnet-core-webapi/02-authorization.md
jerriep: Not necessary. The first parameter is `policy`:

image

articles/server-apis/webapi-owin/02-authorization.md
jerriep: One option is perhaps to just not make it a parameter and instead read it from the `ConfigurationManager` inside the `OnAuthorization` method of the `ScopeAuthorizeAttribute` class.

Not ideal, but I suppose it is better than not checking the issuer at all.

Will wait to hear if you can think of another possibility :)

PR #2699, opened 2016-11-02
Commits for PHP Symfony

camimonse:
At the bottom of the page the instruction was not understandable. For the user was not clear when he has to download the CAs database.

Issue #2651, opened 2016-10-27
Unclear instruction in the PHP Symfony tutorial


camimonse:
In https://auth0.com/docs/quickstart/webapp/symfony at the bottom of the tutorial, as final note, it says: _"There is an issue with CAs database on your computer. Need to download this CAs database..."_

Problem:
It doesn't say whether I have to do it, or I just have to considerate it, or if somehow Auth0 knows that I actually have that problem and that is why is telling me.

Improve:
To add **"If"** at the beginning:

![image](https://cloud.githubusercontent.com/assets/18585436/19772517/7f9fd7fa-9c34-11e6-8f0a-9de6bda4d0b9.png)

camimonse:
this refers to GitHub ticket number 3518

camimonse:
Sanity passed

camimonse:
@mpaktiti This issue should be closed, cause it is been already merged. The issue is fixed :)

mpaktiti:
Thanks @camimonse ! 👍

PR #2834, opened 2016-11-24
Saas images commit has been taken out

camimonse:
Commit with the saas images have been taken out, so we can merge the other commits comming up


camimonse:
this has to do with #2699 and #2651

chenkie:
@camimonse the sceenshots should be taken on a Mac as per the docs guidelines: https://github.com/auth0/docs/blob/master/CONTRIBUTING.md#screenshots

camimonse:
@chenkie yes, actually this commit should not exist. I was reseting the commit, and I had to name it, i put the new name as "Saas images commit has been taken out", but it seems like it didn´t work. My bad. Anyway, this commit is not blocking us from merging the other commits that we had ( #2699 ), for I redid them in a different branch, so we are ok.

chenkie:
cool thanks 👍

Issue #2598, opened 2016-10-21
Information in the PHP symfony quickstart needed


camimonse:
In order to follow the instructions of the PHP Symfony quickstart tutorial:  https://auth0.com/docs/quickstart/webapp/symfony   **composer is needed but is never mentioned** as other pages such as the PHP Symfony Backend: https://auth0.com/docs/quickstart/backend/php-symfony

This picture shows how in the second web page is shown:
![image](https://cloud.githubusercontent.com/assets/18585436/19602155/13749444-977a-11e6-8c22-ee77d37f410a.png)
- It is important to know this, since in the README file and the quickstart are instructions that require composer in order to be fulfilled 
- It would be helpful for the user to have this information.
- It is also a good idea to keep the format of other pages, presenting the information the same way.

camimonse:
this refers to the Auth0 ticket number 3359

PR #2836, opened 2016-11-24
Saying composer is required

camimonse:
In the tutorial we needed to indicate that composer is been used, in
order to make it a clearer explination


camimonse:
this fixes #2598

Issue #2882, opened 2016-12-01
Single Page App JQuery Quickstart lock version is outdated on the SP


rafanog:
In the [Single Page App JQuery quickstart], the lock version is outdated on the SP's side.

 - The one referenced on the index.html
![lock version is outdated on sp](https://cloud.githubusercontent.com/assets/15782359/20807142/1824e762-b7d4-11e6-8573-7688e28c724e.JPG)

 - The one on the "Important snippets" section of the README.md
![lock outdated on sp readme](https://cloud.githubusercontent.com/assets/15782359/20807316/d50a4174-b7d4-11e6-90ce-6dd26cbd752c.JPG)

TechAid ticket: AUTH - 3353

mpaktiti:
Fixed by https://github.com/bambii7/docs/commit/81142a7a0b4369ec9803ae9896cc459a0b99e697

PR #2892, opened 2016-12-04
update lock cdn references to variable lock_url #2882

ofhope:


mpaktiti:
Looks good @bambii7 , thanks for the contribution! 👍

PR #2904, opened 2016-12-06
Revision/i18n guide android

ofhope:
Updating broken link to key value pairs.
Updating language used for clarity.
Adding meta title as per style guide.


mpaktiti (changes requested):
Hello @bambii7 ! Thank you very much for the contribution!

I will need a small change in this PR before I can merge it. If you can, please remove the `auth0android_url` variable from the config file and set the URL directly to `https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/res/values/strings.xml` on the `i18n-guide-android.md`. 

We use that config file only for libraries, so every time a new version is released we only need change that entry and not all the docs referencing the library.

ofhope:
Oh I misunderstood, I was reading in the general guide lines not to hardcode links. But that is only to auth0 sites. Removed :)

mpaktiti:
Thank you! 👍

PR #2905, opened 2016-12-06
Using Rules in Password Exchange Grant

mpaktiti:
- Move using rules with client credentials doc
- Add doc for Rules in password exchange

[Review App](https://auth0-docs-content-pr-2905.herokuapp.com/docs/api-auth/using-rules/password)

Solves https://github.com/auth0/auth0-server/issues/3031

---

- [ ] Add links of these docs to this one -> https://auth0.com/docs/auth0-hooks/extensibility-points
- [ ] Change naming to Hooks instead of Rules? https://auth0.com/docs/auth0-hooks


articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,158 @@
...
+## Creating the Rule
+
+**Note**: You can only create one rule, which will then be executed for **all** clients and APIs.
+
+### 1. Create the Rule For Use with Webtasks
+
+Create a file named `myrule.js`, and enter the following:
+
+```js
+module.exports = function(client, scope, audience, context, cb) {
```
hernanhht: The function signature for `password-exchange` should be 
```js
function(user, client, scope, audience, context, cb)
```

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,158 @@
...
+```har
+{
+  "method": "POST",
+  "url": "https://${account.namespace}/oauth/token",
+  "headers": [
+    { "name": "Content-Type", "value": "application/json" }
+  ],
+  "postData": {
+    "mimeType": "application/json",
+    "text": "{\"grant_type\":\"password\",\"username\": \"user@example.com\",\"password\": \"pwd\",\"audience\": \"https://someapi.com/api\", \"scope\": \"read:sample\", \"client_id\": \"${account.clientId}\"}"
```
hernanhht: You need to add the param `client_secret`

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,158 @@
...
+  }
+}
+```
+
+Before you make the `POST` call you must replace the following values:
+* `username`: Resource Owner's identifier.
+* `password`: Resource Owner's secret.
+* `audience`: API Identifier that the client is requesting access to.
+* `client_id`: Client ID of the client making the request.
```
hernanhht, line 75: Add `client_secret` here too

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,158 @@
...
+The contents of the decoded `access_token` look like this:
+
+```json
+{
+  "iss": "${account.namespace}/",
+  "sub": "${account.clientId}@clients",
+  "aud": "API_IDENTIFIER",
+  "exp": 1472832994,
+  "iat": 1472746594,
+  "scope": "test extra",
```
hernanhht: `"scope": "read:sample extra"`

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,158 @@
...
+
+**Note:** You can use [JWT.IO](https://jwt.io/) to decode, verify and generate JWT.
+
+## Implementation Notes
+
+### Input Parameters
+
+The input parameters for the rule, including sample snippets:
+
+* **client** (`object`): the client asking for the token, including the `client` metadata (a key-value pair that can be set by client)
```
hernanhht: Before `**client**` we need to add:
```
**user** (`object`): The authenticated user.
json
{
    "tenant": "tenant_name",
    "id": "user_id",
    "displayName": "user_name",
    "email": "email",
    "user_metadata": {
         "some_user_metadata": "value"
    },
    "app_metadata": {
         "some_app_metadata": "another value"
    }
}
```

hernanhht (changes requested):

articles/api-auth/using-rules/password.md
mpaktiti: @hernanhht Is it needed for this case only? When we have rules? Because it's not included in the RO grant doc https://auth0.com/docs/api-auth/tutorials/password-grant#executing-the-flow

hernanhht: If the `Token Endpoint Authentication Method` is set to `None` you don't need it. If it sets to `Post` you do need it.
This is related to the `/oauth/token` endpoint itself and not only for `grant_type=password`.

pasted image at 2016_12_06 12_54 pm

mpaktiti:
@hernanhht I addressed your comments and made some more updates to re-use content.

hernanhht (approved):

articles/api-auth/using-rules/client-credentials.md
```diff
@@ -0,0 +1,70 @@
+---
+description: Describes using rules with Client Credentials Grant.
+---
+
+# Using Rules with Client Credentials Grants
+
+You can now add [rules](/rules) into the [client credentials](/api-auth/grant/client-credentials) exchange pipeline where you exchange a `client_id` and `secret` for an `access_token`.
```
rolodato: I would reword this because of a few reasons:

* Most users visiting this doc will have never used client credentials, so they will not know that rules + client credentials was not possible before. So we should remove the "you can now" part.
* I would not link to the standard /rules doc because it does not give the correct guidance about setting up a webtask for this specific scenario. We might need some other terminology like "webtask" instead of rule for this specific case.

articles/api-auth/using-rules/_includes/_create-webtask.md
```diff
@@ -0,0 +1,18 @@
+Create the Webtask. You will need to set the following static metadata fields for the Webtask:
+
+* `wt-compiler = auth0-ext-compilers/${grant}`
+* `auth0-extension = runtime`
+* `auth0-extension-name = ${grant}`
+* `auth0-extension-secret = {random_secret}`
```
rolodato: `{random_secret}` does not have a `$` prefix unlike the other values - we should stick to one convention for consistency. I prefer without `$` but that's just me 😄

articles/api-auth/using-rules/_includes/_response.md
```diff
@@ -0,0 +1,24 @@
+A successful response contains an `access_token`:
+
+```js
+{
+  "access_token": "eyJz93a...k4laUWw",
+  "token_type": "Bearer"
+}
+```
+
+The contents of the decoded `access_token` look like this:
```
rolodato: "Access tokens are opaque to clients" still applies here - the only thing a client should do with access tokens is send them to the API. Decoding an access token is the responsibility of the API, not the client.

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,90 @@
+---
+description: Describes using rules with Resource Owner Password Grant.
+---
+
+# Using Rules with Resource Owner Password Grant
+
+You can now add [rules](/rules) into the [Resource Owner Password Grant](/api-auth/grant/password) exchange pipeline where you exchange a `username` and `password` for an `access_token`, and optionally a `refresh_token`.
```
rolodato: Same recommendation as above, copying for clarity:

* Most users visiting this doc will have never used client credentials, so they will not know that rules + client credentials was not possible before. So we should remove the "you can now" part.
* I would not link to the standard /rules doc because it does not give the correct guidance about setting up a webtask for this specific scenario. We might need some other terminology like "webtask" instead of rule for this specific case.

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,90 @@
...
+  }
+}
+```
+
+Before you make the `POST` call you must replace the following values:
+* `username`: Resource Owner's identifier.
+* `password`: Resource Owner's secret.
+* `audience`: API Identifier that the client is requesting access to.
+* `client_id`: Client ID of the client making the request.
+* `client_secret`: Client Secret of the client making the request. This parameter is only required when the **Token Endpoint Authentication Method** of your Non Interactive Client is set to `Post`. If it is `None` you do not need this parameter. To check the value navigate to [Dashboard Client Settings](${manage_url}/#/clients/${account.clientId}/settings).
```
rolodato: Client secret is always mandatory for non-interactive clients. This is shown as a UI option but it has no effect and will be fixed soon.

articles/api-auth/using-rules/_includes/_create-rule.md
```diff
@@ -0,0 +1,15 @@
...
+  var access_token = {};
+  access_token['https://foo.com/claim'] = 'bar';
+  access_token.scope = scope;
+  access_token.scope.push('extra');
+  cb(null, access_token);
+};
+```
+This is a sample rule that will:
+
+* Add an arbitrary claim (`https://foo.com/claim`) to the `access_token`.
```
rolodato: We should mention that claims need to be namespaced in order to show up in the token at all. I've described the namespacing rules here: https://github.com/rolodato/docs/blob/migration-guide/articles/api-auth/tutorials/migration/scope-custom-claims.md#L80, maybe we can have a common doc for that?

articles/api-auth/using-rules/_includes/_response.md
```diff
@@ -0,0 +1,24 @@
+A successful response contains an `access_token`:
+
+```js
+{
+  "access_token": "eyJz93a...k4laUWw",
+  "token_type": "Bearer"
```
rolodato: The response now includes `expires_in` and `scope`. For example:

```json
{
  "access_token": "eyJ...",
  "expires_in": 86400,
  "scope": "write:widgets read:widgets",
  "token_type": "Bearer"
}
```

articles/api-auth/using-rules/client-credentials.md
```diff
@@ -0,0 +1,70 @@
...
+<%= include('./_includes/_prerequisites.md') %>
+
+## Creating the Rule
+
+**Note**: You can only create one rule, which will then be executed for **all** clients and APIs.
+
+### 1. Create the Rule For Use with Webtasks
+
+<%= include('./_includes/_create-rule', {
+	  args: 'client, scope, audience, context, cb'
```
rolodato: This doesn't work for me, I get this log in webtask:

```json
    {
      "code": 400,
      "message": "Invalid webtask code",
      "error": "The JavaScript function must have one of the following signatures: ([ctx], callback) or (ctx, req, res)."
    }
```

articles/api-auth/using-rules/_includes/_create-webtask.md
mpaktiti: The `${grant}` is actually a variable i'm passing in this doc in order to reuse its content. The `Using Rules with Client Credentials Grants` doc passes the value `client-credentials`, while `Using Rules with Resource Owner Password Grant` passes the value `password-exchange`. It looks like that:

image

articles/api-auth/using-rules/client-credentials.md
mpaktiti: Same error here, I'm following up with Hernan, thanks.

articles/api-auth/using-rules/_includes/_create-rule.md
mpaktiti: Cool, I'll add some info now on the doc and once the one you wrote is published I will link to that.

articles/api-auth/using-rules/client-credentials.md
mpaktiti: @rolodato @hernanhht I talked with Geoff, the problem was that `wt` was creating my webtask at my 'default' profile (which, for me, points to webtask.io). I added at the end of the `wt create` command the: `-p mpaktiti-default` (where `mpaktiti` is my tenant name) and it worked. So if you already have webtask profiles you should use:
```
SECRET=$(openssl rand 32 -base64) && \
wt create myrule.js \
  --meta wt-compiler=auth0-ext-compilers/client-credentials-exchange \
  --meta auth0-extension=runtime \
  --meta auth0-extension-name=credentials-exchange \
  --meta auth0-extension-secret=$SECRET \
  --secret auth0-extension-secret=$SECRET \
  -p "TENANT-default"
```

I'll add a note at the docs.

@rolodato please try and reproduce once more, and if it does not work for you still let me know. Thanks!

mpaktiti: Resource Owner still does not work, we are investigating.

mpaktiti: Resource Owner works fine, my test was wrong. I used the same webtask with client credentials, and in the RO case this is missing the `user` input argument.

articles/api-auth/using-rules/_includes/_runtime.md
```diff
@@ -0,0 +1,10 @@
+The Auth0 Runtime expects you to return an `access_token` that looks like the following:
+
+```json
+{
+  "https://anything.com/foo": "bar",
+  "scope": [ "scope1", "scope2" ]
+}
+```
+
+If you decide not to issue the token, you can return `Error (cb(new Error('access denied')))`.
```
rolodato: Shouldn't this just be `cb(new Error('access denied'))`?

articles/api-auth/using-rules/_includes/_create-webtask.md
```diff
@@ -0,0 +1,22 @@
...
+wt create myrule.js \
+  --meta wt-compiler=auth0-ext-compilers/${compiler} \
+  --meta auth0-extension=runtime \
+  --meta auth0-extension-name=${grant} \
+  --meta auth0-extension-secret=$SECRET \
+  --secret auth0-extension-secret=$SECRET
+```
+
+::: panel-warning Troubleshooting
+If you have more than one webtask profiles, make sure that your webtask is created using your Auth0 tenant's profile. The URL of your webtask should be like `https://${account.tenant}.us.webtask.io/myrule`. To ensure that this is the case append the following at the `wt create` command: `-p ${account.tenant}-default`. This should match the profile name you used when you [installed the Webtask CLI for your tenant](${manage_url}/#/account/webtasks). If the webtask is created under another profile then you will get an `Unauthorized extensibility point` error, when you test your newly created webtask.
```
rolodato: > If you have more than one webtask profiles

should be:

> If you have more than one webtask profile

jhiner:
@mpaktiti We are going to merge an update to the Password Grant rules: https://github.com/auth0/auth0-server/pull/3232. As we discussed previously, this will allow updating both ID tokens and Access Tokens.

The response from the rule/webtask will be like this:
```
  var myCustomAccessTokenClaims = {
     scope: scope.concat('email'),
     'http://mydomain.com/foo': 'bar'
  }
  var myCustomIdTokenClaims = {
    'http://mydomain.com/biz': 'buz'
  }

  cb(null, {accessToken: myCustomAccessTokenClaims, idToken: myCustomIdTokenClaims});
```

articles/api-auth/using-rules/_includes/_runtime.md
```diff
@@ -0,0 +1,10 @@
+The Auth0 Runtime expects you to return an `access_token` that looks like the following:
```
jhiner: Here this needs updated for the new response e.g. `{accessToken: myCustomAccessTokenClaims, idToken: myCustomIdTokenClaims}`

articles/api-auth/using-rules/password.md
```diff
@@ -0,0 +1,114 @@
...
+description: How to add a webtask and run it each time a Resource Owner Password grant is executed.
+---
+
+# Using Rules with Resource Owner Password Grant
+
+You can write functions in JavaScript and have them run every time a [Resource Owner Password grant](/api-auth/grant/password) is executed, using [webtasks](https://webtask.io/).
+
+## Before you begin
+
+<%= include('./_includes/_prerequisites.md') %>
```
AmaanC: Why is having a Non-Interactive Client and an API configured a pre-requisite to Rules for the `password` grant?

articles/api-auth/using-rules/_includes/_runtime.md
mpaktiti: @jhiner  https://github.com/auth0/docs/pull/2905/files#diff-3964657698ab33cf5dcc4e4bcc3bfd37R108

articles/api-auth/using-rules/password.md
mpaktiti: It isn't :) It's required for `client_credentials` and this bullet was left there after the merging of these two docs. I'll fix that right away. Thanks! 👍

mpaktiti:
Add links of these docs to this one -> https://auth0.com/docs/auth0-hooks/extensibility-points

pushpabrol:
https://auth0-docs-content-pr-2905.herokuapp.com/docs/api-auth/using-rules/password#before-you-begin

The link for install Webtask CLI has a typo hence the link is broken

mpaktiti:
Password exchange will be removed from the hooks. Closing this PR.

Issue #2847, opened 2016-11-25
Link name deprecated in PHP Web App


camimonse:
In https://auth0.com/docs/quickstart/webapp/php there is a link with a deprecated name:

Change:

**Application --> Client** here:

![image](https://cloud.githubusercontent.com/assets/18585436/20632866/ef2bd76a-b316-11e6-8dc7-d7422b8034fe.png)

Reason:

 - Better UX: We don't call "apps" or "applications" anymore. Now those are called "Client" and the user may confuse
 - Integrity of the Auth0 page: Is good to handle the same concepts and name for everything Auth0 develops

Refers to the Auth0 Ticket number: 3369

chenkie:
Merged PR

PR #2906, opened 2016-12-06
Changing "application" to "client"

camimonse:
"Client" is how Auth0 refers now in https://manage.auth0.com


camimonse:
this fixes #2847

PR #2922, opened 2016-12-08
Added use safari for web auth guide for iOS

cocojoe:
Added guide for using Safari instead of default webview for Google social authentication.   Google retired webview OAuth, more info; https://developers.googleblog.com/2016/08/modernizing-oauth-interactions-in-native-apps.html


cocojoe:
Swift Guide superseded now with Lock 2.

PR #2949, opened 2016-12-10
Fix `secret` setting in nodejs setup

alexbaumgertner:
Seems like `account.clientSecret` contains base64-encoded secret.

https://github.com/auth0/express-jwt#usage

### DIFF
--- a/snippets/server-apis/nodejs/setup.md
+++ b/snippets/server-apis/nodejs/setup.md
@@ -4,7 +4,7 @@ var app = express();
 var jwt = require('express-jwt');
 
 var jwtCheck = jwt({
-  secret: '<%= account.clientSecret %>',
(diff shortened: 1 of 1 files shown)


chenkie:
Client secrets are actually not Base64 encoded any longer which is why we took out the decoding here :) Only client create prior to Dec 7 will have encoded secrets but anything new won't.

alexbaumgertner:
Thank you for quick answer!

Perhaps I should add a note on new version of secret instead?

alexbaumgertner:
Oh, I found notice – it's already present —

client settings 2016-12-11 21-40-42

alexbaumgertner:
JFI: if you recreate secret it will notice you that your secret not base64 encoded anymore:

client settings 2016-12-18 13-36-43
client settings 2016-12-18 13-37-47

PR #2953, opened 2016-12-12
Update ID Token docs as per pipeline2

mpaktiti:
Initial PR https://github.com/auth0/docs/pull/2774


articles/metadata/lock.md
```diff
@@ -10,11 +10,11 @@ When using Auth0's [Lock](/libraries/lock) library, you can define and update th
 
 ## Defining User Metadata on Signup
 
-For information on adding `user_metadata` on signup, see the section on Lock [Custom Sign Up Fields](/libraries/lock/v10/new-features#custom-sign-up-fields) 
+For information on adding `user_metadata` on signup, see the section on Lock [Custom Sign Up Fields](/libraries/lock/v10/new-features#custom-sign-up-fields)
 
 ## Working with User Metadata
 
-Once you have [implemented the login functionality](/quickstart/spa/vanillajs#3-implement-the-login) for your Lock instance, you can choose to store the newly-created `id_token`. This token is used to retrieve the user's profile from Auth0 or to call APIs.
+Once you have [implemented the login functionality](/quickstart/spa/vanillajs#3-implement-the-login) for your Lock instance, you can choose to store the newly-created `id_token`. This token is used to retrieve the user's profile from Auth0.
```
cabralmartin: This is deprecated. Lock will show a warning when using `getProfile`.
The new method is `getUserProfile` and instead of the `id_token` it receives the  `access_token`.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,107 @@
...
+- If you used `RS256` then the token is signed with a public/private key pair, using RSA. You can verify the signature using the Public Key or Certificate, which you can find at the _[Client Settings](${manage_url}/#/clients/${account.clientId}/settings) > Show Advanced Settings > Certificates_ page.
+
+To check or update the algorithm your Client uses go to _[Client Settings](${manage_url}/#/clients/${account.clientId}/settings) > Show Advanced Settings > OAuth > JsonWebToken Signature Algorithm_. The most secure practice, and our recommendation, is to use `RS256`.
+
+## How to control the contents of an ID token
+
+In order to retrieve an `id_token` the `responseType` should be set to `id_token`, both for client-side and server-side authentication flows.
+
+::: panel-warning OAuth2 as a Service
+If you are not using *OAuth2 as a Service*, then you should use `code` for server side flows and `token` for client side flows.
```
cabralmartin: Official name is `OAuth 2.0 API Authorization`. We should make this name link to the /api-auth docs so users understand what we are referring to.

cabralmartin: On second thoughts, legacy pipeline supports `id_token`, we can remove this panel.

articles/tokens/id-token.md
```diff
@@ -0,0 +1,107 @@
...
+  ${account.namespace},
+  options
+);
+
+lock.show();
+```
+
+The `id_token` will contain only the claims specified as the value of the `scope` parameter.
+
+**NOTE:** You can also add claims to the `id_token` using [Rules](/rules), with the following format: `context.idToken['http://my-custom/claim'] = 'some-value'`.
```
cabralmartin: This is only valid for P2, a warning can be useful here.

articles/metadata/lock.md
mpaktiti: @cabralmartin I removed the last sentence. 

However, the [Reading User Metadata Properties paragraph](https://github.com/auth0/docs/pull/2953/files#diff-4782dfde2f6b51c1920a3160a58b8836R32) uses the `id_token` to invoke `lock.getProfile`. Also, the [Updating Metadata Properties paragraph](https://github.com/auth0/docs/pull/2953/files#diff-4782dfde2f6b51c1920a3160a58b8836R48) mentions this which doesn't look correct: "By including the user's `id_token` in the `Authorization` header, you can make the appropriate `PATCH` call to the [Update a user](/api/management/v2#!/Users/patch_users_by_id) endpoint."

Is this `getUserProfile` released? I can't find any reference in the docs or the [auth0.js repo](https://github.com/auth0/auth0.js).

articles/tokens/id-token.md
mpaktiti: Replaced note with panel and added link to [OAuth 2.0 API Authorization](https://auth0.com/docs/api-auth/tutorials/configuring-tenant-for-api-auth).

articles/tokens/id-token.md
mpaktiti: Removed.

Issue #2961, opened 2016-12-12
Wrong link in PHP Laravel API


camimonse:
In https://auth0.com/docs/quickstart/backend/php-laravel the following links does not redirect to where it supposed to:

![image](https://cloud.githubusercontent.com/assets/18585436/21110319/27a110b0-c073-11e6-8be8-ad9b33137548.png)

- Expected result:
that it takes you to: https://auth0.com/docs/user-profile/normalized

- Real result:
It takes you to: https://auth0.com/docs/user-profile

Auth0 Ticket: 3572

mpaktiti:
Fixed

PR #2962, opened 2016-12-12
Fixed link

camimonse:
The link was taking to: https://auth0.com/docs/user-profile which is the general user information, not specifically "Normalized"


camimonse:
this fixes #2961

PR #2997, opened 2016-12-15
Decorators in python are not annotations

danidee10:
Decorators in python are quite different from annotations (From Java's perspective). Though syntax-wise they're similar, Python decorators are just another way of passing a function as an argument to another function.

### DIFF
--- a/articles/server-apis/python/authenticate.md
+++ b/articles/server-apis/python/authenticate.md
@@ -27,15 +27,15 @@ In this example, we'll be using Flask and we'll be validating the JWT. For that,
 
 ${snippet(meta.snippets.dependencies)}
 
-### 2. Create the JWT Validation annotation
+### 2. Create the JWT Validation decorator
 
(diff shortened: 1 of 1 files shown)


danidee10:
@mpaktiti and @Annyv2 There are also several problems i noticed with the code. The major one being that it's not PEP-8 compliant. Should i make more commits to fix them, or should i keep keep this PR simple?

Annyv2 (approved):
Hi! @danidee10 Thank you very much for your contribution and your feedback. Regarding PEP-8 we will be working on that, but if you could submit another PR so we can work on it together, it would be great. Then again, thank you very much!

mpaktiti:
Thanks for the contribution! 👍

danidee10:
@Annyv2 and @mpaktiti PR submitted https://github.com/auth0-samples/auth0-python-web-app/pull/11

Annyv2:
Thanks @danidee10! We will review it later today (or tomorrow). Thanks again for your contribution!

PR #3016, opened 2016-12-15
Add warning about disabled impersonation

jstrutz:
Warn that impersonation may be disabled by default, and that they should contact support to enable it.

Part of the overall tracked here: https://github.com/auth0/crew-2/issues/247


mpaktiti:
@jstrutz Should I add this panel also at the [User Impersonation doc](https://github.com/auth0/docs/blob/master/articles/user-profile/user-impersonation.md)?

jstrutz:
@mpaktiti - If you could add that warning to the User Impersonation doc too, that would be great.

PR #3018, opened 2016-12-15
Enrollment dev

topherMarie:


woloski:
I would explain that the hosted page is used both for the regular mfa process and for the enrollment and if you would would want to check for existance of the ticket var (not ticket_id) to customize the appearance of the page, or add some specific message, etc

topherMarie:
@woloski - Addressed your comment

PR #3023, opened 2016-12-16
improved instructions on Ember 2. Fix #3022

### LINKED ISSUE #3022: Improving instruction in the Ember 2 tutorial
In https://auth0.com/docs/quickstart/spa/ember2js there is an instruction that can be improved in order to make a better understanding when following along with the tutorial.

1.
- have been --> have to be
image

     -It gives a clearer instruction so the user knows better what to do.

2.
This part should be in the System requirements, relocate the following:

image

     -This for have order in the tutorial

3.
add "you are" --> if you are using...
image

Auth0 ticket 3593

camimonse:
for better understanding. Also, it is been an instruction reordered

### DIFF
--- a/articles/client-platforms/ember2js/01-login.md
+++ b/articles/client-platforms/ember2js/01-login.md
@@ -14,7 +14,7 @@ You can get started by either downloading the seed project or if you would like 
 }) %>
 
 ::: panel-info System Requirements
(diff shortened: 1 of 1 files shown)


camimonse:
fix #3022

Issue #3069, opened 2016-12-21
Deprecated links in the React Native - Android tutorial


camimonse:
In https://auth0.com/docs/quickstart/native/react-native-android are two deprecated links:

![image](https://cloud.githubusercontent.com/assets/18585436/21402929/2c625fa8-c791-11e6-9ddb-b8428777c86d.png)

which are:
- https://github.com/auth0/react-native-lock-android#sms-passwordless
- https://github.com/auth0/react-native-lock-android#email-passwordless

Both redirect to the same repository.

Auth0-3611

PR #3070, opened 2016-12-21
Link changed

camimonse:
It was deprecated, and now the new link redirect to the correct repository: https://github.com/auth0/react-native-lock


camimonse:
fix #3069

PR #3079, opened 2016-12-22
[ instead of { in line 88 and 89

sankhakarfa:
I think this was an typo error please verify it.


mpaktiti:
Indeed, good catch! Thank you for the contribution! 👍

PR #3097, opened 2016-12-26
Update authenticate.md

fanderzon:
When copy pasting the basic lock example for the client side, and the basic express-jwt example from this document for the server side, you end up with `401: invalid token`.

This isn't a great "getting started" experience.

When base64 encoding your secret on the server side, it starts working.


chenkie:
Hey @fanderzon, thanks for this. We actually recently changed client secrets to no longer be Base64 encoded, thus the change to the docs. It's possible that you created your client before the change went into effect. Any new clients won't be Base64 encoded. 

People with existing implementations (where they are Base64 decoding their secrets) don't need to change anything, and anyone with a new implementation can forego decoding altogether. There are a few people in the middle, which sounds like your case. Since Base64 decoding won't be required going forward, I'm going to close this out, but thanks again for your PR :)

PR #3098, opened 2016-12-26
Add silent authentication doc

rolodato:

### DIFF
--- a/articles/api-auth/tutorials/silent-authentication.md
+++ b/articles/api-auth/tutorials/silent-authentication.md
@@ -2,21 +2,21 @@
 description: How to keep users logged in to your single-page application
 ---
 
-# Silent authentication for single-page apps
+# Silent Authentication for Single Page Apps
 
-Single-page web applications normally use the [Implicit Grant flow](/api-auth/grant/implicit) for authentication.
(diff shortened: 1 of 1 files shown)


articles/api-auth/tutorials/silent-authentication.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Using `prompt=none` to retrieve new tokens
+
+When authenticating a user to your application, you would normally redirect them to the [`/authorize` endpoint of Auth0's authentication API](/api/authentication#social).
+This endpoint supports a `prompt` query parameter; if it is set to `none`, the following will happen:
+
+* If the user is still authenticated, Auth0 will call back to your application immediately with renewed ID and access tokens
+* If the user is no longer authenticated or had never authenticated, Auth0 will call back to your application with an error
+
+You can use the [`silentAuthentication` method from auth0.js](https://github.com/auth0/auth0.js#silent-authentication) to authenticate users with `prompt=none`.
```
cabralmartin, line 60: We should expand here describing a little bit of what's the observed behavior of using silentAuth from auth0js, e.g. "it will create a child frame... etc".

rolodato: IMO that would just duplicate the auth0.js docs - we still need to link to them to understand how it's used and how it works, so I don't see that we would gain anything by describing the behavior here.

PR #3100, opened 2016-12-26
v1 migration guide

rolodato:


AmaanC:
There are a few instances where you mention `/userinfo` and `/tokeninfo` without linking to [the API endpoints](https://auth0.com/docs/api/authentication?http#get-user-info). Should we fix that?

rolodato:
@AmaanC Sorry, just catching up to this. Yes, I will fix that. Keep in mind that /tokeninfo will be deprecated, it's only mentioned for reference.

articles/api-auth/tutorials/adoption/scope-custom-claims.md
```diff
@@ -0,0 +1,95 @@
...
+management API to set application-specific information for this user.
+This would be the profile stored by Auth0:
+
+```json
+{
+  "email": "jane@example.com",
+  "email_verified": true,
+  "user_id": "custom|123",
+  "favorite_color": "blue",
+  "app_metadata": {
```
twistedstream: I would use `user_metadata` instead of `app_metadata` for this example since the example properties (favorite_color, preferred_contact) probably make more sense there.

articles/api-auth/tutorials/adoption/scope-custom-claims.md
```diff
@@ -0,0 +1,95 @@
...
+  "aud": "my_client_id",
+  "exp": 1311281970,
+  "iat": 1311280970,
+  "email": "jane@example.com",
+  "email_verified": true
+}
+```
+
+Note that the `user_id` property is sent as `sub` in the ID token, and that `favorite_color` and
+`app_metadata` are not present in the OIDC response from Auth0. This is
```
twistedstream: I suggest changing `app_metadata` to `user_metadata` per my previous comment

articles/api-auth/tutorials/adoption/scope-custom-claims.md
```diff
@@ -0,0 +1,95 @@
...
+because OIDC does not define standard claims to represent all the
+information in this user’s profile. We can, however, define a
+non-standard claim by namespacing it through a rule:
+
+```js
+function (user, context, callback) {
+  const namespace = 'https://myapp.example.com/';
+  context.idToken[namespace + 'favorite_color'] = user.favorite_color;
+  context.idToken[namespace + 'preferred_contact'] = user.app_metadata.preferred_contact;
```
twistedstream: I suggest changing `app_metadata` to `user_metadata` per my previous comment

articles/api-auth/tutorials/adoption/password.md
```diff
@@ -0,0 +1,192 @@
...
+}</code></pre>
+    <ul>
+        <li>The <code>device</code> parameter is only needed if <a href="/tokens/refresh-token">requesting a refresh token</a> by passing the <code>offline_access</code> scope.</li>
+    </ul>
+    </div>
+    <div id="request-oidc" class="tab-pane">
+      <pre class="text hljs"><code>POST /oauth/token HTTP 1.1
+Content-Type: application/json
+{
+  "grant_type": "http://auth0.com/oauth/grant-type/password-realm",
```
twistedstream, line 44: The `/oauth/token` endpoint also supports the standard Password Grant (`grant_type: password`) and not just our enhanced `password-realm`. A customer uses `grant_type: password` they also have to set the `tenant.default_directory` setting to the name of the connection to use for this grant type.

I think `password-realm` is a better apples-to-apples comparison with `/oauth/ro` so it's probably best to keep it as the grant type in the example, but we might want to footnote mention `password` as well.

articles/api-auth/client-types.md
```diff
@@ -0,0 +1,31 @@
...
+The OAuth 2.0 specification [defines two types of clients](https://tools.ietf.org/html/rfc6749#section-2.1): public and confidential.
+When creating a client through the management dashboard, Auth0 will ask you what type of application it represents and use that information to determine the client type.
+
+## Confidential clients
+
+Confidential clients are able to hold credentials, i.e. a client ID and secret, in a secure way without exposing it to users or attackers.
+This means that a trusted backend server is needed to store the secrets.
+The following application types are confidential clients:
+
+* Regular web applications using the [Authorization Code grant](/api-auth/grant/authorization-code)
```
twistedstream: I would also add the two active grant types that are available via the `/oauth/token` endpoint (`password` and `password-realm`) to this list. This supports the use case of a first-party regular web application that doesn't need SSO, but collects user credentials directly. It posts those credentials to its own backend endpoint and then uses either the `password` and `password-realm` grant types to _securely_ authenticate the user with Auth0 using a `client_id` and `client_secret`.

This is related to this [Slack conversation](https://auth0.slack.com/archives/p2-rollout/p1488651715029121) in `#p2_rollout`.

articles/api-auth/client-types.md
```diff
@@ -0,0 +1,31 @@
...
+Both of these grants require clients to authenticate by specifying their client ID and secret when calling the token endpoint.
+
+Since confidential clients are able to hold secrets, they can choose to have ID tokens issued to them signed symmetrically with their client secret (HS256) or asymmetrically using a private key (RS256).
+
+## Public clients
+
+Public clients are not able to hold credentials in a secure way without exposing them to users or attackers.
+The following application types are confidential clients:
+
+* Native desktop or mobile applications using the [Authorization Code grant with PKCE](/api-auth/grant/authorization-code-pkce), [Implicit grant](/api-auth/grant/implicit) or [Password grant](/api-auth/grant/password)
```
twistedstream: Per my [previous comment](https://github.com/auth0/docs/pull/3100/files#r104766198), I would _remove_ the mention of the password grant from both the **Native desktop** and **JavaScript-based client-side web applications** since we should not be promoting any authentication calls across domains: eg. credentials collected in app in DomainA (eg. `app.example.com`) being sent via XHR request to DomainB (eg. `example.auth0.com`).

articles/api-auth/tutorials/adoption/delegation.md
```diff
@@ -0,0 +1,27 @@
...
+title: Delegation and the OIDC-conformant pipeline
+---
+
+<%= include('./_about.md') %>
+
+# Delegation and the OIDC-conformant pipeline
+
+[Delegation](/api/authentication#delegation) is used for many operations, depending on your particular use case:
+
+* Exchanging an ID token issued to one client for a new one issued to a different client
```
twistedstream: We should explain how this particular use case should no longer be necessary anyway with the new pipeline. Something like:

> Typically this was done in the old pipeline because the API itself was represented as a client in Auth0. Therefore a frontend client would need to use the delegation endpoint to obtain another `id_token` that had the audience of the API in order to call the API. This became very chatty, requiring lots of calls to get the right `id_token`s. This entire use case has been properly solved with the new pipeline since APIs are now first class entities and `access_token`s can be issued to any number of clients for a single API and will only have that API's identifier as the audience.

articles/api-auth/tutorials/adoption/oidc-conformant.md
```diff
@@ -0,0 +1,56 @@
...
+
+<%= include('./_about.md') %>
+
+# OIDC-conformant clients
+
+In order to make the transition to the [OIDC-conformant authentication pipeline](/api-auth/tutorials/adoption) more predictable, clients now include an option called "OIDC Conformant", available under **Advanced Settings > OAuth**:
+
+![OIDC-conformant client setting](/media/articles/dashboard/oidc_conformant.png)
+
+For this flag to be visible, your Auth0 tenant needs to have "OAuth 2.0 API Authorization" enabled under [**Account settings > Advanced**](https://${manage_url}/#/account/advanced):
```
twistedstream: I think this flag is now enabled by default (and can't be disabled) right? If so, we can probably remove mention of this switch from this document and just list the differences.

articles/api-auth/tutorials/adoption/oidc-conformant.md
```diff
@@ -0,0 +1,56 @@
...
+    - Custom scope values can be defined by a [resource server (API)](/api-auth/tutorials/adoption/api-tokens).
+* OIDC-conformant clients cannot be the source or target client of a [delegation request](/api-auth/tutorials/adoption/delegation).
+* The `updated_at` claim is returned as a [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time) instead of an [ISO 8601 date string](https://en.wikipedia.org/wiki/ISO_8601) for consistency with the `exp`, `iat` and `nbf` claims.
+
+## I don't want to make all these changes at once!
+
+The "OIDC Conformant" flag will force all of these changes at the same time for a given client, but it's not the only option to gradually transition to the OIDC-conformant authentication pipeline.
+Any authentication requests made with an `audience` parameter will use the new pipeline, and all other requests will continue to work as usual.
+
+If your application doesn't need a resource server but you want opt-in to the new pipeline on a per-request basis, you can use the following `audience` parameter:
```
twistedstream, line 44: Another option is the client can be flagged as OIDC conformant (`client.oidc_conformant = true`), which wouldn't require sending an `audience` parameter at all.

articles/api-auth/tutorials/adoption/delegation.md
rolodato: I mentioned this below to explain why delegation was deprecated:

> Given that ID tokens should no longer be used as API tokens and that refresh tokens should be used only at the token endpoint, this endpoint is now considered deprecated.

I think it's more useful to reinforce the "ID tokens are not access tokens" concept than go into detail about legacy usage of representing APIs as clients and such, especially since delegation is a relatively advanced/complex use case. What do you think @twistedstream ?

articles/api-auth/tutorials/adoption/oidc-conformant.md
rolodato: I'm not sure I follow - the whole article describes the "OIDC Conformant" flag, and the "I don't want to make all these changes at once" section describes the use of the `audience` parameter.

twistedstream: @rolodato: Weird. I don't know what the heck I was thinking with that comment. It looks good to me.

PR #3124, opened 2016-12-29
Expand on warning about using special characters

AmaanC:


articles/extensions/delegated-admin.md
```diff
@@ -233,9 +233,9 @@ function(ctx, callback) {
 }
 ```
 
-> Note: We highly suggest not to use single or double quotes in your department or group name on which you'll want to filter since this might cause issues with the Lucene query.
+**NOTE**: Please do not use single or double quotes, or any other special characters such as `+` or `-` in your department or group name on which you'll want to filter since this might cause issues with the Lucene query, resulting in unexpected behavior.
```
mpaktiti: Hey @AmaanC, I think I'll replace that with a panel to make it more prominent since it may lead the user to errors.

PR #3159, opened 2017-01-06
Updates curl command example.

dmark:
The curl command example as cut and pasted from the online documentation (https://auth0.com/docs/connector/install-other-platforms) has a "\\" in the middle of it which results in the following error on Ubuntu and Mint:

    curl: (1) Protocol " https" not supported or disabled in libcurl

Removing the "\\" fixes the issue. There's probably no need for a line-continuation character here, and if there is, it should be at the end of the line.


mpaktiti:
Thanks for fixing that @dmark ! 👍

PR #3162, opened 2017-01-06
Re-orders steps 6 and 7.

dmark:
I believe the current order of steps 6 and 7 are incorrect. At step 6, there is no config.json file. Creating one using step 6 as a template and then running `node server.js` results in the error:

    Error: Error parsing your JSON configuration file.

I believe the steps should be reversed (this is what worked for me):

6. Run `node server.js` and enter the ticket number / URL from the Auth0 dashboard. This will create the config.json file for you.

7. Edit config.json, adding your LDAP connection and authentication details.

8. Run `node server.js` again.


mpaktiti:
Looks good, thanks for the contribution @dmark ! 👍

PR #3187, opened 2017-01-11
Add realm password grant and removed required from audience parameter

nicosabena:
* Removed `required` label from `audience` parameter on `/authorize` and `/oauth/token` calls (except when `client_credentials` grant type)
* Added info on realm password grant type


mpaktiti:
Hey @nicosabena , the realm support text was too long for the API Explorer so I shorten it to a bullet and added a link for more information. I also added the realm-related parameters at the table.

PR #3196, opened 2017-01-13
Clarify which user_id  param is needed.

thehashrocket:


mpaktiti:
Thanks for the contribution @thehashrocket ! 👍

PR #3219, opened 2017-01-16
Auth0js v8 migration guide

jeffreylees:

### DIFF
--- a/articles/libraries/auth0js/migration-guide.md
+++ b/articles/libraries/auth0js/migration-guide.md
@@ -25,6 +25,28 @@ Initialization of auth0.js in your application will now use `auth0.WebAuth` inst
 </script>
 ```
 
+## The parseHash Method
+
+The `parseHash` method now validates the id_token to avoid CRFS (along with the nonce) attacks. In order for this to work properly, the token should be signed using RS256, and will fail if the token is signed using HS256.
+
+This can be avoided by either switching how your id_tokens are signed, or by manually parsing hashes, rather than using the `parseHash` method.
+
+### Switching from HS256 to RS256
+
+To switch from HS256 to RS256 for a specific client, follow these instructions:
+1. Go to https://manage.auth0.com/#/clients
+1. Select your client
+1. Go to Settings
+1. Click on Show Advanced Settings
+1. Click on the OAuth tab in Advanced Settings
+1. Change the JsonWebToken Signature Algorithm to RS256
+
+And remember that if the token is being validated anywhere else, that those places might need to be altered as well to comply.
+
+### Manually Parsing Hashes 
+
+If you would rather manually parse hashes, to avoid the `parseHash` method since it only works with RS256, feel free to take a look at [this example](https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js#L97) or [this one](https://github.com/auth0/auth0.js/blob/master/src/helper/qs.js#L10) to help you get started.
+
 ## Login
 
 The `login` method of version 7 was divided into several different methods in version 8, based on the type of auth you need, rather than the old `login` method.
@@ -115,7 +137,6 @@ webAuth.client.login({
 ## Refreshing Tokens
 
 Refreshing tokens is now done via the [renewAuth](/libraries/auth0js#using-renewauth-to-acquire-new-tokens) method. 
-
 
 ## User Management
 
--- a/articles/libraries/auth0js/migration-guide.md
+++ b/articles/libraries/auth0js/migration-guide.md
@@ -16,7 +16,7 @@ Take a look below for more information about changes and additions to auth0.js i
 
 Initialization of auth0.js in your application will now use `auth0.WebAuth` instead of `Auth0`
 ```html
-<script src="https://cdn.auth0.com/js/auth0/8.0.4/auth0.min.js"></script>
+<script src="https://cdn.auth0.com/js/auth0/8.1.0/auth0.min.js"></script>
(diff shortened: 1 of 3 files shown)


articles/libraries/auth0js/migration-guide.md
```diff
@@ -0,0 +1,62 @@
...
+description: How to migrate from auth0.js v7 to auth0.js v8
+url: /libraries/auth0js/migration-guide
+---
+
+# Auth0.js v7 to v8 Migration Guide
+
+The following instructions assume you are migrating from **auth0.js v7** to **auth0.js v8**. 
+
+The goal of this migration guide is to provide you with all of the information you would need to update Auth0.js in your application. Of course, your first step is to include the latest version of auth0.js. Beyond that, take a careful look at each of the areas on this page. You will need to change your implementation of auth0.js to reflect the new changes. Take a look below for more information!
+
```
hzalaz, line 14: I'd add a note saying we recommend auth0.js v8 if you have the OAuth2 flag enabled in your tenant

articles/libraries/auth0js/migration-guide.md
```diff
@@ -0,0 +1,62 @@
...
+<script src="https://cdn.auth0.com/js/auth0/8.0.4/auth0.min.js"></script>
+<script type="text/javascript">
+  var webAuth = new auth0.WebAuth({
+    domain:       'YOUR_AUTH0_DOMAIN',
+    clientID:     'YOUR_CLIENT_ID'
+  });
+</script>
+```
+
+## Login
```
hzalaz, line 50: I'd say that the `login` method of v7 was divided in different methods in v8 based on the type of auth you need:

- redirect: hosted login page 
- redirect: social & enterprise login
- popup: hosted login page 
- popup: social  & enterprise login
- non redirect auth: database connection using /oauth/token

hzalaz: In case you wan't to explain why is splitted the login (may need to work on the wording)
Also the rationale about splitting this is because we wanted to make explicit the type of flow used (redirect vs popup) in the code instead of auth0.js assuming what to do based on the parameters received. Also it's easier to know what auth0.js will do since it will just do what is told without making any assumptions.

hzalaz (changes requested):

articles/libraries/auth0js/migration-guide.md
jeffreylees: Think the actual reference doc should explain all of these scenarios too, probably? Right now it only shows a couple of them. I can walk through the different scenarios and show them without much trouble I think. Might add that into this PR and have you review.

hzalaz: @jeffreylees ok but we should explain it in the migration guide since we divided the bloated login method of v7 and people will ask or wonder how to do it. We can leave a small snippet here with basic usage and a link to the index for a proper detailed doc. We can even share the snippet if it works but I'd rather add here a small one without audience or p2 feature (actually I'd refrain of using audience or any other p2 related attribute or option)

articles/libraries/auth0js/migration-guide.md
```diff
@@ -0,0 +1,112 @@
...
+
+### webAuth.authorize()
+
+The `authorize` method can be used for logging in users via the [Hosted Login Page](/libraries/auth0js#hosted-login-page), or via social connections, as exhibited below. The default behavior for `authorize` is redirect, but popup can be specified.
+
+For hosted login, one must call the authorize endpoint
+
+
+```js
+webAuth.authorize({ //Any additional options can go here });
```
hzalaz:
```
webAuth.authorize({
  // Any additional options can go here
});
```

`//` comments all that follows

articles/libraries/auth0js/migration-guide.md
```diff
@@ -0,0 +1,112 @@
...
+</script>
+```
+
+## Login
+
+The `login` method of version 7 was divided into several different methods in version 8, based on the type of auth you need, rather than the old `login` method.
+
+### webAuth.authorize()
+
+The `authorize` method can be used for logging in users via the [Hosted Login Page](/libraries/auth0js#hosted-login-page), or via social connections, as exhibited below. The default behavior for `authorize` is redirect, but popup can be specified.
```
hzalaz: `.authorize()` just redirects and you can't tell it to use a popup. There is another method inside the `popup` "namespace" to perform popup auth like you explained later. What I mean is they are different methods instead of a single one

articles/libraries/auth0js/migration-guide.md
```diff
@@ -0,0 +1,112 @@
...
+
+The goal of this migration guide is to provide you with all of the information you would need to update Auth0.js in your application. Of course, your first step is to include the latest version of auth0.js. Beyond that, take a careful look at each of the areas on this page. You will need to change your implementation of auth0.js to reflect the new changes. Note that we especially recommend migrating to auth0.js v8 if you have the OAuth2 flag enabled in your tenant. 
+
+Take a look below for more information about changes and additions to auth0.js in version 8!
+
+## Initialization of auth0.js
+
+Initialization of auth0.js in your application will now use `auth0.WebAuth` instead of `Auth0`
+```html
+<script src="https://cdn.auth0.com/js/auth0/8.0.4/auth0.min.js"></script>
```
hzalaz: Latest is `8.1.0`

hzalaz (changes requested):
Besides that I'd add passwordless to this migration guide since we just fixed the issues we had but we can do it in another PR

PR #3228, opened 2017-01-16
Fixed callback_path in Rails OmniAuth middleware

pangolingo:
I updated the path to match the `/auth/oauth2/callback` paths used in the rest of the Rails docs. When using the old path (`/auth/auth0/callback`), OmniAuth does not work and does not include `omniauth.auth` information in the request.

Apologies: I did not submit this to the Heroku review.


chenkie:
Thanks for this!

PR #3230, opened 2017-01-17
Remove deprecated auth0.js method

mpaktiti:


mpaktiti:
@jeffreylees Can you check if this PR is valid?

articles/connections/social/devkeys.md
```diff
@@ -28,6 +28,6 @@ When using the Auth0 developer keys, there are a few caveats you need to be awar
 
 4. [Federated Logout](/logout#log-out-a-user) does not work. When using the Auth0 developer keys, calling `/v2/logout?federated` will sign the user out of Auth0, but not out of the Social Identity Provider.
 
-5. `prompt=none` won't work on the [/authorize](/api/authentication/reference#social) endpoint. [Auth0.js' `silentAuthentication` method](https://github.com/auth0/auth0.js#silent-authentication) uses `prompt=none` internally, so that won't work either.
```
mpaktiti, line 31: Looks like `renewAuth` is using `prompt=none` internally so I should probably leave the sentence in the doc and update the method name and link. Code ref: https://github.com/auth0/auth0.js/blob/b9ffbb28c8acb6d126d51d87ed0156fd482b7f3c/src/web-auth/index.js#L211

jeffreylees: Yeah, I'd either change the method name to `renewAuth` and link to the GitHub - or to the docs page here https://auth0.com/docs/libraries/auth0js#using-renewauth-to-acquire-new-tokens (I'll make a card to strengthen the renewAuth section now that there's more info on it), or you could link to both the v7 and v8 methods. I'd probably stick with the v8 method, myself, but both are supported so that's an option. Definitely don't just leave it with the v7 one though.

mpaktiti: Thanks Jeff!

PR #3245, opened 2017-01-18
Appliance Instrumentation

khorne3:


mpaktiti:
@khorne3 Looks good 👍  I expected to find the Instrumentation main article, listed at the main Appliance index, but I assume it was intended.

PR #3256, opened 2017-01-19
Appliance RACI Table

khorne3:

### DIFF
--- a/articles/appliance/index.html
+++ b/articles/appliance/index.html
@@ -48,6 +48,12 @@ title: Auth0 Appliance
     </p>
   </li>
   <li>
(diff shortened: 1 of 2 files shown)


mpaktiti:
@khorne3 I noticed that `Appliance Deployment Modules` and `Appliance Personnel Roles and Responsibilities` are missing from the left hand menu (under `Appliance Planning`). Maybe this is intended, just checking.

khorne3:
@mpaktiti I'm working with the Appliance team on the nav bar. I didn't add them, and maybe they should be there, but I'd like their feedback on what they think are the most important links. I'm not sure we should have everything there.

PR #3273, opened 2017-01-20
Adding Passwordless + other items to auth0js v8 + migration guide

jeffreylees:

### DIFF
--- a/articles/libraries/auth0js/index.md
+++ b/articles/libraries/auth0js/index.md
@@ -9,9 +9,9 @@ url: /libraries/auth0js
 
 Auth0.js is a client-side library for [Auth0](http://auth0.com). Using auth0.js in your web apps makes it easier to do authentication and authorization with Auth0 in your web apps.
 
-::: panel-info Auth0.js v8
-This document covers the most up-to-date version of auth0.js - version 8. If you are already using version 7, you can take a look at the [v7 reference guide](/libraries/auth0js/v7), or take a look at the [v8 migration guide](/libraries/auth0js/migration-guide).
-:::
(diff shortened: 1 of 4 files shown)


articles/libraries/auth0js/index.md
```diff
@@ -85,45 +85,148 @@ Initialize a new instance of the Auth0 client as follows:
...
+
 ## Logout
 
 To logout a user, use the `logout` method. This accepts an options object, which can include a 'client_id', and a 'returnTo' URL. If you want to navigate the user to a specific URL after the logout, set that URL at the 'returnTo' parameter.
 
-::: panel-info returnTo parameter
-Note that if the 'client\_id' parameter _is_ included, the 'returnTo' URL that is provided must be listed in the **client's** "Allowed Logout URLs" in the Auth0 dashboard. 
-However, if the 'client\_id' parameter _is not_ included, the 'returnTo' URL must be listed in the "Allowed Logout URLs" at the **account level** in the Auth0 Dashboard.
-:::
+<div class="alert alert-info">
```
mpaktiti: Hey @jeffreylees , the idea for the alerts was to be used for short versioning information (ie This doc uses v8, for v7 refer to...), not replace panels altogether, so I would change that (and the other two in the doc) . If you think we should replace them altogether maybe we need to consult with the design team beforehand to get their opinion.

jeffreylees: Oh, whoops, misunderstood. Will change them back.

mpaktiti (changes requested):
Replace alerts with panels (or follow up with design team)

PR #3287, opened 2017-01-22
authResult contains idToken

ryantology:
Was just reading the docs and I really hope the `authResult` contains `idToken`. If it does not I am really confused on how to get the JWT using Lock :)


jeffreylees (approved):
Yeah, looks good, when implementing the `accessToken` changes with 10.9 for accessing profile info, etc, changed too many references :D

Thanks @ryantology

PR #3322, opened 2017-01-24
Change link from general Auth0 repo to Docs libraries page

khorne3:

### DIFF
--- a/articles/integrations/aws.md
+++ b/articles/integrations/aws.md
@@ -272,9 +272,7 @@ The result of calling the delegation endpoint will contain the AWS token in the 
 ```
 
 :::panel-info Auth0 Libraries
-The [Auth0 client libraries](/libraries) simplify the process of calling these endpoints. See an example for client-side JavaScript at: [Delegation Token Request](/libraries/auth0js/v7#delegation-token-request).
-
-Please note that this example is for **version 7** of the `auth0js` library; delegation is *not* supported in version 8 of `auth0js`.
+The [Auth0 client libraries](/libraries) simplify the process of calling these endpoints. See an example for client-side JavaScript at: [Delegation Token Request](/libraries/auth0js/v7#delegation-token-request). Please note that this example is for **version 7** of the `auth0js` library; delegation is *not* supported in version 8 of `auth0js`.
 :::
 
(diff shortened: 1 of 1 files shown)


articles/integrations/aws.md
```diff
@@ -271,7 +271,7 @@ The result of calling the delegation endpoint will contain the AWS token in the
 }
 ```
 
-**NOTE:** The Auth0 client libraries simplify calling these endpoints. See our [GitHub repo](https://github.com/auth0/) for the latest SDKs. See an example for client-side JavaScript at: [Delegation Token Request](https://github.com/auth0/auth0.js#delegation-token-request).
+**NOTE:** The [Auth0 client libraries](/libraries) simplify the process of calling these endpoints. See an example for client-side JavaScript at: [Delegation Token Request](https://github.com/auth0/auth0.js#delegation-token-request).
```
mpaktiti: @khorne3 The _Delegation Token Request_ does not exist.

@jeffreylees Can you help out on what Katie should use here?

mpaktiti (changes requested):
Fix Delegation Token Request link.

articles/integrations/aws.md
khorne3: @mpaktiti Actually, I just remembered we're making big changes to delegation. Should we mark this as blocked and waiting?

jeffreylees: In auth0.js v8 the getDelegationToken method no longer exists. There is a `delegation` method, but it's deprecated and so wasn't to be documented.

https://auth0.com/docs/libraries/auth0js/v7#delegation-token-request describes what you want, but if you're going to link to that, you should make some kind of note about how that's auth0.js v7, and how delegation is a deprecated concept in v8, I guess?

I think if it were me I'd just remove that entire block of text, for now, and the sample below it exhibiting it, maybe.

We could also ask someone in Brucke for clarification if needed as to delegation and v8

mpaktiti: I would say just replace the link for now. We still haven't removed delegation from our docs and I'm not clear on when this will be done. Just add a link to v7 as Jeff says and mention that it is not supported in the latest version (8).

khorne3:
@mpaktiti Updated.

PR #3341, opened 2017-01-25
Update anomaly-detection/breach password

joseluisdiaz:
- New screenshots for anomaly detection.
- Describe new actions.


dafortune:
Have the screenshots the mac "chrome"? (seems that they are missing the close, minimize, etc. buttons) The rest looks good to me assuming that that phrase explains the change.

joseluisdiaz:
I saw a few screenshot using Windows Chrome. I assumed that was ok.

nikolaseu:
The first one does not include the change that was merged yesterday in the breached password detection box (and is "gray"/disabled).
Anyway I guess it would look better if the screenshot is the same style as before (with the shadow around the window, which macOS adds automatically)

PR #3342, opened 2017-01-26
Troubleshooting guides for ASP.NET Web API (Core & OWIN)

jerriep:


jerriep:
We need to wait with merging this, as this is based on using API Authentication (i.e. using the `access_token` as opposed to using the `id_token`), and those steps have been hidden for now based on feedback from DSE (see #3362)

Once we re-enable those steps, then we can re-base and merge this PR

jerriep:
This was merged long time ago as part of another PR

PR #3344, opened 2017-01-26
Updated link of http calls to Auth0 API

johnnyasantoss:
Fixed urls from angular1.x/linking-accounts docs


mpaktiti:
Good catch @johnnyasantoss, thanks ! 👍

PR #3356, opened 2017-01-26
Custom Social Connections: Add section on callback url

angelaLazar:
Also provide links to additional provider docs

https://trello.com/c/6sL7HfdJ/554-auth0-extension-custom-social-connecti
ons


articles/extensions/custom-social-extensions.md
```diff
@@ -24,7 +25,7 @@ You will be asked to authorize the Custom Social Connections app. After you do s
```
mpaktiti, line 25: This image needs to be updated, we have more social providers:

image

articles/extensions/custom-social-extensions.md
```diff
@@ -61,6 +62,15 @@ Using the slider, enable this social connection for the apps that you want to us
 
 Once you have enabled/disabled the appropriate apps, click **Save**.
```
mpaktiti, line 64: The customer also commented on this: `It's lacking information on what programming language to use in the fetch userprofile function, what methods are available to that function.
Furthermore it lacks info on how to map fields from the 3rd party API to the ones of Auth0.,`

The language is JS (not sure if we can use another). You could clarify this under paragraph `#### New Connection: Settings`. I don't know how the map fields works but sandrino would know.

articles/extensions/custom-social-extensions.md
```diff
@@ -92,3 +102,19 @@ If the identity provider requires Basic Authentication, you can be use the __Cus
     "Authorization" : "Basic MTIzNDU2OmFiY2RlZg=="
 }
 ```
+
+## Additional Provider Resources
```
mpaktiti: Not sure about this paragraph. It can be helpful to users for sure but it looks like content they can easily find, and if we leave it here we will have to maintain it. For example, when you update the image with the latest, you will have to add links here as well.

mpaktiti (changes requested):
- Update social providers screenshot
- Add more details on Fetch User Profile Script and mapping fields

articles/extensions/custom-social-extensions.md
angelaLazar: done

articles/extensions/custom-social-extensions.md
angelaLazar: Updated to give more details

articles/extensions/custom-social-extensions.md
angelaLazar: You're right, I removed this and added a sentence about searching for the provider's docs.

angelaLazar:
@mpaktiti Addressed your comments in latest commit

mpaktiti (approved):

PR #3365, opened 2017-01-27
Fix parseHash: not it uses a callback

glena:
Closes https://github.com/auth0/auth0-docs/issues/1305


fiddur:
That was quick :)

Seems I put the issue in the wrong place: https://github.com/auth0/auth0-docs/issues/1305

glena:
@fiddur auth0/auth0-docs is the docs app, auth0/docs is the content :)

PR #3375, opened 2017-01-27
Detail XSRF attack

angelaLazar:
https://trello.com/c/a7kZX5Vg/701-using-the-state-parameter


articles/protocols/oauth2/oauth-state.md
```diff
@@ -2,15 +2,17 @@
...
 ---
 
-# Using the State Parameter
+# The State Parameter
 
-The `state` parameter is one of the supported Auth0 [Authentication Parameters](/libraries/lock/v10/sending-authentication-parameters). This page is to help you understand how to utilize this this parameter which is useful to help mitigate [XSRF attacks](https://en.wikipedia.org/wiki/Cross-site_request_forgery) and for providing any contextual information (such as a return url) that you might need after the authentication process is finished.
+The `state` parameter is one of the supported Auth0 [Authentication Parameters](/libraries/lock/v10/sending-authentication-parameters), used to help mitigate [XSRF attacks](https://en.wikipedia.org/wiki/Cross-site_request_forgery).
+
+A XSRF attack can occur when a malicious program causes a user's web browser to perform an unwanted action on a trusted site that the user is currently authenticated. This type of attack specifically target state-changing requests to initiate a type of action instead of getting user data because the attacker has no way to see the response of the forged request.
```
mpaktiti, line 11: I think this article needs some more work. The purpose of the card is to outline the attach scenario but we only give a definition and info on how to use the param. I think that we need a sub-paragraph describing a simple attack scenario, accompanied by a diagram.

mpaktiti (changes requested):
Add an attach scenario to the article

articles/protocols/oauth2/oauth-state.md
angelaLazar: @mpaktiti A diagram is a good idea. Do you know what program can I use to create one?

mpaktiti: Just draw it with whatever suits you best and send it to Julieta from design team. She will make it pretty and send it back to you. Personally sometimes I just draw it on paper and take a photo. Or you could use something like this https://www.draw.io/

angelaLazar:
@mpaktiti I have added in the diagram from design

PR #3429, opened 2017-02-01
Fix documentation for passwordless auth0js calls

lnhrdt:
connection string must be email or sms


mpaktiti:
Good catch, thanks @lnhrdt ! 👍

PR #3441, opened 2017-02-02
New doc on how to verify a token in the API

mpaktiti:


articles/api-auth/tutorials/verify-access-token.md
```diff
@@ -0,0 +1,116 @@
...
+
+- _Token issuer_: The `iss` claim denotes the issuer of the JWT. The value _must_ match the one configured in your API. For JWTs issued by Auth0, `iss` holds your Auth0 domain with a `https://` prefix and a `/` suffix: `https://${account.namespace}/`.
+
+- _Token audience_: The `aud` claim identifies the recipients that the JWT is intended for. For JWTs issued by Auth0, `aud` holds the unique identifier of the target API (field __Identifier__ at your [API's Settings](${manage_url}/#/apis)). If the API is not the intended audience of the JWT, it _must_ reject the request.
+
+### How can my API validate the claims?
+
+To validate the claims, you have to decode the JWT, retrieve the claims (`exp`, `iss`, `aud`) and validate their values.
+
+The easiest way however, is to use one of the libraries listed in the _Libraries for Token Signing/Verification_ section of [JWT.io](https://jwt.io/). Note that not all libraries validate all the claims. In [JWT.io](https://jwt.io/) you can see which validations each library supports (look for the green check marks).
```
jerriep: @mpaktiti It is good to mention these libraries, but note that they are typically used **only** to validate the token. 

It may also be worth mentioning that many web frameworks (such as ASP.NET Core for example) has JWT middleware which handle the token validation as well. Most of the times this will be a better route to take, rather that resorting to using one of these libraries, as the middleware typically integrates well with the framework's overall authentication mechanisms.

articles/api-auth/tutorials/verify-access-token.md
```diff
@@ -0,0 +1,135 @@
...
+Alternatively, you can use one of the libraries listed in the _Libraries for Token Signing/Verification_ section of [JWT.io](https://jwt.io/).
+
+Following the Node.js example of the previous section, the [jwt.verify()](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) method of the [node-jsonwebtoken library](https://github.com/auth0/node-jsonwebtoken), supports an `algorithms` argument, that contains a list of strings with the names of the allowed algorithms.
+
+## Verify the Signature
+
+The API needs to verify the signature. The signature is used to verify that the sender of the JWT is who it says it is and to ensure that the message wasn't changed along the way.
+
+Remember that the signature is created using the header and the payload of the JWT, a secret and the hashing algorithm being used (as specified in the header: HMAC, SHA256 or RSA). The way to verify it, depends on the hashing algorithm:
+- For `HS256`, the Client's Secret is used. You can find this information at your [Client's Settings](https://${manage_url}/#/clients/${account.clientId}/settings).
```
cabralmartin: The API secret is used, not the client's.

image

articles/api-auth/tutorials/verify-access-token.md
```diff
@@ -0,0 +1,135 @@
...
+### How can I validate the claims?
+
+To validate the claims, you have to decode the JWT, retrieve the claims (`exp`, `iss`, `aud`) and validate their values.
+
+The easiest way however, is to use one of the libraries listed in the _Libraries for Token Signing/Verification_ section of [JWT.io](https://jwt.io/). Note that not all libraries validate all the claims. In [JWT.io](https://jwt.io/) you can see which validations each library supports (look for the green check marks).
+
+Following the Node.js example, the [jwt.verify()](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) method of the [node-jsonwebtoken library](https://github.com/auth0/node-jsonwebtoken), validates these claims, depending on the input arguments:
+- `audience`: if you want to check `aud`, provide a value here
+- `issuer`: string or array of strings of valid values for the `iss` field
+- `ignoreExpiration`: if `true` do not validate the expiration of the token
```
cabralmartin: We are advising for best practice, there are not options here.

`audience` has to be set to the API identifier
`ignoreExpiration` *must* be set to `false`.

PR #3444, opened 2017-02-02
Deleted integratedWindowsLogin doc

nicosabena:
The option is not currently available for Lock v10. Deleted to avoid confusions.


mpaktiti:
cc @jeffreylees

PR #3457, opened 2017-02-03
Fix get user info api link

wong2:
since `getProfile` is deprecated, should link to `getUserInfo` here


mpaktiti:
Thanks @wong2 ! 👍

PR #3466, opened 2017-02-03
Lock iOS v2 Library Docs

jeffreylees:


cocojoe:
Do you think it's worth breaking out the sections a bit more like the original? 
https://auth0.com/docs/libraries/lock-ios

If you can update versions for libs as final now, so 2.0 for Lock, 1.2 for Auth0.swift
Also when you write `iOS/macOS` can you use `iOS and macOS` as that's how it was previously used and also consistent with page title.

cocojoe:
Can you set the carthage Lock dependency to

github "auth0/Lock.iOS-OSX" ~> 2.0

Also can you setup a v2 version of this

https://auth0-docs-content-pr-3466.herokuapp.com/docs/libraries/lock-ios/v1/delegation-api

I'm just doing a sample snippet for v2 equivalent.

cocojoe:
```swift
Auth0.authentication()
            .delegation(withParameters: ["id_token" : "<AUTH0 ID TOKEN>", "api_type": "firebase"])
            .start() { result in
                switch result {
                case .success(let payload):
                    // payload will have your firebase token
                case .failure(let error):
                    // Handle Error
                }
            }
```

PR #3484, opened 2017-02-07
Fix links

mpaktiti:


mpaktiti:
Fixed by https://github.com/auth0/docs/pull/3479

PR #3485, opened 2017-02-07
Replace getProfile with getUserInfo

mpaktiti:
[Review app](https://auth0-docs-content-pr-3485.herokuapp.com/docs/metadata/lock)

### DIFF
--- a/articles/metadata/lock.md
+++ b/articles/metadata/lock.md
@@ -31,18 +31,14 @@ if (hash) {
 
 ## Reading User Metadata Properties
 
-You can read the user's `user_metadata` properties the same way you would for any user profile property. This example retrieves the value associated with `user.metadata.hobby`:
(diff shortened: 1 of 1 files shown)


mpaktiti:
@jeffreylees This is about this card https://trello.com/c/ukxpEHrx/837-using-metadata-with-auth0-s-lock-library Except for your comment, I made some additional changes since there were more references to the `id_token`. Can you have a look?

jeffreylees (approved):
This looks fine to me @mpaktiti. I might change the element ID from 'name' to 'hobby' just to match what we're grabbing, but that's not really a necessary thing.

The big thing is getUserInfo with the accessToken, which you have here. I assume there's no need to show any more of a Lock demo than that bit, here, especially with the note/link about initializing Lock. That link could also point to the docs page if you want it to.

PR #3487, opened 2017-02-07
Add content on state param

mpaktiti:


articles/rules/redirect.md
```diff
@@ -32,18 +31,22 @@ function (user, context, callback) {
...
+## What to do afterwards
 
 An authentication transaction that has been interrupted by setting `context.redirect` can be resumed by redirecting the user to the following URL:
 
 ```text
 https://${account.namespace}/continue?state=THE_ORIGINAL_STATE
 ```
 
+::: panel-info State
+State is an opaque value the clients adds to the initial request that Auth0 includes when redirecting back to the client. We highly recommend using this param, since it can be used by the client to prevent [CSRF attacks](/security/common-threats#cross-site-request-forgery-xsrf-or-csrf-). By `THE_ORIGINAL_STATE` we mean the value you provided during user authentication.
```
AmaanC: That's not right. In this context, `THE_ORIGINAL_STATE` is not the state the client passes to Auth0 in the original request. `THE_ORIGINAL_STATE` is the state that Auth0 generates and sends to the redirect URL.

If your Rule has this configured:
```
    context.redirect = {
        url: "https://redirect.com/foo"
    };
}
```

For example.

1. `client.com/login`
2. `<account>.auth0.com/authorize?...&state=foo123`
3. Social IdP / DB connection call
4. `redirect.com/foo?state=bar42` (note that this is entirely independent of the state the client passes to Auth0 - this state is generated for the transaction between Auth0 and the redirect URL)
5. `<account>.auth0.com/continue?state=bar42`
6. Rules run again with `context.protocol === "redirect-callback"`
7. `client.com/callback?code=...&state=foo123`

In this context, `redirect.com/foo` needs to handle the `state` parameter `bar42` and send it back to Auth0's `/continue`.

articles/rules/redirect.md
```diff
@@ -32,18 +31,22 @@ function (user, context, callback) {
...
+:::
+
 When a user has been redirected to the `/continue` endpoint, all rules will be run again.
 
-::: panel-danger Caution:
-Make sure to send back the original state to the `/continue` endpoint, otherwise Auth0 will lose the context of the login transaction.
+::: panel-danger Caution
+Make sure to send back the original state to the `/continue` endpoint, otherwise Auth0 will loose the context of the login transaction.
```
AmaanC: `lose` is correct in this situation. Also, I think it might be nice to elaborate on what "losing the context of login transaction" means. Something like:

```
Make sure to send back the original state to the `/continue` endpoint, otherwise Auth0 will lose the context of the login transaction and the user will not be able to login due to to an `invalid_request`.
```

articles/rules/redirect.md
```diff
@@ -29,27 +29,32 @@ function (user, context, callback) {
...
 ```text
 https://${account.namespace}/continue?state=THE_ORIGINAL_STATE
 ```
 
-::: panel-info State
-State is an opaque value the clients adds to the initial request that Auth0 includes when redirecting back to the client. We highly recommend using this param, since it can be used by the client to prevent [CSRF attacks](/security/common-threats#cross-site-request-forgery-xsrf-or-csrf-). By `THE_ORIGINAL_STATE` we mean the value you provided during user authentication.
-:::
+By `THE_ORIGINAL_STATE` we mean the value that Auth0 generated and sent to the redirect URL. For example, if your rule redirected to `https://example.com/foo`, Auth0 would use a redirect URL similar to: `https://example.com/foo?state=abc123` (`abc123` being the `THE_ORIGINAL_STATE`). In this case in order to resume the authentication transaction you should redirect to `https://${account.namespace}/continue?state=abc123`.
+
+To extract the state value, check the `request.query` object for Database logins (`context.request.query.state`), or the `request.body` object for Social/Enterprise logins (`context.request.body.state`).
```
AmaanC: _Technically_ that's right, but it's not relevant. (It's only correct for when `ctx.protocol === 'redirect-callback'`; extracting the state at this point is useless as far as I can tell, because you've already resumed the transaction by sending it to `/continue`).

The way to extract the state is simply to have the route on your redirect URL to save the state and send it back to `/continue` - this will vary based on frameworks used and languages. 

In [Express](http://expressjs.com/en/api.html#req.query), for example on `redirect.com`:

```
app.get('/foo', function(req, res) {
  // Some magic here
  res.redirect('https://<account>.auth0.com/continue ' + req.query.state);
});
```

articles/rules/redirect.md
```diff
@@ -29,27 +29,32 @@ function (user, context, callback) {
...
 ```
 
-Once all rules have finished executing, the user will be redirected to the specified URL.
+Once all rules have finished executing, the user will be redirected to the specified URL. 
+
+Auth0 will also pass a state value in that URL, for example `https://example.com/foo?state=abc123`.
+
+<div class="alert alert-info"><strong>What is state?</strong> State is an opaque value the clients adds to the initial request that Auth0 includes when redirecting back to the client. We highly recommend using this param, since it can be used by the client to prevent <a href="/security/common-threats#cross-site-request-forgery-xsrf-or-csrf-">CSRF attacks</a>.</div>
+
+You need to extract this value since you will need it in order to resume the authentication transaction after the redirect (see `THE_ORIGINAL_STATE` at the next paragraph).
```
AmaanC: I think it might be better to take this line out entirely. It's misleading because it follows this sentence:

> State is an opaque value the clients adds to the initial request that Auth0 includes

See below for more.

articles/rules/redirect.md
```diff
@@ -29,27 +29,32 @@ function (user, context, callback) {
 }
 ```
 
-Once all rules have finished executing, the user will be redirected to the specified URL.
+Once all rules have finished executing, the user will be redirected to the specified URL. 
+
+Auth0 will also pass a state value in that URL, for example `https://example.com/foo?state=abc123`.
+
+<div class="alert alert-info"><strong>What is state?</strong> State is an opaque value the clients adds to the initial request that Auth0 includes when redirecting back to the client. We highly recommend using this param, since it can be used by the client to prevent <a href="/security/common-threats#cross-site-request-forgery-xsrf-or-csrf-">CSRF attacks</a>.</div>
```
AmaanC: True in general, but in this case `state` is an opaque value that Auth0 passes to the redirect URL (which may be entirely independent of the client) - could you fix the wording?

If you look at my earlier review's explanation, this paragraph talks about `state` as being `foo123`, where we're actually documenting `bar42` (in that example). `bar42` was the state between Auth0 and the redirect URL domain - so it isn't including in the initial request to Auth0, nor is it sent back to the client.

It is correct that both states are there to prevent CSRF, though, so we can leave that part in.

AmaanC (changes requested):

articles/rules/redirect.md
mpaktiti: Sorry but I can't seem to grasp this. I thought I did but from what I understand I was wrong. You say `extracting the state at this point is useless because you've already resumed the transaction by sending it to /continue` but the way I see it you need to extract this value in order to send it to continue:
- Auth0 redirects to `https://example.com/foo?state=abc123`
- The client extracts the `abc123` value
- Client resumes the transaction by redirecting to `https://DOMAIN/continue?state=abc123`

Not sure what's wrong with how I describe this. Could you perhaps edit this PR?

mpaktiti: Basically my bullets above describe exactly what I understand from your express script :( I'm really confused as to what is wrong with the wording at this point.

AmaanC: I think you do understand it based on what you're saying. The flaw is in this bit:

```
To extract the state value, check the `request.query` object for Database logins (`context.request.query.state`), or the `request.body` object for Social/Enterprise logins (`context.request.body.state`).
```

That implies that the `state` is being extracted in the Rule itself (either before the redirect or _after_ it's been continued).
As you said, the _client_ (or rather the domain we redirect to, to be accurate) is supposed to extract `state` and send it back to `/continue`.

I'll look into editing the wording tomorrow (or Monday) and maybe we can sync then

mpaktiti:
Thanks for the help @AmaanC ! 👍

PR #3492, opened 2017-02-08
Update github-deploy.md

cdnsteve:
The existing stage throws an error and didn't match latest stage documentation in: https://auth0.com/docs/api/management/v2#!/Rules/post_rules


mpaktiti:
Thanks @cdnsteve ! 👍

PR #3505, opened 2017-02-11
I think you mixed up access and id tokens

SteveALee:
But I'm not 100% sure. Usage seems to vary, but I assume this is for OpenID which is quite strict.

### DIFF
--- a/articles/api-auth/grant/implicit.md
+++ b/articles/api-auth/grant/implicit.md
@@ -14,9 +14,9 @@ The **Implicit Grant** (defined in [RFC 6749, section 4.1](https://tools.ietf.or
 
  1. The Client initiates the flow and redirects the user to the Authorization Server
  2. The user authenticates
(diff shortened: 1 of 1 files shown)


mpaktiti:
Hello @SteveALee ! Thanks for taking the time to make this update. However we intended to use `access_token`. This article is about getting a token to call an API and in that case an `access_token` should be used, since the `id_token` is meant for the client only.

SteveALee:
@mpaktit Yes, I got myself mixed up when writing. 

To be honest I found varied use for both tokens across docs and blog posts and mixed statements about which were optional, this  got my head in a spin. Reading the OpenID spec I see Access token is indeed for resource access and id_token identifies the user.

I won't create another issue for the varied use as many are historical. It is confusing as you read around though. Your blog on Refresh tokens explains part of the cause - your getting OpenID cert (congrats BTW) and their strict usage was a change.

thanks anyway.

PR #3506, opened 2017-02-12
Update 01-login.md

michalkosmowski:
Added step about `MainApplication.java`. Fixes: http://stackoverflow.com/questions/40550099/lockmodule-init-undefined-error-in-auth0-react-android-getting-started


chenkie:
@hzalaz can you review/approve? We'd need to resolve conflicts if this is good to go.

Thanks

PR #3509, opened 2017-02-13
Rewrite doc on Mgmt API v2 Token

mpaktiti:


articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
 
-## How to get a Management APIv2 Token
+There are two ways to get a Management APIv2 Token: [get one manually using the Dashboard](#get-a-token-manually), or [automate the process](#automate-the-process) (build a simple command line tool that generates tokens). In this article we will see how you can do either.
 
-**Create a Non Interactive Client**
+## Get a token manually
 
-A Non Interactive Client is a client that interacts with an API where there is no user involved. It's a machine to machine interaction. This must be used instead of a Single Page or Native apps because those cannot meet the necessary security requirements for executing this type of flow.
+<div class="alert alert-info">
+  <strong>Heads up!</strong> Before you decide to follow the manual process, you should know that the Management APIv2 token has a validity of <strong>24 hours</strong>. After that the token will expire and you will have to get a new one. If this doesn't work for you, you can <a href="#automate-the-process">automate the process</a>.
```
mauricioschneider: I would go with the Warning background instead (yellow one) since this is a breaking change.

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
+  :::
+
+If you followed this article so far, then you already are at the _Non Interactive Clients_ tab of the `Auth0 Management API`. If not, go to [APIs](${manage_url}/#/apis), click the **Edit** icon for `Auth0 Management API`, go to _Non Interactive Clients_ and expand your non interactive client, using the pointing down arrow, next to the _Authorized_ toggle.
+
+All the scopes, for the various [Management API v2](/api/management/v2) endpoints are listed here. Tick the ones you want to grant to the client and click the **Update** button.
+
+  ::: panel-info I don't know which scopes to choose!
+  To find out which scopes you should choose, go the [Management API v2](/api/management/v2) and note the scopes listed at each endpoint. For example, the [Get all Clients](/api/management/v2#!/Clients/get_clients) endpoint requires the scopes `read:clients` and `read:client_keys`, while the [Create a Client](/api/management/v2#!/Clients/post_clients) endpoint requires the scope `create:clients`. If your client needs to read and create users, then you need to tick the checkboxes for these three scopes: `read:clients`, `read:client_keys` and `create:clients`.
```
Annyv2: This is also explained in the "what are the scopes" section with the same examples so it's a little bit repetitive.

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
+  <strong>Heads up!</strong> Before you decide to follow the manual process, you should know that the Management APIv2 token has a validity of <strong>24 hours</strong>. After that the token will expire and you will have to get a new one. If this doesn't work for you, you can <a href="#automate-the-process">automate the process</a>.
+</div>
 
-To create a new Non Interactive Client, go to the [Clients section](${manage_url}/#/clients) of the dashboard then click the **CREATE CLIENT** button.
+Let's see how you can get a token manually. Note, that the first two steps of the process need to be executed _only_ the first time. This might be true also for the third step, if the endpoints you need to access do not change. In that case you can skip to the [Get the Token](#4-get-the-token) section.
 
-Enter a name for your new client, select **Non Interactive Clients** and then click the **CREATE** button.
+### 1. Create a Client
+
+First, you need to create a Non Interactive Client. If you already have one skip this paragraph.
```
mauricioschneider: Should we recommend customers to create a Non Interactive Client exclusively for authorizing access to the Mgmt API and not just reuse an existing NIC for that purpose?

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
 
-**Get the token**
+### 4. Get the Token
 
-To test the client interaction with the API, go to the **Test** section under **Auth0 Management API**.
+To get a token, go to the *Test* section under `Auth0 Management API`.
+
+This page will give you code snippets on how to form a request to get a token.
+
+Check the _Response_, a token is already generated for you. Click **Copy Token**.
```
Annyv2: I don't see this button in my tenant, will it be added later for all tenants?

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
-Enter a name for your new client, select **Non Interactive Clients** and then click the **CREATE** button.
+### 1. Create a Client
+
+First, you need to create a Non Interactive Client. If you already have one skip this paragraph.
+
+  ::: panel-info What is a Non Interactive Client?
+  A Non Interactive Client represents a program that interacts with an API where there is no user involved. An example would be a server script that would be granted access to consume a Zip Codes API. It's a machine to machine interaction. This must be used instead of a Single Page or Native apps because those cannot meet the necessary security requirements for executing this type of flow. If you want to read more about calling APIs this way, refer to [Calling APIs from a Service](/api-auth/grant/client-credentials).
+  :::
+
+Go to the [Clients section](${manage_url}/#/clients) of the dashboard and click the **Create Client** button.
```
mauricioschneider, line 35: Can we recommend using a descriptive name for the client, like "Auth0 Management API Client"?

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
+  main()
 ```
 
-You can then change the `Client Secret` in that page to change your Management APIv2 secret.
+## Frequently Asked Questions
+
+__How long is the token valid for?__</br>
+The Management APIv2 token has a validity of __24 hours__. After that the token will expire and you will have to get a new one.
+
+__I hate this! The Management APIv2 token used to never expire, why would you change that?__</br>
```
Annyv2: Perhaps we should remove the hate this part? and put something like "why was it changed?" instead of "why would you" to make the change seem more natural and avoid telling users that it has generated some discomfort

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
+This python script gets a Management API v2 access token, uses it to call the [Get all clients](/api/management/v2#!/Clients/get_clients) endpoint, and prints the response in the console:
+
+```python
+def main():
+  import json, urllib, urllib2
+
+  # Configuration Values
+  DOMAIN = "${account.namespace}"
+  AUDIENCE = "https://${account.namespace}/api/v2/"
+  CLIENT_ID = "${account.clientId}"
```
Annyv2: Maybe add a comment that explains that it's the client id and secret of the non interactive client created previously? and a note that it's the non interactive client the one who has to be authorized to access all clients by setting up the appropriate scope

articles/api/management/v2/tokens.md
```diff
@@ -1,86 +1,278 @@
...
-Toggle the slider to authorize your client.
+![Authorize Non Interactive Client](/media/articles/api/tokens/authorize-noninteractive.png)
 
-**Choose the scopes**
+Toggle the slider from `Unauthorized` to `Authorized` for your client.
 
-You will see the available scopes that you can be granted for this client (which can be revoked at any time). Choose the desired scopes and then click the **UPDATE** button.
+### 3. Choose the Scopes
+
+The last step, before you get a token, is to select which scopes should be granted to this client.
```
mauricioschneider: We should tell the customers that if they need different sets of scopes for different applications they plan to call the mgmt api from, they better create a new non-interactive client for each.

For example:
NIC1: create:clients, read:clients
NIC2: create:users,read:users

Instead of having just one non interactive client with all the previous scopes and re-use it in different clients (applications).

mauricioschneider (changes requested):
Review the manual token generation section.

articles/api/management/v2/tokens.md
mpaktiti: Hmm interesting, I thought it was there for all tenants. I will check with crew-2.

This is what I see:

image

mpaktiti: Not an issue, button is in place.

PR #3526, opened 2017-02-15
New doc on apiv2 tokens

mpaktiti:
Explains why we removed the token generator and mentions a "work-around" for creating tokens the old way which is using http://JWT.io or any JTW library.


articles/api/management/v2/tokens-flows.md
```diff
@@ -0,0 +1,72 @@
...
+#### Why this changed
+
+Having a token that never expires can be very risky, in case an attacher gets hold of it. If the token expires within a few hours the attacker has only a small window to access your protected resources.
+
+
+## Can I still get a non-expiring token?
+
+Using Auth0 UI, no you cannot. However, if you need a work-around to the old way, you can generate a token using [JWT.io](https://jwt.io/).
+
+<div class="alert alert-danger">Long-lived tokens compromise your security. Following this process is <strong>NOT</strong> recommended.</div>
```
mauricioschneider: It's also worth mentioning that you cannot revoke them, something like "Long-lived and non-revocable tokens"

articles/api/management/v2/tokens-flows.md
```diff
@@ -0,0 +1,72 @@
...
+
+Once this is done you can get a new token either [using the dashboard](#4-get-the-token) (if you use the API sporadically) or [by configuring a server process](#1-get-a-token) (if you need a token frequently) that will get a new token every 24 hours.
+
+<div class="alert alert-info">For details on how to follow this new process refer to <a href="/api/management/v2/tokens">The Auth0 Management APIv2 Token</a></div>
+
+#### Why this changed
+
+In order to generate the token, the Management API had access to your __Global Client Secret__ (used to sign the token). This is information that should __not__ be exposed to web browsers.
+
+The new OAuth 2.0 Client Credentials grant implementation does not pose such risks, since the secret is not exposed on your browser. Once you do the initial configuration, you can get a token either by visiting the dashboard, or by making a simple `POST` request to `/oauth/token`.
```
mauricioschneider: I would go into a little more detail at the end:
```
request to `/oauth/token`.

vs

request to the `/oauth/token` endpoint of our Authentication API.
```

articles/api/management/v2/tokens-flows.md
mpaktiti: Mauricio, I'm going to change this back. You can revoke these tokens using https://auth0.com/docs/api/management/v2#!/Blacklists/post_tokens I tested and it works. The only thing required is that the token has the `jti` claim.

PR #3545, opened 2017-02-17
Update diagram for User Migration

khorne3:


khorne3:
@mpaktiti Extra "Yes" branch removed.

PR #3557, opened 2017-02-20
Fixed typo

cesalberca:
Changed AuthServce for AuthService


mpaktiti:
Good catch, thanks @cesalberca ! 👍

PR #3562, opened 2017-02-21
Fix curl example for password exchange in /oauth/token

santiagorp:
Added missing `https://` to curl example for *Resource Owner Password* grant in `/oauh/token` endpoint.

Without this fix the command will not work unless the user figure out that the `https://` is missing in the url.


mpaktiti:
Good catch @santiagorp thanks! 👍

PR #3563, opened 2017-02-21
Move Note in step 4

DonGlover:
It was below the image. Moved it before the image where it will be seen.


mpaktiti:
Yeap, better that way. Thanks for the contribution @DonGlover ! 👍

PR #3566, opened 2017-02-21
Update docs for pwd and refresh token rules

jhiner:
We are adding Rules execution to the Password and Refresh token exchanges. This PR is to update the documentation reflecting these changes and also the migration notice.

### DIFF
--- a/articles/rules/redirect.md
+++ b/articles/rules/redirect.md
@@ -13,7 +13,10 @@ Rules can also be used to programatically redirect users before an authenticatio
 * Forcing users to change passwords.
 
 ::: panel-danger Caution
-Redirect rules won't work for the [Resource Owner endpoint](/api/authentication/reference#resource-owner) or the [Refresh Token exchange](/tokens/preview/refresh-token#rules). You can detect resource owner logins from a rule by checking `context.protocol === 'oauth2-password'`. To detect a Refresh Token exchange, check `context.protocol === 'oauth2-refresh-token'`.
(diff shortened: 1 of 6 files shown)


jhiner:
These are WIP for now. I will follow up with someone on the docs team when they can be merged (probably later this week, pending the doc's team review).

articles/rules/redirect.md
```diff
@@ -13,7 +13,7 @@ Rules can also be used to programatically redirect users before an authenticatio
 * Forcing users to change passwords.
 
 ::: panel-danger Caution
-Redirect rules won't work for the [Resource Owner endpoint](/api/authentication/reference#resource-owner) authentication endpoint. You can detect resource owner logins from a rule by checking `context.protocol === 'oauth2-resource-owner'`.
+Redirect rules won't work for the [Resource Owner endpoint](/api/authentication/reference#resource-owner) or the [Refresh Token exchange](/tokens/preview/refresh-token#rules). You can detect resource owner logins from a rule by checking `context.protocol === 'oauth2-password'`. To detect a Refresh Token exchange, check `context.protocol === 'oauth2-refresh-token'`.
 :::
```
jhiner: One thing here is that the OLD Resource Owner endpoint is not going away, so we need to make sure its mention stays in this doc. We need to add oauth/token refresh token and password exchanges. Customers can check oauth2-resource-owner if it's the oauth/ro endpoint, or oauth2-password if oauth/token (password) or oauth2-refresh-token if oauth/token (refresh_token)

PR #3568, opened 2017-02-21
Update 03-user-sessions.md

huguesbr:
Update based on sample.

It seem there is a bug in SDK... It won't work using `oidcConformant = true`....

It was really frustrating to try it and not having it working until I did this fix.
The base `Lock.classic().onAuth {}` work but the later call `Auth0.authentication().userInfo(token: token).start()...` will fail with a unknown error... (even if the API call seem to succeed)

https://github.com/auth0-samples/auth0-ios-swift-v2-sample/blob/master/03-User-Sessions/Auth0Sample/HomeViewController.swift#L49-L52

### DIFF
--- a/articles/native-platforms/ios-swift/03-user-sessions.md
+++ b/articles/native-platforms/ios-swift/03-user-sessions.md
@@ -61,7 +61,9 @@ Once the user has logged in, you get a `Credentials` object, as follows:
 Lock
     .classic()
     .withOptions {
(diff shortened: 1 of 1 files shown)


cocojoe:
This is a known issue that is being resolved, this is why in the sample code it doesn't use `oidcConformant`.  However can appreciate the frustration here, let me review and fix this in another PR I have open.

PR #3577, opened 2017-02-22
Add "description" property information to Clients overview.

Cherna:
As added in the following PRs:
- https://github.com/auth0/api2/pull/1469
- https://github.com/auth0/manage/pull/2781

We added a `description` property to clients and this should be reflected in our documentation.

This additions should be reviewed to make sure we are not missing any occurrence of the Clients properties list where we should add this also. 

image


mpaktiti:
Hey @Cherna , I checked, this is the only doc we list all Client Settings. The change is live so I'm merging this PR. Thanks! 👍

mpaktiti:
Minor change: I'm moving this after the Client Secret so the listing follows the same order as they are displayed in the dashboard.

Cherna:
@mpaktiti Thanks :D

PR #3580, opened 2017-02-23
Update Auth API auth0.js samples to v8

angelaLazar:
First part of updating the Auth API auth0.js samples to v8 

https://trello.com/c/pN2CGhqi/826-authentication-api-update-auth0-js-sam
ples-to-v8

### DIFF
--- a/articles/api/authentication/_login.md
+++ b/articles/api/authentication/_login.md
@@ -15,6 +15,7 @@ GET https://${account.namespace}/authorize?
 ```
 
 ```javascript
+// Script uses auth0.js v8. See Remarks for details.
 <script src="${auth0js_urlv8}"></script>
 <script type="text/javascript">
   var webAuth = new auth0.WebAuth({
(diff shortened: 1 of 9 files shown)


angelaLazar:
@mpaktiti I think this should be ready to merge now

articles/api/authentication/_login.md
```diff
@@ -229,59 +195,41 @@ GET https://${account.namespace}/authorize?
 ```
 
 ```javascript
-<script src="${auth0js_url}"></script>
+// Script uses auth0.js v8. See Remarks for details.
```
mpaktiti, line 198: @jeffreylees This auth0.js script is not valid, it has username/password while the passive authentication is just a redirect to the Auth0 Login Page that will show the Login Widget where the user can login manually with email and password. 

I think we need something like the [buildAuthorizeUrl](https://github.com/auth0/auth0.js#auth0authentication). Can you please check with brucke and find the correct script for this case? Maybe we should add this to the auth0.js index as well.

jeffreylees: Yes, `buildAuthorizeUrl` is what we need to use. It takes the params that an authorize request would need.

```
webAuth.client.buildAuthorizeUrl({
  clientID: 'YOUR_CLIENT_ID', // string
  responseType: 'token', // code or token
  redirectUri: 'http://example.com',
});
```

Something like this? Also will want to recommend setting a `state` probably.

mpaktiti: Thanks Jeff! I tried it, works fine. 👍

PR #3582, opened 2017-02-23
Update tenant settings cURL call to post p2 merge

twistedstream:
Since we're now post-p2 merge we don't need to set the `enable_pipeline2` flag. Also, the `enable_dynamic_client_registration` property needs to live inside of a `flags` object in the request. I think this was a typo in the original version of this doc.


mpaktiti:
Looks good, thanks for fixing this @twistedstream ! 👍

PR #3583, opened 2017-02-23
updated vue tutorial to v2

vmartynets:
Corresponding to https://github.com/auth0-samples/auth0-vue-samples/pull/14


chenkie:
Thanks for this. We're just about to push the P2 samples live which uses Vue 2, so closing this.

PR #2328, opened 2016-09-16
[DO NOT MERGE] Objective-C Quickstart

mpaktiti:
Initial content taken from PRs:
00-introduction: https://github.com/auth0/docs/pull/1878
01-login: https://github.com/auth0/docs/pull/1833
04-user-profile: https://github.com/auth0/docs/pull/1724
05-linking-accounts: https://github.com/auth0/docs/pull/1775
06-rules: https://github.com/auth0/docs/pull/1818
07-authorization: https://github.com/auth0/docs/pull/1819
08-calling-APIs: https://github.com/auth0/docs/pull/1879
09-MFA: https://github.com/auth0/docs/pull/1877
10-customizing-lock: https://github.com/auth0/docs/pull/1789

PR #3588, opened 2017-02-24
Quickstart obj-c multistep

cocojoe:
Maintenance update to PR #2328
Reworked Intro/Login inline with current guides
Code Tweaks
Formatting
Copy Tweaks
Fixed broken links
Fixed Project Specs


articles/native-platforms/ios-objc/00-introduction.md
```diff
@@ -0,0 +1,130 @@
...
+</plist>
+```
+
+## Done!
+
+**You're all set up to start integrating Auth0 in your app!** 🎉
+
+The tutorial guides contained in this section will teach you the different things you can do with it. It's recommendable that you follow them sequentially in order, as they are progressive and there is content that usually needs to be learned from previous tutorials.
+
+Said that, we suggest you to start with the [login tutorial](01-login.md)...
```
jeffreylees: Should be "That said," rather than "Said that," or you could remove and start with "We suggest...".

Also in the paragraph above this, I'd probably use "It's recommended" - just sounds better to me, although there's nothing wrong with how it is.

articles/native-platforms/ios-objc/02-custom-login.md
```diff
@@ -158,25 +156,23 @@ You can use the `profile.name` to show the user's name or show the users profile
...
-                // You signed up correctly, and you have the Credentials,
-                // so no need to make the user login. They are already in.
-            }
-        });
-}];
+                        if(error) {
+                            // Something went wrong, let the user know
+                        } else {
+                            // You signed up correctly, and you have the Credentials,
+                            // so no need to make the user login. He's already in.
```
jeffreylees: We try to stay gender neutral for users, so probably "They are already in" was accurate, or "The user is already in". Or just remove that last line "He's already in."

articles/native-platforms/ios-objc/10-customizing-lock.md
```diff
@@ -0,0 +1,192 @@
...
+```objc
+[theme registerColor: [UIColor redColor] forKey: A0ThemePrimaryButtonNormalColor];
+```
+So, for example, if you want to achieve something like this:
+
+<div class="phone-mockup"><img src="/media/articles/native-platforms/ios-swift/Custom-Lock-Widget-Screenshot.png" alt="Mobile example screenshot"/></div>
+
+You will need to customize several parts. This is how you do it:
+
+```objc
```
jeffreylees: It might be better to use an ordered list here and just indent the code blocks to stay inline under their list item, in place of the commented numbering inside each block.

jeffreylees (commented):
Just a few things to be copyedited that I noticed, nothing big.

PR #3605, opened 2017-02-28
Add info on npm test packages

khorne3:

### DIFF
--- a/articles/connections/database/mysql.md
+++ b/articles/connections/database/mysql.md
@@ -29,9 +29,9 @@ Click **Custom Database** and turn on the **Use my own database** switch.
 
 You have to provide a login script to authenticate the user that will execute each time a user attempts to log in. Optionally, you can create scripts for sign-up, email verification, password reset and delete user functionality.
 
-::: panel-info Note
-When creating users, Auth0 calls the `get_user` script before the `create` script. Be sure that you have implemented both.
-:::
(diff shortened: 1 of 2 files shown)


articles/connections/database/mysql.md
```diff
@@ -79,6 +79,10 @@ This script connects to a **MySQL** database and executes a query to retrieve th
 
 If you are using [IBM's DB2](https://www.ibm.com/analytics/us/en/technology/db2/) product, [click here](/connections/database/db2-script) for a sample login script.
 
+::: panel-info
+You can use the `auth0-custom-db-testharness` [library](https://www.npmjs.com/package/auth0-custom-db-testharness) to deploy, execute, and test the output of Custom DB Scripts using a Webtask sandbox environment.
+:::
+
```
mpaktiti: @khorne3 I don't know why this keeps looking like this:

image

Maybe it's the missing title? When I'm adding it usually it gets fixed. Anyway, I will replace this with an [alert](https://github.com/auth0/docs/blob/master/CONTRIBUTING.md#alerts). I think it looks better for short notes, wdyt?

khorne3: I don't know why it does that. I noticed you removed some of my titles, so I stopped adding them.

Alerts are good--I'll start using those instead of panels.

PR #3607, opened 2017-03-01
Do not base64 decode client secret

zlotnika:
Apps created past 2016-12-06 do not have their secrets base64 encoded.


chenkie:
Thanks for your PR! looks like something got updated in the meantime. Can you resolve the conflict and resubmit?

zlotnika:
sure thing

chenkie:
thanks!

PR #3627, opened 2017-03-03
Set auth0jsv8 variable

mpaktiti:


mpaktiti:
@jeffreylees That's a var created by Angela for the API auth0.js scripts https://github.com/auth0/docs/blob/master/config/vars.yml#L4 Thought we could use it here.

jeffreylees (approved):
Sounds fine to me, and then if we ever get to a place where they're all using the same versions, we could re-assess, but having separate variables for 7 and 8 sounds like a good idea to me.

PR #3663, opened 2017-03-08
Mentioned required parameters when executing `passwordlessVerify` met…

dctoon:
Mentioned that the properties `redirectUri` and `responseType` must be set when initialising `WebAuth` when calling the password less verify method `passwordlessVerify`.

If one of the properties is missing you either get the message `invalid_request: Missing required parameter: response_type` or `server_error: Unable to issue redirect for OAuth 2.0 transaction` which is not clear for the consumer of the library.

### DIFF
--- a/articles/libraries/auth0js/index.md
+++ b/articles/libraries/auth0js/index.md
@@ -208,6 +208,7 @@ In addition, _one_ of the two following options must be sent:
 * `phoneNumber`: a string containing the user's phone number, to which the code or link was delivered via SMS
 * `email`: a string containing the user's email, to which the code or link was delivered via email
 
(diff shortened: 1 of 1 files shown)


jeffreylees (approved):
Maybe a slight re-word?:

> Note that in order to use `passwordlessVerify`, the options `redirectUri` and `responseType: 'token'` **must** be specified when first initializing `WebAuth`. 

@dctoon we will be going through the libraries ensuring that each parameter for each method is carefully documented - probably next week, which should offer some improvement. Thanks for this catch!

dctoon:
@jeffreylees applied re-word feedback

PR #3665, opened 2017-03-08
Fix for React error caused by TOC

nkohari:


nkohari:
This is pre-empted by removing the TOC from the document instead.

PR #3670, opened 2017-03-08
Saml test

khorne3:


khorne3:
Added a way for people to get and use their entity IDs, since I don't think we have a parameter alias for it currently. I used the HAR snippet format, even though it's probably likely the user will just use cURL and not actual code.

PR #3672, opened 2017-03-08
[Do not merge] Update dyn reg doc to use dashboard settings instead of API

rolodato:


khorne3:
Closing due to inactivity. Would be happy to continue work on this in the future.

PR #3680, opened 2017-03-10
Why you should always use access tokens to secure an API

mpaktiti:
New p2 doc

### DIFF
--- a/articles/api-auth/why-use-access-tokens-to-secure-apis.md
+++ b/articles/api-auth/why-use-access-tokens-to-secure-apis.md
@@ -1,0 +1,16 @@
+---
+title: Why you should always use access tokens to secure an API
+description: Explains the differences between access token and ID token and why the later should never be used to secure an API.
+---
+
+# Why you should always use access tokens to secure an API
+
+There is much confussion on the Web as to the differences between the OpenID Connect and OAuth 2.0 specifications, and their respective tokens. As a result many developers publish insecure applications, compromising their users security. The contradicting implementations between identity providers do not help either.
+
+This article is an attempt to clear what is what and convince you that you should always use an access token to secure an API, and never an ID token.
+
+## Two complementary specifications
+
+OpenID Connect is used for __user authentication__. It enables you to verify the identity of a user and get some basic profile information.
+
+OAuth 2.0 is used to __grant authorization__. It enables you to 
--- a/articles/api-auth/why-use-access-tokens-to-secure-apis.md
+++ b/articles/api-auth/why-use-access-tokens-to-secure-apis.md
@@ -39,6 +39,52 @@ Now that we saw what these tokens can be used for, let's see what they cannot be
 
 - __An `id_token` should not be used for API access__. Each token contains information on the intended audience (recipient). According to the OpenID Connect specification, the audience (claim `aud`) of each `id_token` must be the `client_id` of the client making the authentication request. If it isn't you shouldn't trust the token. An API, on the other hand, expects a token with the audience set to the API's unique identifier. So unless you are in control of both the client and the API, sending an `id_token` to an API will not work.
 
+## Compare Tokens
+
+To better understand what we just read, let's look at the contents of example tokens.
+
+The (decoded) contents of an `id_token` look like the following:
+
+```json
+{
+  "iss": "http://${account.namespace}/",
+  "sub": "auth0|123456",
+  "aud": "${account.clientId}",
+  "exp": 1311281970,
+  "iat": 1311280970,
+  "name": "Jane Doe",
+  "given_name": "Jane",
+  "family_name": "Doe",
+  "gender": "female",
+  "birthdate": "0000-10-31",
+  "email": "janedoe@example.com",
+  "picture": "http://example.com/janedoe/me.jpg"
+}
+```
+
+This token is meant for __authenticating the user to the client__. Note that the audience (`aud` claim) of the token is set to the client's identifier, which means that only this specific client should consume this token.
+
+For comparison, let's look at the contents of an `access_token`:
+
+```json
+{
+  "iss": "https://${account.namespace}/",
+  "sub": "auth0|123456",
+  "aud": [
+    "my-api-identifier",
+    "https://${account.namespace}/userinfo"
+  ],
+  "azp": "${account.clientId}",
+  "exp": 1489179954,
+  "iat": 1489143954,
+  "scope": "openid profile email address phone read:appointments email"
+}
+```
+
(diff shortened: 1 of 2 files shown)


articles/api-auth/why-use-access-tokens-to-secure-apis.md
```diff
@@ -0,0 +1,95 @@
...
+:::
+
+The `access_token` can be any type of token (not necessarily a JWT) and is meant for the API. Its purpose is to inform the API that the bearer of this token has been authorized to access the API and perform specific actions (as specified by the `scope` that has been granted). In the example we used earlier, after you authenticate, and provide your consent that the to-do application can have read/write access to your calendar, an `access_token` is sent from Google to the to-do application. Each time the to-do application wants to access your Google Calendar it will make a request to the Google Calendar API, using this `access_token` in an HTTP `Authorization` header.
+
+## How NOT to use each token
+
+Now that we saw what these tokens can be used for, let's see what they cannot be used for.
+
+- __An `access_token` should not be used for authentication__. The reason for that is that it holds no information about the user. It cannot tell us if the user has authenticated and when. That's why OAuth 2.0 cannot be used for authentication. Instead, OpenID Connect works on top of OAuth 2.0 and provides user authentication.
```
cabralmartin: maybe add a note saying that access_tokens should be treated as opaque by clients. clients should not attempt to decode them or depend on a particular access_token format.

articles/api-auth/why-use-access-tokens-to-secure-apis.md
```diff
@@ -0,0 +1,95 @@
...
+
+- __An `id_token` should not be used for API access__. Each token contains information on the intended audience (recipient). According to the OpenID Connect specification, the audience (claim `aud`) of each `id_token` must be the `client_id` of the client making the authentication request. If it isn't you shouldn't trust the token. An API, on the other hand, expects a token with the audience set to the API's unique identifier. So unless you are in control of both the client and the API, sending an `id_token` to an API will not work.
```
cabralmartin: the biggest case against id_tokens for api access is that the id_token (since it is issued to a particular client) is signed with a secret that is known to the client. if an API were to accept such token, it has no way of knowing if the client modified the token (say by adding more scopes) and then signed it again.

articles/api-auth/why-use-access-tokens-to-secure-apis.md
```diff
@@ -0,0 +1,95 @@
...
+    "https://${account.namespace}/userinfo"
+  ],
+  "azp": "${account.clientId}",
+  "exp": 1489179954,
+  "iat": 1489143954,
+  "scope": "openid profile email address phone read:appointments email"
+}
+```
+
+This token is meant for __authorizing the user to the API__. As such, it is completely opaque to clients, meaning that clients should not care about the contents of this token. Note that the token does not contain any information about the user itself besides their ID (`sub` claim), it only contains authorization information about which actions the client is allowed to perform at the API (`scope` claim).
```
cabralmartin: excellent! you may disregard my previous note on the subject of opaque tokens or reinforce it at the token.

articles/api-auth/why-use-access-tokens-to-secure-apis.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Why you should always use access tokens to secure an API
+
+There is much confusion on the Web about the differences between the OpenID Connect and OAuth 2.0 specifications, and their respective tokens. As a result many developers publish insecure applications, compromising their users security. The contradicting implementations between identity providers do not help either.
+
+This article is an attempt to clear what is what and explain why you should always use an access token to secure an API, and never an ID token.
+
+## Two complementary specifications
+
+OpenID Connect is used for __user authentication__. It enables you, the user, to verify your identity and give some basic profile information, without sharing your credentials.
```
cabralmartin: OpendId is a specification that augments OAuth. Oauth only dealt with authorization and doesnt care about how the authorization server verifies who is the user (authentication). this was a problem because home grown solutions for this were created (e.g. facebook connect).
openid augmented oauth by specifying how the AS can authenticate users. as a result, the spec is everything oauth had + authentication (the collateral of this was the introduction of id_tokens).

jaredhanson:
LGTM

articles/api-auth/why-use-access-tokens-to-secure-apis.md
```diff
@@ -0,0 +1,101 @@
...
+    "https://${account.namespace}/userinfo"
+  ],
+  "azp": "${account.clientId}",
+  "exp": 1489179954,
+  "iat": 1489143954,
+  "scope": "openid profile email address phone read:appointments email"
+}
+```
+
+This token is meant for __authorizing the user to the API__. As such, it is completely opaque to clients, meaning that a client should not care about the contents of this token, decode it or depend on a particular token format. Note that the token does not contain any information about the user itself besides their ID (`sub` claim), it only contains authorization information about which actions the client is allowed to perform at the API (`scope` claim).
```
twistedstream, line 90: I'm not sure saying that the access token is meant for "authorizing the user to the API" is accurate. Isn't it really about "authorizing the client to the API"? OAuth2 doesn't really concern itself with a user per se. It provides flows that allow users to grant access, but the access is being granted to the client, not themselves. User authorization is something that has already occurred within the Authorization Server.

cc @jaredhanson

jaredhanson: I understand what you are getting at here, and you are technically correct.  It's impossible, in the strictest sense, for a human user to interact _directly_ with an API.  There's always a software-based agent (aka, the client) operating on behalf of the user that allows interaction with the API.

Keeping things in layman's terms, I have no issue with the phrase "authorizing the user to the API" since its for all intents and purposes usually the desired outcome.

articles/api-auth/why-use-access-tokens-to-secure-apis.md
ChristianCiach: > the id_token (since it is issued to a particular client) is signed with a secret that is known to the client

I don't think this is true. The id_token is signed using the private key of the IDP, so by validating the id_token (which is a JWT) you can validate that the token really was issued by the IDP, thus, the included information about the user is valid.

cabralmartin: @ChristianCiach It really depends on the signing algorithm. If the `id_token` is using a symmetric key algorithm (such as HS256) the key used is the client secret (this is [per spec](http://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation)). If the `id_token` is signed using one of the asymmetric key algorithms (e.g. RS* family) then Auth0 will use the tenant private key for signing it.

PR #3695, opened 2017-03-14
Authentication API versioning

mpaktiti:
- Add legacy version for Authentication API
- Add panels at the top of each endpoint to describe the version and point to the other one


mpaktiti:
The Authentication API is a peculiar case, versioning won't work for it with the as-is implementation.

Issue #2407, opened 2016-09-29
Update Rules document


cabralmartin:
[https://auth0.com/docs/rules](https://auth0.com/docs/rules)
- Add a note explaining that Rules are run _after_ authentication and _before_ authorization
- Explain how can rules can modify the `scope` property on the access token by overriding it

PR #3723, opened 2017-03-16
Update Rules for pipeline2

mpaktiti:
Fixes https://github.com/auth0/docs/issues/2407

- Add a note explaining that Rules are run after authentication and before authorization

- Explain how can rules can modify the scope property on the access token by overriding it


articles/rules/preview/index.md
```diff
@@ -0,0 +1,312 @@
...
+  callback(null, user, context);
+}
+```
+
+<div class="alert alert-info">
+  You can add <code>console.log</code> lines for <a href="#debugging">debugging</a> or use the <a href="/extensions/realtime-webtask-logs">Real-time Webtask Logs Extension</a>.
+</div>
+
+
+### Add roles to a user
```
mpaktiti: Is this example valid for p2?

articles/rules/preview/index.md
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+
+This will cause a redirect to your callback url with an `error` querystring parameter containing the message you set. (e.g.: `https://yourapp.com/callback?error=unauthorized&error_description=Only%20admins%20can%20use%20this`). Make sure to call the callback with an instance of `UnauthorizedError` (not `Error`).
+
+<div class="alert alert-info">
+  Error reporting to the app depends on the protocol. OpenID Connect apps will receive the error in the querystring. SAML apps will receive the error in a <code>SAMLResponse</code>.
+</div>
+
+### API Authorization: Modify Scope
```
mpaktiti: Added new example

articles/rules/preview/index.md
```diff
@@ -0,0 +1,312 @@
...
+  // change scope
+  context.accessToken.scope = ['array', 'of', 'strings'];
+
+  callback(null, user, context);
+}
+```
+
+The user will be granted three scopes: 'array', 'of', and 'strings'.
+
+### API Authorization: Add Claims
```
mpaktiti: Added new example

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+
+* __Profile enrichment__: query for information on the user from a database/API, and add it to the user profile object.
+* Create __authorization rules__ based on complex logic (anything that can be written in JavaScript).
+* __Normalize attributes__ from different providers beyond what is provided by Auth0.
+* Reuse information from existing databases or APIs for migration scenarios.
+* Keep a __white-list of users__ and deny access based on email.
+* __Notify__ other systems through an API when a login happens in real-time.
+* Enable counters or persist other information. (For information on storing user data, see: [Metadata in Rules](/rules/metadata-in-rules).)
+* Enable __multifactor__ authentication, based on context (e.g. last login, IP address of the user, location, etc.).
+* Modify tokens: Change the returned __scopes__ of the `access_token` and/or add claims to it, and to the `id_token`. At the moment you can do this for the [Password Exchange](/api-auth/grant/password#rules).
```
cabralmartin: rules can modify tokens on any flow, not just password.

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+**Rules** are functions written in JavaScript that are executed in Auth0 as part of the transaction every time a user authenticates to your application. They are executed after the authentication and before the authorization.
+
+Rules allow you to easily customize and extend Auth0's capabilities. They can be chained together for modular coding and can be turned on and off individually.
+
+![](/media/articles/rules/flow.png)
+
+* **Step 1:** An app initiates an authentication request to Auth0.
+* **Step 2:** Auth0 routes the request to an Identity Provider through a configured connection.
+* **Step 3:** The user authenticates successfully.
+* **Step 4:** The `user` object representing the logged in user is passed through the Rules pipeline, and returned to the app.
```
cabralmartin: in the new flows, we don't return the user object to the app. we return any of {id_token, token, code}.

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+
+To create a Rule, or try the examples below, go to [New Rule](${manage_url}/#/rules/create) in the Rule Editor on the dashboard.
+
+### *Hello World*
+
+This rule will add a `hello` attribute to all users authenticating through any provider.
+
+```js
+function (user, context, callback) {
+  user.hello = 'world';
```
cabralmartin: this is no longer valid in the new flows. this should be
`context.idToken["http://mynamespace/hello"] = "world";`

articles/rules/preview/index.md
cabralmartin: use case yes, example no. in the past, we would modify the user object because we would then merge it with the id_token and would return the whole package.
in the new implementation, we need to actually modify the id_token with what we want in it.

`context.idToken["http://sample"] = ['role1', 'role2', ... ]` <- pseudo code, need to actually try it and figure the actual grammar.

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+```js
+{
+  "email": "johnfoo@gmail.com",
+  "family_name": "Foo",
+  "user_id": "google-oauth2|103547991597142817347"
+  //... other properties ...
+}
+```
+
+The `context` object will be:
```
cabralmartin: context may include `idToken` and `accessToken` properties. this is important because now a lot of the old behavior is now handled by these two properties.

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+  "clientName": "my app",
+  "clientMetadata": {
+    "myKey1": "myValue2",
+    "myKey2": "myValue2"
+  }
+  "connection": "google-oauth2"
+}
+```
+
+After the rule executes, the output that the application will receive is the following `user` object:
```
cabralmartin: the following `id_token`

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+
+```js
+{
+  "email": "johnfoo@gmail.com",
+  "family_name": "Foo",
+  "user_id": "google-oauth2|103547991597142817347",
+
+  //... other props ...
+
+  "roles": ["guest", "admin"]  // NEW PROPERTY ADDED BY THE RULE
```
cabralmartin: this won't be valid on the new pipeline

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+
+This will cause a redirect to your callback url with an `error` querystring parameter containing the message you set. (e.g.: `https://yourapp.com/callback?error=unauthorized&error_description=Only%20admins%20can%20use%20this`). Make sure to call the callback with an instance of `UnauthorizedError` (not `Error`).
+
+<div class="alert alert-info">
+  Error reporting to the app depends on the protocol. OpenID Connect apps will receive the error in the querystring. SAML apps will receive the error in a <code>SAMLResponse</code>.
+</div>
+
+### API Authorization: Modify Scope
+
+This will override the returned scopes of the `access_token`. The rule will run after user authentication and before authorization.
```
cabralmartin: we need to review with crew-auth. i know we can run rules now even after consent.

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+function(user, context, callback) {
+
+  // change scope
+  context.accessToken.scope = ['array', 'of', 'strings'];
+
+  callback(null, user, context);
+}
+```
+
+The user will be granted three scopes: `array`, `of`, and `strings`.
```
cabralmartin: excellent

articles/rules/preview/index.md
```diff
@@ -0,0 +1,314 @@
...
+
+### API Authorization: Add Claims
+
+This will add one custom namespaced claim at the `access_token`, and two at the `id_token`.
```
cabralmartin: this example is great, and is the general direction we are taking on the new pipeline. you no longer modify the "user" object and that auto magically gets to the app. now the power is on the rule. if you want something on the id_token you need to explicitly code for it on a rule by adding claims to the id_token/access_token. this is important because no longer every client gets a full dump of the data of the user, but the authorization server (via the rules) can say "to this client this information"

cabralmartin: also, this is what we will use for every other use case that needed modification of the user object. you can add custom claims to send roles information, extended profile properties, etc

PR #3752, opened 2017-03-21
OKTA OpenID Connect identity provider

CriGoT:
Adds article describing how to setup OKTA as an Identity Provider using OpenID Connect


CriGoT:
Since there are no other documents for OpenID identity providers I created a new location but I'm not sure is the right one.

PR #3766, opened 2017-03-23
Update Scopes for pipeline2

mpaktiti:


articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+toc: true
+---
+
+# Scopes
+
+<div class="alert alert-info">
+<strong>Heads up!</strong> This article describes the latest Auth0 implementation, according to the <a href="/api-auth">API Authorization flows</a>. If you are looking for the legacy Scopes doc refer to <a href="/scopes">Scopes</a>.
+</div>
+
+In OpenID Connect (OIDC) we have the notion of __scopes__, or __claims__. These are strings, sent as part of the `scope` request parameter. These claims can be standard (as defined by the specification) or custom.
```
cabralmartin: `scope` and `claim` while they look similar they are a lil different. A `claim` is an extended attribute that depending on the token can be said to be related to the user identity (say `email`) or to the users access rights (custom claim in access token).
`scope`s on the other hand refer to reach the client will have when executing on behalf of the user. `scope: openid` is the client asking to retrieve the user's information. `scope:read:todo` is the client trying to read the users TODOs.
In particular, a `scope:email` value is the client asking to retrieve the email of the user that being a standard claim.

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+Scopes are used in the following cases:
+
+- When you want to get an `id_token` in the response. In this case, you need to include `openid` as part of the `scope` request parameter: `scope=openid`.
```
cabralmartin: I'm not sure this is true. Including `scope:openid` will add `userinfo` as audience to the access token.
If users want to get an `id_token` they should use `response_type:id_token`

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+- When you want to get additional user information, like email or picture. For details refer to [Standard Claims](#standard-claims).
+
+- When you want to access an API. In this case, you need to define custom scopes for your API and add these values  as part of the `scope` request parameter, for example: `scope=read:contacts`. For details refer to [API Scopes](#api-scopes).
```
cabralmartin: i'd rephrase. `when you want to have granular access control in your api`.

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+
+## Standard Claims
+
+OpenID Connect specifies a set of [standard claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims). These are user attributes, intended to supply the client with user details such as email, name and picture.
+
+This information can be returned in the `id_token` or in the response from [the /userinfo endpoint](/api/authentication#get-user-info).
+
+__NOTE__: The `updated_at` claim is returned as a [Unix timestamp](https://en.wikipedia.org/wiki/Unix_time) instead of an [ISO 8601 date string](https://en.wikipedia.org/wiki/ISO_8601) for consistency with the `exp`, `iat` and `nbf` claims.
```
cabralmartin: We need to double check with @jhiner about this. I'm not sure if its an epoch or not.

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+After Auth0 has redirected back to the app, you can extract the `id_token` from the hash fragment of the URL.
+
+When decoded, this token contains the following claims:
+
+```json
+{
+  "name": "John Doe",
+  "nickname": "john.doe",
+  "picture": "https://myawesomeavatar.com/avatar.png",
+  "updated_at": "2017-03-30T15:13:40.474Z",
```
cabralmartin, line 62: here it is. it's not an epoch, we need to update the former entry.

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+  const namespace = 'https://myapp.example.com/';
+  context.idToken[namespace + 'favorite_color'] = user.favorite_color;
+  context.idToken[namespace + 'preferred_contact'] = user.user_metadata.preferred_contact;
+  callback(null, user, context);
+}
+```
+
+Any non-Auth0 HTTP or HTTPS URL can be used as a namespace identifier, and any number of namespaces can be used. The namespace URL does not have to point to an actual resource, it’s only used as an identifier and will not be called by Auth0. This follows a [recommendation from the OIDC specification](https://openid.net/specs/openid-connect-core-1_0.html#AdditionalClaims) stating that custom claim identifiers should be collision-resistant.
+
+While this is not mandatory according to the specification, Auth0 will always enforce namespacing, meaning that any non-namespaced claims will be silently excluded from tokens.
```
cabralmartin: promote to warning?

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+
+Once you create an API and define the scopes, the client applications can request these defined permissions when they initiate an authorization flow, and include them in the access token as part of the scope request parameter.
+
+If you wanted to expand [our example](#example-asking-for-standard-claims) to include also the `read:contacts` permission, then to initiate the authentication flow, using Implicit Grant, you would use something like this:
+
+```text
+https://${account.namespace}/authorize?
+  audience=YOUR_API_AUDIENCE&
+  scope=openid%20profile%20email%20read:contacts&
+  response_type=id_token&
```
cabralmartin: update response_type to include `token`

articles/scopes/preview/index.md
```diff
@@ -0,0 +1,143 @@
...
+  client_id=${account.clientId}&
+  redirect_uri=${account.callback}&
+  nonce=YOUR_CRYPTOGRAPHIC_NONCE
+  state=YOUR_OPAQUE_VALUE
+```
+
```
cabralmartin, line 138: can we add a short p with the diff from the former example? diff being the change on `scope`, `response_type`, and the adding of `audience`. would also help to show the new response

articles/scopes/preview/index.md
mpaktiti: This is copied from the adoption guide. I'll double check in any case.

articles/scopes/preview/index.md
mpaktiti: I'll follow up with Justin, this is in other docs as well, thanks.

articles/scopes/preview/index.md
mpaktiti: It is not at the moment (will be in the future). I'll change this, thanks.

PR #3806, opened 2017-03-29
Corrected URL for Configuring Customer Error Pages

pfilias:


pfilias:
Funny. Carlos and I were doing an Ops Handover call with Grean and the link to customize error pages was bad, leading to a 404. We all laughed!

mpaktiti:
Thanks @pfilias ! 👍

PR #3811, opened 2017-03-30
Fix typos in server-api.md

danedmunds:
Fix two small typos

`there is now interactive user` -> `there is no interactive user`
`which scopes will included` -> `which scopes will be included`


mpaktiti:
Thanks @danedmunds ! 👍

PR #3814, opened 2017-03-30
Change DNS name in sample hosts file

twistedstream:
Don't suggest .local since it has special behavior on Macs and can result in very slow DNS lookups.
We should suggest something else like .dev


mpaktiti:
Thanks @twistedstream ! 👍

PR #3817, opened 2017-03-31
updated swift dependencies, fixes #3738

### LINKED ISSUE #3738: Outdated Versions - Lock and Auth0 pod packages
In the iOS Swift Quickstarts on Step 0:

https://auth0.com/docs/quickstart/native/ios-swift/00-getting-started

There is an instruction telling the user to reference outdated versions:

pod 'Lock', '~> 2.0'
pod 'Auth0', '~> 1.2'

Those lines should be referencing the following versions:

pod 'Lock', '~> 2.1'
pod 'Auth0', '~> 1.4'

Or leave only the major versioning:

pod 'Lock', '~> 2'
pod 'Auth0', '~> 1'

outdatedversions

vmartynets:

### DIFF
--- a/articles/quickstart/native/ios-swift/_includes/_dependencies.md
+++ b/articles/quickstart/native/ios-swift/_includes/_dependencies.md
@@ -4,7 +4,7 @@ If you are using Carthage, add the following lines to your `Cartfile`:
 
 ```ruby
 github "auth0/Lock.iOS-OSX" ~> 2.1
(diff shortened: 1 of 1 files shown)


cocojoe:
The problem with these kinds of updates, is this is now out dated :) I am going to unify the dependencies so both Lock Library and Lock QS use same versions.  I'll close this as it's update in #4029

PR #3826, opened 2017-04-03
Auth0.js versioning and param docs, copy changes

jeffreylees:


mpaktiti:
@jeffreylees I made some changes at the alerts. Moved them to the top and updated the text so it instructs the user to use the dropdown, instead of providing links. If you apply versioning on another doc make sure you use the same format. Thanks.

PR #3830, opened 2017-04-04
Additional instructions for Azure AD

nicosabena:
Some additional instructions for allowing users from external directories, creating connections and giving consent.


angelaLazar:
Replaced by: https://github.com/auth0/docs/pull/3847

PR #3836, opened 2017-04-04
Add documentation about creating private routes with React Router v4

danidewitt:


chenkie:
thanks for your PR! some new docs have gone up since and they cover Router v4.

PR #3882, opened 2017-04-09
Update index.md

richardmarr:
Need to highlight differences between Private SaaS and Cloud


mpaktiti:
Thanks @richardmarr , we are actually working on a doc comparing cloud vs. Appliance and MFA is one of them.

PR #3900, opened 2017-04-12
Pass 'id_token' to tokenNotExpired function

mbennett:
The latest angular2-jwt version uses a default key of 'token' rather than 'id_token' in the tokenNotExpired function.  So this quickstart needs to pass 'id_token' to function properly.


chenkie:
Thanks for your PR! This fix actually went in from another PR. Much appreciated though.

PR #3907, opened 2017-04-13
Client Grant Types docs

khorne3:


articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
+---
+  description: Using the Grant Types property on Clients
+  toc: true
+---
+
+# Client Grant Types
+
+Depending on your use case, Auth0 provides many authorization flows that suit your needs. As such, you'll need to track the type of flow you're using for a given client. As such, Auth0 includes a `grant_type` property on each Client for this purpose.
```
jhiner: Recommended instead: 
Depending on your use case, Auth0 provides many authentication and authorization flows that suit your needs, and you may wish to limit the use of certain grant types for a particular client. Auth0 includes a `grant_type` property on each Client for this purpose.

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Grant Types
+
+The following is a list of grant types valid for Auth0 Clients. Each of the items represents one of the three types of authorization flows.
+
+There are three possible types of authorization flows:
+
+* Auth0-customized;
```
jhiner: I'd use this list instead:

Specification-conformant (such as Open ID Connect)
Auth0 extension
Auth0 legacy

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+* Legacy;
+* OIDC specification-conformant.
+
+The following is a list of Auth0-customized/OIDC specification-conformant `grant_types` that currently exist for use in Auth0:
```
jhiner: Shoudl be Auth0 extension/OIDC specification-conformant

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+* `http://auth0.com/oauth/legacy/grant-type/delegation/refresh_token`
+* `http://auth0.com/oauth/legacy/grant-type/delegation/id_token`
+* `http://auth0.com/oauth/legacy/grant-type/access_token`
+
+## Edit the `grant_type` Property
+
+To edit the `grant_type` property for your Auth0 Client, you'll need to make the appropriate call to the [Update a Client endpoint](/api/management/v2#!/Clients/patch_clients_by_id) of the [Management API](/api/management/v2).
+
+:::panel-warning Legacy Grant Types
+New Auth0 customers **cannot** add *any* of the legacy grant types to their Clients. Only customers as of TBD can add legacy grant types to their existing Clients.
```
jhiner: As of TBD, new Auth0 customers....

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+:::
+
+Auth0 requires the `grant_type` property, so using a Client *without* this property populated results in errors.
```
jhiner: Attempting to use any flow with a client lacking the appropriate `grant_type` for that flow will result in errors.

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Existing Clients
+
+For existing Clients, Auth0 will populate the `grant_type` property with the correct value so that this change doesn't affect functionality.
```
jhiner: Something like this might make it clearer that we are added ALL grant types to clients...

In order to avoid breaking existing customers, all existing Auth0 clients for all customers as of TBD will be populated with all Auth0 legacy, Auth0 extension, and spec-formant grant_types.

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+
+### New Clients
+
+At creation, new Clients (regardless of whether the owning tenant exists or is new), will be assigned the appropriate default grant type depending on the following two Client flags:
```
jhiner: Can probably remove `(regardless of whether the owning tenant exists or is new)`

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+
+* `oidc_conformant`: This flag is `true` if you've enabled the [**OIDC Conformant** slider](/api-auth/tutorials/adoption/oidc-conformant) in the *Advanced Settings* section of your Client Settings page.
+* `is_first_party`: By default, all new Clients will have this flag set to `true`.
+
```
jhiner: The challenge here is that by default in Dashboard, all clients are created:

- is_first_party: true,
- oidc_conformant: false

So we may need to clarify that newly created clients from the Dashboard will have these grant types, and you can only specify these options during client creation when you are doing it via APIv2.

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+* `client_credentials`
+* `refresh_token`
+* `http://auth0.com/oauth/grant-type/mfa-oob`
+* `http://auth0.com/oauth/grant-type/mfa-otp`
+* `http://auth0.com/oauth/grant-type/mfa-recovery-code`
+
+For third-party Clients, Auth0 limits used of the Password Exchange flows by default. To use this flow, you must explicitly enable it by making the appropriate `PATCH` call to the [Update a Client endpoint](/api/management/v2#!/Clients/patch_clients_by_id) of the [Management API](/api/management/v2).
+
+## Secure Alternatives to the Legacy Grant Types
+
```
jhiner: I owe you some more input on this specifically, to complete this info.

jhiner (changes requested):
Left some initial feedback.

articles/clients/grant-types/client-grant-types.md
```diff
@@ -0,0 +1,124 @@
...
+    <th>If you want to use this legacy grant type...</th>
+    <th>Consider using...</th>
+  </tr>
+  <tr>
+    <td>`http://auth0.com/oauth/legacy/grant-type/ro`</td>
+    <td>`http://auth0.com/oauth/grant-type/password-realm`</td>
+  </tr>
+  <tr>
+    <td>`http://auth0.com/oauth/legacy/grant-type/ro/jwt-bearer`</td>
+    <td></td>
```
jhiner: I think for these cases, for now we can say "Contact Support"

articles/clients/grant-types/client-grant-types.md
khorne3: @jhiner Is there another term that we can use instead of extension? I think it might get confused with Dashboard Extensions.

articles/clients/grant-types/client-grant-types.md
khorne3: @jhiner I'm a bit confused here. Do you want me to remove the table and just put "Contact Support" or do you want this for just this specific endpoint?

articles/clients/grant-types/client-grant-types.md
jhiner: Extension is the proper term, and we use it elsewhere such as here: https://auth0.com/docs/api-auth/grant/password#realm-support

articles/clients/grant-types/client-grant-types.md
jhiner: Sorry, should have been clearer...

I meant that for cases such as `http://auth0.com/oauth/legacy/grant-type/ro/jwt-bearer` where there is no current replacement secure feature available, we can say `Contact Support` (right now it's just an empty cell). Or, if you think it's better, we could say more like `If current customer as of (date), patch using APIv2 to use this. If new customer after (date), contact support.`

articles/clients/client-grant-types.md
```diff
@@ -0,0 +1,134 @@
...
+    <td>`http://auth0.com/oauth/legacy/grant-type/delegation/refresh_token`</td>
+    <td>Use the `oauth/token` endpoint to <a href="/api-auth/tutorials/adoption/refresh-tokens">obtain refresh tokens</a>. See <a href="/api-auth/tutorials/adoption/refresh-tokens">OIDC-conformant refresh tokens</a> for additional information.</td>
+  </tr>
+  <tr>
+    <td>`http://auth0.com/oauth/legacy/grant-type/delegation/id_token`</td>
+    <td>Contact Support for assistance.</td>
+  </tr>
+  <tr>
+    <td>`http://auth0.com/oauth/legacy/grant-type/access_token`</td>
+    <td>Contact Support for assistance.</td>
```
jhiner: For this, let's instead say "Use browser-based social authentication"

Instead of contact support...

khorne3: @jhiner Done.

jhiner:
@khorne3 We have changed our approach a little bit. I put the latest info here (https://docs.google.com/a/auth10.com/document/d/1-PvgdV5Ov732s7V9AfZMjIbXzD9qHECBiD_UVbOchwM/edit?usp=sharing): 

- note the error message that will be used if the proper grant type is not present on the client: “Grant type (grant type) not allowed for the client.”
- Newly created clients (for all existing and new tenants going forward) will get certain default grant types depending on whether or not it is a public or confidential clients:

Public clients will get these grant types:

implicit,
authorization_code,
refresh_token

Confidential clients will also get the client_credentials grant type:

implicit,
authorization_code,
refresh_token
client_credentials

Public clients have `token_endpoint_auth_method === 'none'` which is the value set for Native and Single Page clients created in Dashboard. Confidential clients have `token_endpoint_auth_method !== 'none'` and is the default for new clients with Regular Web App or Non-Intactive clients created in Dashboard. If a token_endpoint_auth_method is not specified, it is considered a Confidential client.

We will also allow customers to add these grant types to trusted first party clients (is_first_party: true):

password,
http://auth0.com/oauth/grant-type/password-realm,
http://auth0.com/oauth/grant-type/mfa-oob,
http://auth0.com/oauth/grant-type/mfa-otp,
http://auth0.com/oauth/grant-type/mfa-recovery-code

- Instead of `Contact Support for assistance` in the "features" table, let's say this: 
Because of the security implications, this feature is disabled by default. We ask that you contact our support to request that this feature be enabled. We will discuss the use case with you to ensure that usage of this doesn’t open your application up to an security vulnerabilities.

khorne3:
See https://github.com/auth0/docs/pull/4054

PR #3954, opened 2017-04-20
Changes in lock section

arielger:
- Remove the alerts from all Lock v10 articles (there is no need to alert the user that he is seeing the latest version). Only keep alerts for older versions.
- Move all alerts below the main title
- Make Lock section titles more consistent (add `Lock:` at the beginning, change titles)

### DIFF
--- a/articles/libraries/lock/v10/api.md
+++ b/articles/libraries/lock/v10/api.md
@@ -4,7 +4,7 @@ toc: true
 description: Details on the Lock V10 API.
 ---
 
-# API
+# Lock: API Reference
 
 Lock has many methods, features, and configurable options. This reference is designed to direct you to the ones that you need, and discuss how to use them. Click below to go straight the method you're looking for, or just browse! If you're looking for information about events emitted by Lock, they're listed under the [on()](#on-event-callback-) method section!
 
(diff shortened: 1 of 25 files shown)


jeffreylees:
>Remove the alerts from all Lock v10 articles (there is no need to alert the user that he is seeing the latest version). Only keep alerts for older versions.

As long as you don't think that will cause an uptick in people who use Lock v9 going to look for documentation to help them with something, not realizing that they're in the wrong version. I know that we specifically chose to use alerts on the latest version when v10 came out, but that was many months ago now, so it might very well be time to drop them. Just food for thought.

>Make Lock section titles more consistent (add Lock: at the beginning, change titles)

Since this is happening anyway, do you think there's any merit to doing a prefix of `Lock 10:` or `Lock 9:`, so that in search results one can tell the difference between two results that have similar, or the same names?

arielger:
I think that the version selector should be clear enough for the users to see that they are in the v10 doc.

Also i want to note that we have a problem with the v9 articles and the sidebar (the hierarchy of articles for older versions is not shown in the sidebar). It would be great to address this.

I think that prefixing all the titles with `Lock 10` will be redundant also. I think that the user should know that the default search result will correspond to the latest Lock version.

jeffreylees:
Taking the sidebar discussion to slack

PR #3956, opened 2017-04-21
Update Google Apps SAML configuration

neoeinstein:
Based on recent integration with GApps suite.

Following the requirements specified by Google, the NameID needs to be an email address.
Source: https://support.google.com/a/answer/6330801?hl=en

This changes the identifier format and probe to grab the email as expected.

Also uses sha256 for signature and digest because Google supports the newer hashing algorithm.


nicosabena (approved):
Looks good to me.

PR #3966, opened 2017-04-24
update generate Code Challenge Java example

tom-odon:
md.update() now accepts 'bytes' variable as an argument instead of 'input' variable in JavaDoc


mpaktiti:
Thank you @tom-odon ! 👍

PR #3976, opened 2017-04-24
Change "algorithms" property to "algorithm"

hmerzin:
Line 60: Changed `algorithms: ['RS256']` to `algorithm: 'RS256'` to avoid "invalid algorithm" error being thrown when trying to verify JWTs on the server.


chenkie:
hey thanks for your PR.

both `algorithm` and `algorithms` should work. can you confirm whether it's still an issue?

hmerzin:
Sure. I'll double check

kmaida:
I've confirmed that both work.

hmerzin:
Awesome! Maybe I was using an outdated version or something. Will reinstall tonight 👍

PR #3998, opened 2017-04-26
Clarify that /signup endpoint accepts a user_metadata parameter

angelaLazar:
https://trello.com/c/dBTIi8Wz/1202-add-clarification-to-dbconnections-si
gnup-endpoint-of-auth-api


mpaktiti:
Looks great! 👍

PR #4028, opened 2017-05-02
Add Xamarin iOS quickstart for OIDC Client

jerriep:
https://auth0-docs-content-pr-4028.herokuapp.com/docs/quickstart/native/xamarin-ios

The logo for this is not correct. Seem we will need to update the Styleguide CSS from `[data-name="xamarin"]` to `[data-name="^xamarin"]`

https://github.com/auth0/styleguide/blob/a1d51176adf55ff44025fc49282c97ce4622ac9f/packages/components/src/circle-logo/index.styl#L234


articles/quickstart/native/xamarin-ios/01-login.md
```diff
@@ -0,0 +1,171 @@
...
+
+Callback URLs are the URLs that Auth0 invokes after the authentication process. Auth0 routes your application back to this URL and appends additional parameters to it, including a token. Since callback URLs can be manipulated, you will need to add your application's URL to your client's *Allowed Callback URLs* for security. This will enable Auth0 to recognize these URLs as valid. If omitted, authentication will not be successful.
+
+First you will need to register the URL Type:
+
+1. Open your application's `Info.plist` file in Visual Studio for Mac, and go to the **Advanced** tab. 
+2. Under **URL Types**, click the **Add URL Type** button
+3. Set the **Identifier** as `Auth0`, the **URL Schemes** the same as your application's **Bundle Identifier**, and the **Role** as `None`
+
+This is an example of the XML representation of your `info.plist` file after you have added the URL Type:
```
cocojoe, line 53: `Info.plist`

articles/quickstart/native/xamarin-ios/01-login.md
```diff
@@ -0,0 +1,171 @@
...
+### Accessing the tokens
+
+On successful login, the login result will contain the `id_token` and `access_token` in the `IdentityToken` and `AccessToken` properties respectively.
+
+```csharp
+var loginResult = await client.LoginAsync();
+
+if (!loginResult.IsError)
+{
+    Debug.WriteLine($"id_token: {loginResult.IdentityToken}");
```
cocojoe, line 134: Wondering why there is a mixture of casing, you have `loginResult` then the credential properties are upper camel case e.g. `IdentityToken` also why not `idToken` inline with native iOS and Android?

articles/quickstart/native/xamarin-ios/01-login.md
```diff
@@ -0,0 +1,171 @@
...
+    Debug.WriteLine($"An error occurred during login: {loginResult.Error}")
+}
+```
+
+### Accessing the tokens
+
+On successful login, the login result will contain the `id_token` and `access_token` in the `IdentityToken` and `AccessToken` properties respectively.
+
+```csharp
+var loginResult = await client.LoginAsync();
```
cocojoe, line 130: be nicer to return a 'credentials' object inside the result and use that for accessing token properties?

articles/quickstart/native/xamarin-ios/01-login.md
```diff
@@ -0,0 +1,171 @@
...
+### Obtaining the User Information
+
+On successful login, the login result will contain the user information in the `User` property, which is a [ClaimsPrincipal](https://msdn.microsoft.com/en-us/library/system.security.claims.claimsprincipal(v=vs.110).aspx).
+
+To obtain information about the user, you can query the claims. You can for example obtain the user's name and email address from the `name` and `email` claims:
+
+```csharp
+if (!loginResult.IsError)
+{
+    Debug.WriteLine($"name: {loginResult.User.FindFirst(c => c.Type == "name")?.Value}");
```
cocojoe, line 148: This looks a bit ugly for the intended audience, can't we pull out the user object and access that, also can't we have a getter around this e.g.

`User.name`
`User.email`

etc

cocojoe (changes requested):
I feel some areas read  more like a proof of concept than a QS guide.  However this may just be a case of there has not been the TLC time for this one.

articles/quickstart/native/xamarin-ios/01-login.md
jerriep: Mixing of case = C# naming conventions
Naming of `IdentityToken` is because that is what the underlying library I am using decided to call it. I am returning their result from the `LoginAsync` exactly as-is

articles/quickstart/native/xamarin-ios/01-login.md
jerriep: I am returning it exactly as the underlying OIDC Client library is returning it. I am not altering or wrapping their response in any way. I wanted to keep the Auth0 OIDC Client library as a very thin wrapper as a very thing layer over the underlying OIDC Client library.

Also, changing that at this stage is a breaking change across the other platform which this library supports

articles/quickstart/native/xamarin-ios/01-login.md
jerriep: Once again, I am returning the Login result exactly as-is from the underlying library. The `User` object is `ClaimsPrincipal` https://msdn.microsoft.com/en-us/library/system.security.claims.claimsprincipal(v=vs.110) which is a standard .NET class.

Returning this and letting the user access the claims they need is consistent with the other .NET platforms such as ASP.NET

articles/quickstart/native/xamarin-ios/01-login.md
jerriep: 👍

jerriep:
@cocojoe 

> I feel some areas read more like a proof of concept than a QS guide

Can you please expand? Or is this what you addressed in the other comments?

articles/quickstart/native/xamarin-ios/01-login.md
cocojoe: Thought it would be a C# thing, just wondered.

cocojoe:
@jerriep It was related to comments, so this is fine.

cocojoe (approved):

jerriep:
Merged both Xamarin quickstarts into a single one. See #4074

PR #4052, opened 2017-05-04
Symfony api: update quickstart

glena:
Please confirm `api_identifier: ${account.audience}` is correct.


chenkie:
I had to adjust some other things to bring the quickstart in line with the other API ones. I took your updates from here and pushed them in another PR. Thanks for the PR :)

PR #4105, opened 2017-05-10
Link to scopes doc

khorne3:


khorne3:
@mpaktiti This is for https://trello.com/c/99QasNlA/954-custom-claims-namespacing. I think I did everything requested, but please let me know if that's not the case.

PR #4113, opened 2017-05-11
QuickStart - iOS Swift - Centralized Login & Embedded Login

cocojoe:
Restructure Getting Started
Removed redundant includes
Renamed Login -> Embedded Login
Separate Dependencies

Sample PR: https://github.com/auth0-samples/auth0-ios-swift-v2-sample/pull/11


cocojoe:
Need a first pass on this to check this is inline with what we want to achieve. Have not updated samples yet.  Will then update ObjC inline with this.

articles/quickstart/native/ios-swift/_includes/_login_centralized.md
```diff
@@ -0,0 +1,56 @@
+The first step in adding authentication to your iOS application is to provide a way for your users to log in. The fastest, most secure, and most feature-rich way to do this with Auth0 is to use the [centralized login page](https://auth0.com/docs/hosted-pages/login).
```
chenkie, line 1: Can we add something to "wrap up" this step of the quickstart? For example, show an image of the hosted Lock in a phone frame. There's a phone frame in the styleguide that can be used, just need to apply the class to the image:

```
<div class="phone-mockup">
  <img src="..." alt="Mobile example screenshot" />
</div>
```

articles/quickstart/native/ios-swift/_includes/_login_centralized.md
```diff
@@ -0,0 +1,56 @@
...
+            print("Error: \(error)")
+        case .success(let credentials):
+            // Do something with credentials e.g.: save them.
+            // Auth0 will automatically dismiss the hosted login page
+            print("Credentials: \(credentials)")
+        }
+}
+```
+
+As you can see, upon successful authentication the user's `credentials` will be returned.
```
chenkie: Drop the "As you can see" at the beginning.

cocojoe:
As this never got merged yesterday, I'll start adding the p2 parts now.

PR #4126, opened 2017-05-12
Lock iOS customization

jeffreylees:

### DIFF
--- a/articles/libraries/lock-ios/v2/customization.md
+++ b/articles/libraries/lock-ios/v2/customization.md
@@ -305,7 +305,7 @@ Image used as the Lock background
 
 ```swift
 .withStyle {
-  $0.backgroundColor = LazyImage?
+  $0.backgroundImage = LazyImage(name: "company_logo")
 }
 ```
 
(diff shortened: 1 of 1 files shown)


articles/libraries/lock-ios/v2/customization.md
```diff
@@ -155,6 +299,16 @@ Color used as the Lock background color.
 }
 ```
 
+### backgroundImage
+
+Image used as the Lock background
+
+```swift
+.withStyle {
+  $0.backgroundColor = LazyImage?
```
cocojoe: What is the source for this, wondering if I documented it incorrectly somewhere? As this should be like the icons e.g.

```
LazyImage(name: "company_logo")
```

jeffreylees: Oh, and that property name is wrong too :grin:.

I'll look back and see where it came from. Might be from that Styles.swift and I just didn't look closely enough.

jeffreylees:
@mpaktiti merge conflicts resolved

PR #4135, opened 2017-05-15
Replace deprecated article UI components

arielger:
We are moving to a new set of UI components for the Docs articles:

The main idea is to:
- Only keep two types of alerts and panels (default and warning) -> Remove other types
- Replace all `blockquotes` html elements by notes (default alerts)
- Replace `warning-banner` component by warning alert.
- Replace `notification` component by alerts.

Design:
image


mpaktiti (approved):
@arielger This includes changes on all files except for the quickstarts (it has some, but not all), for which I'll create a new PR. This one has too many files already.

PR #4145, opened 2017-05-16
Unique user fields noted

jeffreylees:


mpaktiti:
Hey @jeffreylees I reverted the note to the old format since the new UI elements are not merged yet, hence the format is broken. Good thinking though! I'll add a note to revert this back once we merge all new UI elements.

PR #4167, opened 2017-05-17
Use rules to restrict access to API scopes

khorne3:


articles/api-auth/api-access-policies.md
```diff
@@ -0,0 +1,39 @@
+---
+  description: Writing rules to restrict user/client access to an API
+---
+
+# Restrict Client or User Access to APIs
+
+By default, any user associated with an Auth0 client can request an API's scope(s). If you would like to restrict access to the API's scopes based on the user's role, you can do so via [rules](/rules). Then, if a restricted user attempts to request scopes not permitted to those with their role, they will receive an HTTP 401 response.
```
cabralmartin: ` If you would like to restrict access to the API's scopes based on the user's role,`

Not only user roles but any condition we can think of, i.e. we could restrict a certain client id or user id altogether, we could use location, etc

articles/api-auth/api-access-policies.md
```diff
@@ -0,0 +1,39 @@
...
+
+The following [rule](/rules) sample demonstrates how you would check for and permit/deny access depending on the `audience` parameter.
+
+```js
+function (user, context, callback) {
+
+  /*
+   * This script denies access to user-based flows using
+   * the value of the `audience` parameter, which is empty
+   * if you have a default set in the Management Dashboard.
```
cabralmartin: I'm not sure I understand this comment. Is this saying that `audience` will be empty if users define one in the Advanced Settings of the Management Dashboard?
I was under the impression that the `audience` param will be populated depending on wether or not it was sent as part of the query to `/authorize`.

articles/api-auth/api-access-policies.md
```diff
@@ -0,0 +1,39 @@
...
+
+  var audience = '';
+
+  audience = audience || (context.request && context.request.query && context.request.query.audience);
+
+  if (audience === 'http://todoapi2.api' || !audience) {
+    return callback(new UnauthorizedError('end_users_not_allowed'));
+  }
+
+  audience = audience || (context.request && context.request.body && context.request.body.audience);
```
cabralmartin: Silly question, would it be possible to merge these checks? Looks like dup code.

articles/api-auth/api-access-policies.md
```diff
@@ -0,0 +1,39 @@
...
+  if (audience === 'http://todoapi2.api'  || !audience) {
+    return callback(new UnauthorizedError('end_users_not_allowed'));
+  }
+
+  return callback(null, user, context);
+}
+```
+
+You can easily modify the script above to check the `scopes`, `clients`, or other value.
```
cabralmartin: Can we add an example for `client` showing how to 401 a specific client id that whats to consume a specific API?

cabralmartin (changes requested):

khorne3:
@cabralmartin Can you please (re)review? Thanks.

articles/api-auth/restrict-requests-for-scopes.md
```diff
@@ -0,0 +1,39 @@
...
+function (user, context, callback) {
+
+  /*
+   *  Denies access to user-based flows based on audience and
+   *  client ID.
+   */
+
+  // If you don't pass audience in the query string
+  // or body of the authorization request, the rule uses the
+  // default audience (leave the audience variable empty).
```
cabralmartin, line 22: Have we tried this rule with default audiences to check this is indeed the behaviour? From reading the code this gives me the impression that we are going to 401 to that client id in cases where no audience is defined, which sounds like a bug.

khorne3: You're correct; I'll fix it.

articles/api-auth/index.html
```diff
@@ -0,0 +1,183 @@
+---
```
mpaktiti: @khorne3 What is this file? Added by mistake?

articles/api-auth/restrict-requests-for-scopes.md
```diff
@@ -0,0 +1,56 @@
...
+```js
+function (user, context, callback) {
+
+  /*
+   *  Denies access to user-based flows based on client ID
+   */
+
+  var client_id = '';
+
+  client_id = context.clientID;
```
mpaktiti: @khorne3  Is there a point to this line? The `client_id` is not used anywhere in the script.

khorne3:
@mpaktiti Fixed. Thanks!

PR #4169, opened 2017-05-18
Use the auth0 object thats created, not the Classtype

seank90:
When setting setOIDCConformant it states to use an object called "auth0" when it should be the Auth0 object thats created called "account"


jeffreylees (approved):

mpaktiti:
Thank you @seank90 and @jeffreylees ! 👍

PR #4173, opened 2017-05-18
Use Auth0 to Secure a CLI

khorne3:


khorne3:
I removed the final example and just updated the doc to have better instructions. Let me know if you want me to craft a full sample.

darkyen (commented):
The POST to Token Endpoint method aren't needed anymore as setting the Client Type in dashboard to Native should do the same.

PR #4190, opened 2017-05-20
Plugin install command correction

fitzmode:
**First Change** 
The listed commands to install included both ionic and cordova in same statement. i.e. `ionic cordova plugin add ...`

Was
```bash
ionic cordova plugin add cordova-plugin-safariviewcontroller
```

Now

```bash
ionic plugin add cordova-plugin-safariviewcontroller
```


chenkie:
I've chatted with some folks about this and it sounds like the installation commands will differ between CLI versions. I think it's best if we keep it verbose so I'll close this but thanks for your PR anyway!

PR #4209, opened 2017-05-22
Add markdownlinter as a pre-commit hook

arielger:
Add [markdownlint](https://github.com/DavidAnson/markdownlint) as a pre-commit hook using husky and lint-stagged libraries.

Now when you try to commit a markdown file with style errors, you will get something like this:

image


package.json
```diff
@@ -5,14 +5,24 @@
...
+  },
+  "lint-staged": {
+    "*.md": "npm run lint-markdown"
   },
   "engines": {
     "node": "4.5"
   },
   "devDependencies": {
     "commander": "^2.8.1",
+    "filehound": "^1.16.1",
```
ramasilveyra: Why you need `filehound` ? It's a peerdep?

package.json
```diff
@@ -5,14 +5,24 @@
...
+    "husky": "^0.13.3",
     "imagemin": "^4.0.0",
-    "imagemin-optipng": "^4.3.0"
+    "imagemin-optipng": "^4.3.0",
+    "lint-staged": "^3.4.2",
+    "markdownlint": "^0.4.1",
+    "markdownlint-cli": "^0.3.1"
```
ramasilveyra, line 24: Maybe you don't need to list `markdownlint-cli` as devDep.

check-markdown.js
```diff
@@ -0,0 +1,54 @@
+/* eslint-disable  no-console */
+const markdownlint = require('markdownlint');
+
+const options = {
+  files: process.argv.slice(2),
```
ramasilveyra: I think that you will loose some changed files, to properly support running this script from `lint-staged` and from npm script or directly you need to do something like this: https://github.com/auth0/auth0-website/blob/master/tools/scripts/prettier#L36-L54

To debug you can modify `lint-staged` source to see the stdout here https://github.com/okonet/lint-staged/blob/c7283b77c2cf3637e9f06f953b919a556ffde205/src/runScript.js#L19-L20 with:
```js
// ...
.then(result => {
	console.log(result.stdout);
// ...
```

ramasilveyra:
With the last commit I believe that you can just use `markdownlint-cli`. 
> CLI argument --config is not mandatory. If it is not provided, markdownlint-cli looks for file .markdownlint.json in current folder, or for file .markdownlintrc in current or all upper folders. The algorithm is described in details on rc package page.

package.json
arielger: I will try that ;)

PR #4217, opened 2017-05-23
Removing implementation section and referencing library docs

jeffreylees:


mpaktiti:
@jeffreylees Can you please follow up with design (Juli) to update the diagrams at these two docs? Add a new trello card to track this. Thanks.

jeffreylees:
Will do

PR #4218, opened 2017-05-23
client_id -> clientID

seiyria:
just fixing a small thing I noticed while reading through the hosted login pages docs


mpaktiti:
Thank you @seiyria ! 👍 

@jeffreylees There are more instances of `client_id` in the doc, can you check and fix please? Thanks.

Issue #4002, opened 2017-04-27
WPF/WinForms tutorial no snippet location


vmartynets:
(AUTH-3918)
[WPF/Winforms tutorial](https://auth0.com/docs/quickstart/native/wpf-winforms#integration) snippets have no a file path/name is placed in the first line.
![image](https://cloud.githubusercontent.com/assets/1114365/25484128/664e648a-2b61-11e7-8311-55cdd10b2029.png)

PR #4254, opened 2017-05-25
Added Path/File name

mehreencs87:
Fixes https://github.com/auth0/docs/issues/4002


chenkie:
@mehreencs87 same issues here as #4256

PR #4255, opened 2017-05-25
Fix a typo. "attacker" is spelled wrong.

azazi-sa:
- Fix a typo in docs by renaming "attacher" to "attacker"


mpaktiti:
Thank you @azazi-sa ! 👍

PR #4277, opened 2017-05-26
Update HAR snippets for test endpoints

khorne3:


mpaktiti:
@khorne3 There is another snippet in this doc that uses `/oauth/ro`. This should probably be changed since pipeline2 had it replaced in favour of password grant. As another task, can you confirm this with crew-auth and plan for the update? Thanks.

PR #4282, opened 2017-05-27
how to generate a JWT

angelaLazar:
https://trello.com/c/KSPrgiww/1338-wordpress-jwt-authentication

### DIFF
--- a/articles/cms/wordpress/jwt-authentication.md
+++ b/articles/cms/wordpress/jwt-authentication.md
@@ -4,7 +4,7 @@ description: Explains how to install the WordPress JWT Authentication and integr
 
 # WordPress JWT Authentication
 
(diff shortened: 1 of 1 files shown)


mpaktiti (changes requested):
This doesn't look enough. To be honest I didn't understand the first thing about this plugin. When someone logs in Wordpress, using Auth0, this plugin sends the token to Auth0 to verify? 

I'm also wondering whether the user's question is addressed by your comment. It looks like you are suggesting that they manually generate JWTs which I doubt is the case. Aren't JWTs generated by the Authorization Server (thus, Auth0)?

Please work on this doc a bit more. It needs to be more clear.

angelaLazar:
@mpaktiti Unfortunately I not able to fully understand the wordpress plugin to update these docs because I do not have access to a self-hosted WordPress site to get the plugin (using a WordPress.com site does not allow installing plugins). 

I have tried in the past to get access to the Auth0 wordpress but it is used for site production and it's plugins cannot be altered. Let me know how you think we should proceed with this. If we want to set up a wordpress.org site then it requires having a domain to host it on.

PR #4302, opened 2017-05-30
Incorrect links for auth0-as-identity-and-service-provider

stevezau:
Fix incorrect incorrect/mixed links in auth0-as-identity-and-service-provider.


mpaktiti:
Good catch, thanks @stevezau ! 👍

PR #4345, opened 2017-06-01
Change views/user.js to routes/user.js in 04-user-profile (second code snippet)

laserpants:
I believe the correct file in the second code snippet is `routes/user.js`.


chenkie:
thanks!

PR #4364, opened 2017-06-05
Add user unblocked log info

dafortune:
Add information about the log when an anomaly detection block is released.


dafortune:
Review but do not merge, we are waiting update to be released by the pipeline

dafortune:
This is ready to review

PR #4375, opened 2017-06-06
Cross-Origin Auth Doc

jeffreylees:
Trying a new PR to fix Heroku name-taken issue


jeffreylees:
@chenkie Intro added. Maybe look that, and the rest of the doc over if you have time sometime? Do we want to keep the link to Jared's doc re: the cookies-disabled iframe workflow, or remove it, or somehow summarize that, or, what do you think?

jhiner:
FYI we also need to add the instructions to set the `cross_origin_loc: 'foo'` on the client.

More details about Cross_origin_loc can be found here: https://gist.github.com/jhiner/c57dc3ca9501e4b34154a4f23ab1809a

jeffreylees:
Any additions / thoughts / edits @chenkie ?

chenkie:
Can we get better screenshots for the cookie settings thing and the code snippet? I'm not sure why there is some yellow highlighting in the cookie settings screenshot.

chenkie:
The opening paragraph needs some adjustment. There are commas where there shouldn't be. Let's go with this:

For most situations, Auth0 recommends that authentication transaction be handled at the [Hosted Login Page](/hosted-pages/login) as it offers the easiest and most secure way to authenticate users. In some cases however, it's understood that there will be a need to embed the Lock widget or a custom login form directly in the application. In those cases, **cross-origin authentication** can be used.

Cross-origin authentication is done by creating security checks using 3rd party cookies. It can help prevent phishing when creating an SSO experience using embedded login forms on multiple domains. It also helps to create a secure login experience and way to acquire tokens on those embedded login forms, even if SSO is not the goal.

woloski:
rollout plan:

* we add the setting on the dashboard (crew2)
* we write the doc explaining how this feature works and what are pros/cons (jeff + ryan + aaguiar + justin)
* existing samples that use this capability should have the instructions on how to enable this capability using the dashboard ryan aaguiar
* quickstarts that explain Login have at the end “How to embed the login on your app” that explains the pros/cons of doing it plus how to enable it using the dashboard ryan aaguiar (this point to the doc above)
* customers who are asking to use this today, we send them the instruction to use the API and tell them we are adding this to the dashboard (dse - justin make sure they know about this)
* This become part of the Legacy Deprecation communication that eugene is working on. We are deprecating the current Lock/auth0js behaviors with this

chenkie:
Closing in favor of #4611

PR #4410, opened 2017-06-12
Correct Verifier and Challenge Base64 String

jedvalmar:
Verifier and Challenge codes should also trim equal sign when creating a base64 string


cocojoe:
I can see yes this part of the spec https://tools.ietf.org/html/rfc7636#section-3 However best if @hzalaz can check this

hzalaz (approved):

PR #4470, opened 2017-06-17
Add doc on Client Settings

khorne3:
Index: https://auth0-docs-content-pr-4470.herokuapp.com/docs/clients

Client Settings: https://auth0-docs-content-pr-4470.herokuapp.com/docs/clients/client-types-and-settings


articles/clients/client-types-and-settings.md
```diff
@@ -0,0 +1,51 @@
+---
+description: Describes the differences between the Auth0 Client types and what they mean when it comes to settings
+---
+
+# Settings for Client Types
```
mpaktiti: I would just name this `Client Settings`, the way it is now it's like this `Client Types` is some kind of object.

articles/clients/client-types-and-settings.md
```diff
@@ -0,0 +1,51 @@
...
+* **Native**: used for desktop or mobile apps that run natively on the the device
+* **Single Page Web Applications**: apps with a JavaScript front-end utilizing an API
+* **Regular Web Applications**: traditional web apps with refresh behavior
+* **Non Interactive Clients**: CLIs, daemons, or services running on the back-end
+
+## Settings
+
+By default, the settings for all client types include the following:
+
+* Name: the name of your Client;
```
mpaktiti: I would mark as bold each setting: **Name:** instead or Name:

articles/clients/client-types-and-settings.md
```diff
@@ -0,0 +1,51 @@
...
+* Allowed Logout URLs: the URLs to which a user can be redirected to after logout;
+* Allowed Origins (CORS): URLs allowed to make JavaScript requests to the Auth0 API
+* JWT Expiration (seconds): the length of time before the Auth0-issued ID token expires;
+* Use Auth0 instead of the IdP to do Single Sign On: if enabled, Auth0 handles SSO instead of the IdP
+
+![Client Settings Page](/media/articles/clients/settings.png)
+
+Clients of type **Non Interactive Clients** or **Regular Web Page Applications** have an additional setting called *Token Endpoint Authentication Method*. This defines the token endpoint's requested authentication method, and the accepted values are:
+
+* **None**, for a public client without a client secret
```
mpaktiti: Use `this formatting` for values, and bold for field/view names

articles/clients/client-types-and-settings.md
```diff
@@ -0,0 +1,51 @@
...
+* **Post**, for a client using HTTP POST parameters
+* **Basic**, for a client using HTTP Basic parameters 
+
+### Advanced Settings
+
+The *Advanced Settings* section allows you to:
```
mpaktiti: Avoid italics (tip I got from design), use bold instead

articles/clients/client-types-and-settings.md
mpaktiti: I would also rename the file to `client-settings.md` to simplify

mpaktiti (changes requested):
Left some formatting comments.

I was also thinking that maybe we could use a dropdown for this doc: the dropdown at the top would list the client type, and only the relevant settings would be displayed. Of course you would use document fragments for the text they have in common. Something like this https://auth0.com/docs/connections/passwordless/ios-sms-objc What do you think?

khorne3:
@mpaktiti Good idea on using a dropdown for this doc. I think we should publish it as is for now, and I've created a card for changing its format.

PR #4471, opened 2017-06-17
Add link to reset password doc.

khorne3:


mpaktiti:
@khorne3 For future reference, nothing should be above H1, all notes/panels/warnings should be after that.

PR #4492, opened 2017-06-20
Update instructions on setting grant type and add screenshots

khorne3:


jhiner:
The part at the bottom:
```
To enable a legacy grant type, you will need to make the appropriate PATCH call to the Update a Client endpoint of the Management API.
```
Can you also please note that you can do this in Dashboard, not the mgmt API. Thanks.

jhiner:
Also, the bit about updated token endpoint auth method:

https://auth0.com/docs/clients/client-grant-types#public-clients

```
You can update the token_endpoint_auth_method flag using the Management API's Update a Client endpoint.
```

It's in the live doc but I don't see it in this PR. This should be changed to mention the Dashboard instead.

jhiner:
`Trusted first-party clients can additionally use the following grant_types:`

Please remove `refresh_token` from this list, since it is allowed for 3rd party clients.

PR #4494, opened 2017-06-21
Declaring Variables

vicsant:

### DIFF
--- a/articles/quickstart/backend/java-spring-security/01-authorization.md
+++ b/articles/quickstart/backend/java-spring-security/01-authorization.md
@@ -50,7 +50,7 @@ public class SecurityConfig extends WebSecurityConfigurerAdapter {
     @Override
     protected void configure(HttpSecurity http) throws Exception {
         JwtWebSecurityConfigurer
-                .forRS256(apiAudience, issuer)
+                .forRS256(https://${account.namespace}/userinfo, ${account.clientId})
                 .configure(http)
                 .authorizeRequests()
                 .antMatchers(HttpMethod.GET, "/login").permitAll()


lbalmaceda (changes requested):
`apiAudience` and `issuer` are variables read by the `AppConfig` class from the [application.properties](https://github.com/auth0-samples/auth0-spring-security-api-sample/blob/master/01-Authorization/src/main/resources/auth0.properties.example) file. I don't know if the audience is auto-filled by our docs-tool, but the domain should be.

```java
    @Value(value = "${auth0.apiAudience}")
    private String apiAudience;
    @Value(value = "${auth0.issuer}")
    private String issuer;
```

The snippets you see in the doc refer to that same code from the example. That's how @chenkie wanted it to be: "using snippets exactly as shown in the samples".

If you choose to change them then the changes introduced in this PR are wrong:
1. There's a similar line a few lines above this snippet configuring the `JwtWebSecurityConfigurer` with the same parameters that it's not being changed here, so the change should apply there too.
2. The audience is particular for each user-defined API, and it's clearly NOT the /userinfo endpoint. Again, I'm not sure that the docs-tool will replace this value. If it's not clear then I would rewrite the setup part of the article.
3. hard-coded string values must be between `"`. So for example, if you plan to pass the "namespace" and "clientId" it would look like:

```java
        .forRS256("https://${account.namespace}/", "${account.clientId}")
```

If you have doubts about the snippets, I strongly suggest you download the simple sample from the repo and try to run it. If it fails then the snippet is wrong.

chenkie:
fixed in another pr

PR #4504, opened 2017-06-21
Update forum link

Annyv2:


mpaktiti:
@Annyv2 I added  a var for this and replaced the hardcoded values, next time this changes we'll only need to change the var

PR #4507, opened 2017-06-21
Bump the composer version for auth0/login

aaronranard:
Installing 4.0 with Laravel Spark failed as it used firebase/php-jwt v3.0.0 but Spark requires 4.0+

This is fixed by using "~5.0"


chenkie:
Thanks!

PR #4520, opened 2017-06-23
Merge getting started into login chapter

chenkie:


jerriep (approved):
I went through all of them and did not see anything broken. I also did not notice any problems looking at the changed files.

All seems good.

PR #4541, opened 2017-06-27
Update deprecation notices in Auth API Explorer

khorne3:
https://auth0-docs-content-pr-4541.herokuapp.com/docs/api/authentication#database-ad-ldap-active-

https://auth0-docs-content-pr-4541.herokuapp.com/docs/api/authentication#get-token-info

https://auth0-docs-content-pr-4541.herokuapp.com/docs/api/authentication#resource-owner

### DIFF
--- a/articles/api/authentication/legacy/_login.md
+++ b/articles/api/authentication/legacy/_login.md
@@ -168,7 +168,7 @@ curl --request POST \
 }) %>
 
 ::: warning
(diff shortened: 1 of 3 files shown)


mpaktiti:
@khorne3 I don't think we should change that, sorry I didn't decide on this before you put any effort on that. I think the messages are more confusing like that. If they don't like the description we should update it across the board, including the adoption guide, not only this panel. Let's leave it as is for now.

PR #4553, opened 2017-06-28
Update renewAuth examples to reflect new behavior

twistedstream:
You no longer have to call `parseHash` inside the silent authentication callback page. Instead you just postMessage back the local URL hash and `renewAuth` performs `parseHash` automatically.
Also fixing a related issue with using `nonce` and `renewAuth`.


mpaktiti:
cc @jeffreylees

PR #4577, opened 2017-07-01
Fix broken ${manage_url} links

nightowlengineer:
Noticed a broken link when reading https://auth0.com/docs/api/management/v2/tokens, fixed the same issue in a couple of other locations


mpaktiti:
Thank you fixing that @lake54 ! 💯

nightowlengineer:
No problem @mpaktiti, thanks for merging 👍

PR #4579, opened 2017-07-01
Fixed a typo (you referred to 'class' instead of 'id').

stheine:
Fixed a typo (you referred to 'class' instead of 'id').

I guess the change also needs to be applied to articles/quickstart/spa/jquery/_includes/_centralized_login.md


chenkie:
thanks!

PR #4611, opened 2017-07-05
add cross origin doc

chenkie:


jhiner:
We need to add a warning here: Only works with database users. AD/LDAP, WAAD, and ADFS connections do not work with CO/auth at this time, but will be added in the future.

PR #4615, opened 2017-07-06
Lock add oidc

chenkie:


jmangelo:
@chenkie should we not be documenting the `audience` option in Lock at the same time we document `oidcConformant`? 

I was assuming that with cross origin support we would get OIDC and API Authorization working in Lock (when embedded in the client application itself).

PR #4619, opened 2017-07-06
Remove unsafe practice from example

yungchin:
Relying on information from the unverified header during verification is an unsafe practice, as described here:
https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/


chenkie:
thanks!

PR #4623, opened 2017-07-06
User Profile - OIDC-conformant auth

khorne3:
https://auth0-docs-content-pr-4623.herokuapp.com/docs/user-profile/normalized


khorne3:
Spent two hours fixing broken versioning yesterday (which only resulted in me breaking my instance of npm). Gave up, came back to it today, got it right on the first try. 🤦‍♂️

articles/user-profile/normalized/auth0/index.md
```diff
@@ -2,6 +2,10 @@
 description: Auth0 normalizes common user properties in the User Profile.
 ---
 
+::: version-warning
+The user profile returned after successful authentication using an OIDC-compliant flow may differ from the Auth0 normalized profile. Please use the toggle to see the claims available on that profile.
+:::
+
```
mpaktiti: Move the panel under the H1. I would also like to see a different text. You say what the other page is but you don't say what this page is. What is an OIDC-compliant flow? Try to make it more clear and add a link for more info.

I would expect something like this:
"This article describes the normalized user profile, which is a protocol-agnostic representation of this user as defined by Auth0....." Describe what this is and then say that if you are looking for the other thing use the toggle. You can check out this doc https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims

mpaktiti: Also, you need to update this paragraph https://github.com/auth0/docs/pull/4623/files#diff-0aeba265d924efd23d617f971e37b079R196 : 
- The Authentication API /userinfo endpoint, returns the OIDC user profile, not the normalized
- The User Profile section of our quickstarts, uses methods that return the OIDC user profile, not the normalized
- The way to retrieve the normalized user profile is using the Management API https://auth0.com/docs/api/management/v2#!/Users/get_users_by_id
- Don't know if you can retrieve the normalized user profile using Lock or auth0.js (Call the management API endpoints using these libraries? Is there another way? Cross check)

articles/user-profile/normalized/oidc/index.md
```diff
@@ -0,0 +1,63 @@
+---
+description: Claims for a user profile returned via an OIDC-Compliant authorization flow
+---
+
+# User Profiles Returned from OIDC-Compliant Pipelines
```
mpaktiti: This version needs a version-warning that explains the doc as well.

mpaktiti: How can I get this user profile? We need to mention this. Maybe add a  `How to retrieve the User Profile` paragraph, like the other doc has? See here https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims#standard-claims

articles/user-profile/normalized/oidc/index.md
```diff
@@ -0,0 +1,63 @@
...
+
+When you're using an OIDC-conformant authentication flow, the user profile you receive in return may differ slightly from the [Auth0 Normalized User Profile](/user-profile/normalized).
+
+The following is a [non-normative example of such a response](https://openid.net/specs/openid-connect-basic-1_0.html#StandardClaims):
+
+```json
```
mpaktiti, line 14: Shouldn't the sample JSON include all fields that we list below?

mpaktiti (changes requested):
Left some comments, let me know if not clear

articles/user-profile/normalized/auth0/index.md
khorne3: Re: Lock and auth0js - you currently can get the normalized profile (looks like the functionality exists because of specific client use case needs), but it's being deprecated in the near future, so I'm not going to mention it.

khorne3:
@mpaktiti Merge conflicts should be fixed. Thanks!

khorne3:
@mpaktiti Do you need anything else from me on this?

mpaktiti:
guess what Katie... 🥁

PR #4628, opened 2017-07-07
Fixed error where "Property 'expectedScopes' does not exist on type '…

Andrew-Casner:
…{ [name: string]: any; }'."

I kept running into the error that "Property 'expectedScopes' does not exist on type '{ [name: string]: any; }'." when building my angular application. By using as any you can avoid this typescript error.


chenkie:
thanks!

PR #4647, opened 2017-07-11
Update Logstash extension config parameters

khorne3:
https://auth0-docs-content-pr-4647.herokuapp.com/docs/extensions/logstash

### DIFF
--- a/articles/extensions/logstash.md
+++ b/articles/extensions/logstash.md
@@ -18,6 +18,12 @@ At this point you should set the following configuration variables:
 - **BATCH_SIZE**: The ammount of logs to be read on each execution. Maximun is 100.
 - **LOGSTASH_URL**: Your Logstash URL as defined for use with `logstash-input-http` plugin.
 - **LOGSTASH_INDEX**: Your Logstash Index to which the logs will be routed.
+- **LOGSTASH_TOKEN** (optional): The token required for your Logstash deployments that will be included in the querystring
+- **LOGSTASH_USER** (optional): The Logstash user
+- **LOGSTASH_PASSWORD** (optional): The password associated with your Logstash user
+- **START_FROM**: The checkpoint ID of the log from where you want to start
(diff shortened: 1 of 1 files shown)


articles/extensions/logstash.md
```diff
@@ -4,33 +4,43 @@ description: How to install and configure the Auth0 Logs to Logstash extension.
...
+| Parameter        | Description |
+|:-----------------|:------------|
+| **Schedule** | The frequency with which logs should be exported. The schedule can be customized even further after creation. |
+| **BATCH_SIZE** | The ammount of logs to be read on each execution. Maximun, and default, is `100`. |
+| **LOGSTASH_URL** <br/><span class="label label-danger">Required</span> | Your Logstash URL as defined for use with `logstash-input-http` plugin. |
+| **LOGSTASH_INDEX** <br/><span class="label label-danger">Required</span> | Your Logstash Index to which the logs will be routed. |
+| **LOGSTASH_TOKEN** | The token required for your Logstash deployments that will be included in the querystring. |
+| **LOGSTASH_USER** | The Logstash user. |
+| **LOGSTASH_PASSWORD** | The password associated with your Logstash user. |
+| **START_FROM** | The checkpoint ID of the log from where you want to start. |
```
mpaktiti, line 26: @khorne3 Hey I changed the format from bullets to table, I think it looks better that way, what do you think?

Also, about `START_FROM` I cannot see it in my dashboard, is this released?

khorne3: @mpaktiti Table is great; thank you!

Do you have an existing Logstash extension already installed? I didn't, so I had to install a fresh copy?

mpaktiti: Interesting, I didn't have it at the time but I had it in the past. When I re-installed it I didn't see this field. I tried with another account which never had this extension, and I can see it.

PR #4657, opened 2017-07-12
Correct typo in spa-api.md

nodediggity:
This PR is intended to fix a small typo in the docs around SPA + API authentication.
At present access_token is spelt incorrectly on line 82. 

This became apparent when searching the docs for 'access_token' and not finding this line, which was in fact the line I required.


mpaktiti:
Thank you @nombienombie ! 👍

PR #4666, opened 2017-07-13
Fixed Rotate Secret API documentation link

SKalt:
the prior link, /api/v2/clients/{id}/rotate-secret was broken: {id} was urlencoded rather than replaced, and /api/v2/clients/rotate-secret did not point to the docs.


mpaktiti:
Good catch, thank you @SKalt ! 👍

PR #4677, opened 2017-07-14
Auth0.js migration guide additions 

jeffreylees:

### DIFF
--- a/articles/libraries/auth0js/v8/migration-guide.md
+++ b/articles/libraries/auth0js/v8/migration-guide.md
@@ -16,7 +16,7 @@ Take a look below for more information about changes and additions to Auth0.js i
 
 The first question to answer before getting into the changes is why to migrate your app to the new version at all. Here are a few quick points that address that:
 
-* With version 8 of the Auth0.js SDK, you can use our latest and most secure authentication pipeline, compliant with the OpenID Connect specification. For more information, refer to the below section [Use of API Auth and Metadata](#use-of-api-auth-and-metadata).
+* With version 8 of the Auth0.js SDK, you can use [our latest and most secure authentication pipeline](/api-auth/intro), compliant with the OpenID Connect specification. For more information, refer to the below section [Use of API Auth and Metadata](#use-of-api-auth-and-metadata).
 * Auth0.js v8 was rewritten from scratch, improving its cohesion and performance and coming with more tests to be utilized.
 * Long term support - As is often the case with new iterations of projects, v8 will be supported for significantly longer than v7.
 
@@ -33,8 +33,8 @@ However, if your application is currently relying on being able to request metad
 
 Alternatively, you could also simply request the metadata in a different way, for example with a rule to add custom claims to either the returned `id_token` or `access_token` as described in the [custom claims](/scopes/current#custom-claims) section of the scopes documentation.
 
-::: panel-note Function Reference
-Note that you can find detailed information about supported methods in the [Auth0.js v8](/libraries/auth0js) documentation, and generated documentation on all methods [here](http://auth0.github.io/auth0.js/global.html) for further reading.
+::: note
+You can find detailed information about supported methods in the [Auth0.js v8](/libraries/auth0js) documentation, and generated documentation on all methods [here](http://auth0.github.io/auth0.js/global.html) for further reading.
 :::
 
 ## Initialization of auth0.js
--- a/articles/libraries/auth0js/v8/migration-guide.md
+++ b/articles/libraries/auth0js/v8/migration-guide.md
@@ -4,16 +4,15 @@ title: Auth0.js v7 to v8 Migration Guide
 description: How to migrate from auth0.js v7 to auth0.js v8
 toc: true
 ---
-
 # Auth0.js v7 to v8 Migration Guide
 
-The following instructions assume you are migrating from **auth0.js v7** to **auth0.js v8**. 
+The following instructions assume you are migrating from **auth0.js v7** to **auth0.js v8**.
 
-The goal of this migration guide is to provide you with all of the information you would need to update Auth0.js in your application. Of course, your first step is to include the latest version of auth0.js. Beyond that, take a careful look at each of the areas on this page. You will need to change your implementation of auth0.js to reflect the new changes. 
+The goal of this migration guide is to provide you with all of the information you would need to update Auth0.js in your application. Of course, your first step is to include the latest version of auth0.js. Beyond that, take a careful look at each of the areas on this page. You will need to change your implementation of auth0.js to reflect the new changes.
 
 Take a look below for more information about changes and additions to auth0.js in version 8!
 
-## Why Migrate?
+## Reasons to Migrate
 
 The first question to answer before getting into the changes is why to migrate your app to the new version at all. Here are a few quick points that address that:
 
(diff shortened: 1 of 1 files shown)


articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -4,26 +4,37 @@ title: Auth0.js v7 to v8 Migration Guide
...
 
 The first question to answer before getting into the changes is why to migrate your app to the new version at all. Here are a few quick points that address that:
 
 * v8 is tailored for usage with the Authentication API and for OIDC compliance.
 * v8 was rewritten from scratch, improving its cohesion and performance and coming with more tests to be utilized.
 * Long term support - As is often the case with new iterations of projects, it is likely that v8 will be supported for quite a bit longer than v7.
 
+### Use of API Auth and Metadata
+
+If you need to use [API Auth](/docs/api-auth) features, we recommend that you upgrade to [auth0.js v8](/docs/libraries/auth0js/v8).
```
mpaktiti: Remove `/docs` from links: `/docs/api-auth` -> `/api-auth`

(there are more later in the doc)

mpaktiti: Also I'd like to see some more context. `API Auth` doesn't really mean something. I get that you link to the doc, but you should add a sentence or two and then say for more information refer to API Authorization.

articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -4,26 +4,37 @@ title: Auth0.js v7 to v8 Migration Guide
...
-## Why Migrate?
+## Reasons to Migrate
 
 The first question to answer before getting into the changes is why to migrate your app to the new version at all. Here are a few quick points that address that:
 
 * v8 is tailored for usage with the Authentication API and for OIDC compliance.
 * v8 was rewritten from scratch, improving its cohesion and performance and coming with more tests to be utilized.
 * Long term support - As is often the case with new iterations of projects, it is likely that v8 will be supported for quite a bit longer than v7.
 
+### Use of API Auth and Metadata
```
mpaktiti: I would add something about this paragraph in the first bullet earlier:

`Auth0.js v8 is tailored for usage with the Authentication API and for OIDC compliance. For more information, refer to [Use of API Auth and Metadata](#use-of-api-auth-and-metadata)`

Also, I think the this first sentence (`tailored for usage with the Authentication API and for OIDC compliance`) is kind of obscure. Maybe use more plain language and give more info? Sth like: with version 8 of the library, you can use our latest and more secure authentication pipeline, that complies with the OpenID Connect specification.

articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -4,26 +4,37 @@ title: Auth0.js v7 to v8 Migration Guide
...
+
+If you need to use [API Auth](/docs/api-auth) features, we recommend that you upgrade to [auth0.js v8](/docs/libraries/auth0js/v8).
+
+If, however, your application is relying on being able to request metadata via [scopes](/docs/scopes), you have two choices:
```
mpaktiti: I don't understand what `If, however, your application is relying on being able to request metadata via [scopes](/docs/scopes)` means. Also, the doc you linked, is versioned and the version this navigates is the OIDC-compliant (plus, it's broken due to the extra `/docs`.

articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -4,26 +4,37 @@ title: Auth0.js v7 to v8 Migration Guide
...
+
+* Continue to use v7 until it is no longer an option
+* Use v8 without using API Auth. Do not mark your client as OIDC Conformant, do not pass the `audience` parameter. For example, if using [authorize()](http://auth0.github.io/auth0.js/global.html#authorize), [loginWithResourceOwner()](http://auth0.github.io/auth0.js/global.html#loginWithResourceOwner), [loginWithCredentials()](http://auth0.github.io/auth0.js/global.html#loginWithCredentials), or [signupAndLogin()](http://auth0.github.io/auth0.js/global.html#signupAndLogin) (_Note that some of the methods mentioned here use the legacy Resource Owner grant in order to be used and are unavailable to some customers._).
```
mpaktiti: - Same comment as earlier `API Auth` means nothing to most users.
- Remove italics from the last note.
- Why say `some of the methods mentioned here`? It's 4 methods in total, why not say which use the legacy RO grant? Also, most people will not know what this legacy RO grant is (or a grant in general). Add a link maybe?
- You say `For example, if using x, y, z, or w`. Syntactically it's not correct, you're missing the second part of the if clause. I would rephrase this entire bullet. 
- Why do you link outside this doc? Is this new http://auth0.github.io/auth0.js/global.html ? It's not that I don't like it (on the contrary, I always thought these docs should not be just a library reference) but why link outside our docs, if you have this info here? Also why not add a note panel at the beginning of this doc to link maybe to this external reference?

articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -4,26 +4,37 @@ title: Auth0.js v7 to v8 Migration Guide
...
+
+Alternatively, you could also simply request the metadata in a different way, for example with a rule to add custom claims to either the returned `id_token` or `access_token` as described in the [custom claims](https://auth0.com/docs/scopes/current#custom-claims) section of the scopes documentation.
```
mpaktiti: Use relative links.

articles/libraries/auth0js/v8/migration-guide.md
```diff
@@ -169,14 +180,13 @@ To switch from HS256 to RS256 for a specific client, follow these instructions:
...
-### Manually Parsing Hashes 
+### Manually Parsing Hashes
 
 If you would rather manually parse hashes, to avoid the `parseHash` method since it only works with RS256, feel free to take a look at [this example](https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js#L97) or [this one](https://github.com/auth0/auth0.js/blob/master/src/helper/qs.js#L10) to help you get started.
 
 ## Refreshing Tokens
 
-In [auth0.js v7](/libraries/auth0js/v7#refresh-token), the `renewIdToken` and `refreshToken` methods were used to renew tokens. In [auth0.js v8](/libraries/auth0js#using-renewauth-to-acquire-new-tokens), 
-refreshing tokens is now done via the `renewAuth` method. If a user is already authenticated, `renewAuth` can acquire a new token for that user.
+In [auth0.js v7](/libraries/auth0js/v7#refresh-token), the `renewIdToken` and `refreshToken` methods were used to renew tokens. In [auth0.js v8](/libraries/auth0js#using-renewauth-to-acquire-new-tokens), refreshing tokens is now done via the `renewAuth` method. If a user is already authenticated, `renewAuth` can acquire a new token for that user.
```
mpaktiti: - Remove `now` from `is now done`. 
- At the 1st sentence you say `renew` and at the 2nd `refresh`. I get that the meaning is the same but it's better to be consistent. 
- I'd add a short sentence at the beginning of this paragraph saying what "refreshing tokens" is

mpaktiti (changes requested):
Hey Jeff, left some comments.

jeffreylees:
@mpaktiti thank you for the feedback - ugh, the links, must have been asleep at the wheel 🙊 

I've tried to re-word some things to make them more understandable. Also removed all the excess (probably unnecessary) specific method links, and instead added a panel noting where they can get more information (via our docs, or the generated ones). 

I'll take another pass over this later today, maybe, as well.

mpaktiti:
Looks good Jeff, thanks 👍  I fixed the panel (it's either `panel` or `note`) and added one more link towards the api-auth intro doc.

PR #4680, opened 2017-07-14
Fix typos in Scopes documentation

wags:
I spotted a couple typos when reading the `Scopes` documentation. This PR fixes the ones that I noticed.


chenkie:
thanks for your fix!

PR #4684, opened 2017-07-17
Describe how to update fields in a custom DB

jerriep:
https://auth0-docs-content-pr-4684.herokuapp.com/docs/user-profile/customdb


mpaktiti:
@jerriep Looks good, my only concern is the title, it's a bit too long. How about `Update Users using a Custom Database`? This is closer to the other doc `Authenticate Users with Username and Password using a Custom Database` (which you can change if you like to simple `Authenticate Users using a Custom Database`, I think the user/pass clarification is redundant).

jerriep:
@mpaktiti I updated as per your suggestions

PR #4692, opened 2017-07-18
Document allowShowPassword option

jeffreylees:


articles/libraries/lock/v10/customization.md
```diff
@@ -123,6 +124,16 @@ Examples of `allowedConnections`:
 
 ![Lock - Allowed Connections](/media/articles/libraries/lock/v10/customization/lock-allowedconnections-social.png)
 
+### allowShowPassword {Boolean}
```
mpaktiti, line 126: @jeffreylees In some other methods you also have screenshots of how the option looks, for example https://auth0.com/docs/libraries/lock/v10/customization#language-string- Since this one adds a checkbox, shouldn't we have a screenshots here too?

jeffreylees: Sure, can do one

articles/libraries/lock/v10/customization.md
```diff
@@ -123,6 +124,16 @@ Examples of `allowedConnections`:
 
 ![Lock - Allowed Connections](/media/articles/libraries/lock/v10/customization/lock-allowedconnections-social.png)
 
+### allowShowPassword {Boolean}
+
+This option determines whether or not add a checkbox to the UI which, when selected, will allow the user to show their password when typing it. The option defaults to `false`.
```
mpaktiti: @jeffreylees You are missing a `to` here: `This option determines whether or not to add`

PR #4693, opened 2017-07-19
Update typo in _centralized_login.md

wasong:
fix typo from *responsbile* to *responsible*


mpaktiti:
Good catch, thank you @wasong ! 👍

PR #4696, opened 2017-07-19
Add links to quickstarts on logout doc

jerriep:
https://auth0-docs-content-pr-4696.herokuapp.com/docs/logout#implementing-in-your-application

### DIFF
--- a/articles/logout/index.md
+++ b/articles/logout/index.md
@@ -216,3 +216,45 @@ To prevent a Session Participant from being notified, you can set `logout.slo_en
 #### Non Single Logout Scenario
 
 If your Service Provider does not support SAML SLO, but provides a redirect URL where the user will be redirected to after logging out of the SP, the best thing to do is configure the redirect URL to `https://${account.namespace}/logout`. This won't notify other session participants that a logout was initiated, but it will at remove the session from Auth0.
+
+## Implementing in your Application
+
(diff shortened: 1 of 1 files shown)


mpaktiti:
@jerriep I'm thinking whether it's a good idea to list all the quickstarts individually. The steps numbering changes quite often and I'm afraid we'll keep ending up with broken links.

What if we used generic links to quickstarts and the next-steps element instead? Something like:

```
## Implementing in your Application

For guidance and sample code on how to implement logout functionality in your application please refer to our quickstarts:

::: next-steps
- [Quickstarts for mobile and native apps: Android, iOS and more](/quickstart/native)
- [Quickstarts for Single Page apps: Angular, React and more)](/quickstart/spa)
- [Quickstarts for regular web apps: ASP .NET, Node.js and more](/quickstart/webapp)
:::
```

Wdyt?

jerriep:
@mpaktiti Yeah I can do that, but the only thing is that it is not on the sample place for all the quickstarts. For some of them, e.g. Android and iOS, it is located only under the "Session Handling" step, so it may not be immediately obvious to a user where to find the actual logout code

It you don't think that's too much of an issue then I can change to link just to the high level quickstart categories as you suggested.

mpaktiti:
@jerriep I see your point. Let's leave it as is for the moment.

PR #4704, opened 2017-07-20
Rename appliance

khorne3:


mpaktiti:
@khorne3 I don't see anywhere what `PSaaS` stands for, we only use the acronym. Are you sure you want me to merge this as is?

khorne3:
@mpaktiti You're right. I've updated all index pages with a definition (though I didn't define SaaS...do you think I should?).

mpaktiti:
@khorne3 Looks good Katie 👍  (no need to define SaaS)

PR #4711, opened 2017-07-21
Added C# sample to authorization-code-grant-pkce docs

ghost:
Added sample for creating the verifier (step 1) and code challenge (step 2) in C#


jerriep:
@mpaktiti I am not sure about this, as we have a proper SDK (https://github.com/auth0/auth0-oidc-client-net) which takes care of all this low level stuff for the users. I cannot see a C# user ever having to implement this themselves. I am afraid that adding code samples like this will add more confusion

What is your opinion?

mpaktiti:
@jerriep I agree, thanks for the review

@oxiqu We will not be using this but thank you for taking the time to send this PR 👍

PR #4723, opened 2017-07-23
fix links to sections of V2 management API explorer

rmcloughlin:
Currently many links just go to the introduction page rather than the appropriate section.


mpaktiti:
Good catch! Thanks for contributing @rmcloughlin ! 👍

PR #4727, opened 2017-07-25
Delete duplicate which word typo

serhatcan:
Minor typo fix


mpaktiti:
Good catch, thanks @srhtcn ! 👍

PR #4732, opened 2017-07-26
Incorrect Example Parameter for webAuth.parseHash

ibuchan72390:
In order for the hash parameter to be properly passed to the webAuth.parseHash method, it needs to be represented as an options object instead of a simple string pass.  Digging into the library seems to show that the original method never truly got passed as the hash value and the default window.location.hash was being used instead of the provided parameter.


chenkie:
thanks!

PR #4739, opened 2017-07-28
Updating to use latest changes to rule

jeffreylees:


mpaktiti (changes requested):
Αdd a panel saying that instead of hardcoding credentials in the rule, they can use the `configuration` object instead.

PR #4758, opened 2017-08-01
Fix for vertical bar not showing in subject example

borice:


mpaktiti:
Good catch, thank you @borice ! Since we try to avoid using html tags in our markdown, I'll go ahead and replace the html encoding of the character, so we can use the markdown syntax and still display the character properly. Thanks again for bringing this to our attention!

borice:
Hi.  Do whatever you think is right, but no fix other than what I made works.... when you use the UTF8 code in markdown, it doesn't get replaced inside that table element.  Only the fix I provided works.

And you already have HTML in your markdown anyway: "<br/><br/>"

mpaktiti:
@borice Yes we do have and we are trying to replace the existing, and not introduce new elements, if we have a markdown alternative.

The doc looks ok to me now https://auth0.com/docs/tokens/id-token#id-token-payload Thanks again for bringing this to our attention and taking the initiative to fix it! 👍

PR #4775, opened 2017-08-04
remove link no longer valid

sopelt:
state has been used incorrectly in the linked article


mpaktiti:
Good catch, thank you @sopelt ! 👍

PR #4777, opened 2017-08-04
Update API

khorne3:
https://auth0-docs-content-pr-4777.herokuapp.com/docs/api/authentication#get-user-info

### DIFF
--- a/articles/api/authentication/_userinfo.md
+++ b/articles/api/authentication/_userinfo.md
@@ -67,6 +67,20 @@ Given the Auth0 [access token](/tokens/access-token) obtained during login, this
 
 This endpoint will work only if `openid` was granted as a scope for the `access_token`.
 
+Response Notes:
+
+* `Email` claim returns a snapshot of the email at the time of login
+* Standard claims (other than `email` return the latest value (unless the value comes from an external IdP)
+* Custom claims return a snapshot of the value at the time of login
+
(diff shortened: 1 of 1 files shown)


articles/api/authentication/_userinfo.md
```diff
@@ -67,6 +67,20 @@ Given the Auth0 [access token](/tokens/access-token) obtained during login, this
 
 This endpoint will work only if `openid` was granted as a scope for the `access_token`.
 
+Response Notes:
```
mpaktiti: Hey Katie, I think we should move all this info as bullets under the `Remarks` paragraph. 

```
- The `Email` claim returns a snapshot of the email at the time of login
- Standard claims (other than `email`) return the latest value (unless the value comes from an external IdP)
- Custom claims return a snapshot of the value at the time of login
- To access the most up-to-date values for the `email` or custom claims, you must get new tokens. You can log in using silent authentication (where the `prompt` parameter for your call to the [`authorize` endpoint](/api/authentication#authorization-code-grant) equals `none`)
- To access the most up-to-date values for standard claims that were changed using an external IdP (for example, the user changed their email address in Facebook)., you must get new tokens. Log in again using the external IdP, but *not* with silent authentication.
```

If you think they should be grouped, add a new paragraph `Response Notes`

mpaktiti (changes requested):
Consider moving this info  under another section

PR #4794, opened 2017-08-09
Update password-reset.md to display list of custom variables

khorne3:
https://auth0-docs-content-pr-4794.herokuapp.com/docs/hosted-pages/password-reset

### DIFF
--- a/articles/hosted-pages/password-reset.md
+++ b/articles/hosted-pages/password-reset.md
@@ -77,7 +77,7 @@ Within the Password Reset Page Editor, you'll see the following JavaScript embed
   </script>
 ```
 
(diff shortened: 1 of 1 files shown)


articles/hosted-pages/password-reset.md
```diff
@@ -17,6 +17,23 @@ Using the [Auth0 Dashboard](${manage_url}/#/password_reset), you can enable your
 
 Once you've enabled the Password Reset Page for your account, you'll be able to use the text editor built into the Auth0 Dashboard to change your HTML, style your page using CSS, and alter the JavaScript used to retrieve custom variables. After you've made your changes, and make sure to click _Save_.
 
+### Custom Variables
```
mpaktiti, line 20: That looks good Katie. I'm wondering though, wouldn't it be better if we added one example? The script for a password reset page retrieving and displaying a couple of these variables?

mpaktiti: I'm also thinking that some of these vars are self-explanatory (`email`) but others not so much (`tenant.support_url`). Maybe add example values or descriptions?

khorne3:
@mpaktiti Doc should be more helpful now :)

PR #4795, opened 2017-08-09
Clarify user object properties during login

jerriep:
https://auth0-docs-content-pr-4795.herokuapp.com/docs/rules/current#rule-syntax
https://auth0-docs-content-pr-4795.herokuapp.com/docs/user-profile/user-profile-structure

Based on user feedback

### DIFF
--- a/articles/rules/current/index.md
+++ b/articles/rules/current/index.md
@@ -44,6 +44,10 @@ A Rule is a function with the following arguments:
 
 * `callback`: a function to send back potentially modified tokens back to Auth0, or an error. Because of the async nature of Node.js, it is important to always call the `callback` function, or else the script will timeout.
 
+::: note
+Rules execute after the actual login, so for the `user` object, properties such as the `last_ip` and `last_login` would be the values associated with the login that triggered the rules.
+:::
+
 ## Examples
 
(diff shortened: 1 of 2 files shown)


articles/rules/current/index.md
```diff
@@ -44,6 +44,10 @@ A Rule is a function with the following arguments:
 
 * `callback`: a function to send back potentially modified tokens back to Auth0, or an error. Because of the async nature of Node.js, it is important to always call the `callback` function, or else the script will timeout.
 
+::: note
```
mpaktiti: @jerriep I would move this right after the `user` bullet. We should also add this info in this doc https://auth0-docs-content-pr-4795.herokuapp.com/docs/user-profile/user-profile-structure Wdyt?

jerriep: @mpaktiti Yeah, I had it there but did not like it because I felt it broke up the flow of the bullets. I will move it there and let you see what it looks like.

As for the other document, that is where I wanted to add it to originally, because that is the document on which the user gave feedback.

In the end I did not put it there because that page is not really applicable to Rules, but to the user profile in general. I will add a note though, and make it clear that in the instance that the user profile is accessed from within a rule, that those conditions apply...

jerriep: @mpaktiti I made the changes as requested. Let me know if anything else is needed

PR #4801, opened 2017-08-11
Updating for new Salesforce UI

david-bc:
Note that I added links to externally hosted images. If you would prefer to host them in the static folder, downloading and committing them would not take long


mpaktiti:
@david-bc Thank you for the contribution! We moved the images to our repo (and made some other changes along the way) 👍

PR #4816, opened 2017-08-15
Update migration.md

jeffreylees:
https://auth0-docs-content-pr-4816.herokuapp.com/docs/libraries/lock-ios/v2/migration

https://trello.com/c/6B0ekwVm

### DIFF
--- a/articles/libraries/lock-ios/v2/migration.md
+++ b/articles/libraries/lock-ios/v2/migration.md
@@ -272,6 +272,6 @@ Auth0
 
 ### Delegation
 
-Delegation is not available through Lock. It can be implemented via a legacy method in [Auth0.Swift](/libraries/auth0-ios) for tenants which existed prior to June 2017, but delegation is deprecated and not recommended for most use cases. See the [Grant Types](/clients/client-grant-types#secure-alternatives-to-the-legacy-grant-types) documentation for more details.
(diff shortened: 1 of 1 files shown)


articles/libraries/lock-ios/v2/migration.md
```diff
@@ -271,21 +270,8 @@ Auth0
...
-  .start() { result in
-    switch result {
-    case .success(let payload):
-      // payload will have your firebase token
-    case .failure(let error):
-      // Handle Error
-    }
-  }
-```
+Delegation is not available through Lock. It can be implemented via a legacy method in [Auth0.Swift](/libraries/auth0-swift) for tenants which existed prior to June 2017, but delegation is deprecated and not recommended for most use cases. See the [Grant Types](/clients/client-grant-types#secure-alternatives-to-the-legacy-grant-types) documentation for more details.
```
mpaktiti: Maybe it's me, and the fact that I don't know the first thing about Auth0.Swift, but I followed the link and didn't understand what is the alternative that Auth0 recommends to me.

mpaktiti (changes requested):
Left a comment, it does not look clear to me but maybe it's my ignorance getting in the way. Wdyt?

articles/libraries/lock-ios/v2/migration.md
jeffreylees: Putting a pin in this one until we decide how to handle this type of situation re: our Slack conversation

PR #4817, opened 2017-08-15
Adding Password Manager section

jeffreylees:


mpaktiti (changes requested):
Looks good, but we still have at the roadmap an entry for `1Password support` https://github.com/auth0/docs/pull/4817/files#diff-e74f6490c53bc4463a02a94e1d7dbde6L233 Is the support conditional or did you just forget to remove it?

image

jeffreylees:
Nope, I didn't even notice that there, thanks!

PR #4818, opened 2017-08-15
typos, grammar

roryok:
'results from' is acceptable, but 'results when' doesn't sound exactly right to me.


mpaktiti:
@roryok You are right, I'm going to fix that. Thank you! 👍

PR #4820, opened 2017-08-15
change 'add on' to 'Addon' for consistency

roryok:


mpaktiti:
@roryok Thanks again! 💯

PR #4821, opened 2017-08-15
Update using-auth0-with-multi-tenant-apps.md

serhatcan:
As you know there is an important difference between user and app metadata. For multi-tenancy, users can not use user_matedata, they should use app_metadata. Although you may refer to both of these metadata fields here as User Metadata, I think this is confusing. Suggest changing User Metadata to App Metadata


mpaktiti:
You are right @srhtcn , thanks for clarifying that! 👍

PR #4833, opened 2017-08-16
Reworded first paragraph

roryok:
Hopefully makes things a little clearer. There was a sentence in there starting with 'But', so I reworded the whole thing to remove that. =)


mpaktiti:
Looks good, thanks @roryok ! 👍

PR #4842, opened 2017-08-18
Update Guides page

khorne3:
https://auth0-docs-content-pr-4842.herokuapp.com/docs/topics/guides


khorne3:
What do you think? To light? Just right? I was trying to keep the list from getting overwhelming, since I think beginners would be most interested in something like this.

mpaktiti (approved):
Much better 👍

PR #4844, opened 2017-08-18
Added notes about metadata structure

cabralmartin:
Added notes about metadata structure


articles/metadata/index.md
```diff
@@ -103,6 +103,36 @@ Currently, Auth0 limits the total size of your user metadata to 16 MB. However,
...
+
+```json
+    {
+        "address": "My Street"
+    }
+```
+
+This second user will be violating the formerly inferred schema since `address` is now a String. The net result is that any User Search containing the `address` field in the query will not return the second user as part of the results.
+
+Keep in mind that while the user won't participate of searches on the `address` attribute, you will be able to search for any other attribute and the result will contain all matching users.
```
hugoArregui: I think this is a bit misleading: the user won't participate in searches on any metadata attribute, no search address

PR #4845, opened 2017-08-19
SAML endpoint reference confusing

dabruhce:
I believe this once was a multi-step tutorial.  The reference to use the metadata from the previous step is confusing if the user doesn't have much experience with the Auth0 platform.  This data is found in the client advanced settings under the end points tab, in a field labeled SAML Metadata URL.

Updated the ref in the table to give slight guidance on where to find the saml settings.


mpaktiti:
Thank you @tkntobfrk ! 👍

PR #4846, opened 2017-08-21
Updating HLP docs and recommending from others

jeffreylees:
Adding HLP to sidebar, recommending HLP, updates to HLP docs

### DIFF
--- a/articles/libraries/auth0js/v8/index.md
+++ b/articles/libraries/auth0js/v8/index.md
@@ -149,9 +149,9 @@ webAuth.redirect.loginWithCredentials({
 });
 ```
 
-::: warning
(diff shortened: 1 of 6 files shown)


articles/libraries/auth0js/v8/index.md
```diff
@@ -149,9 +149,9 @@ webAuth.redirect.loginWithCredentials({
 });
 ```
 
-::: warning
-`webauth.redirect.loginWithCredentials` will be deprecated, and we recommend that you use `webauth.client.login` instead. However, if you're using `webauth.redirect.loginWithCredentials` on a Hosted Login Page, you can continue to do so.
-:::
+The use of `webauth.redirect.loginWithCredentials` is not recommended when using Auth0.js in your apps; it is recommended that you use `webauth.client.login` instead.
```
mpaktiti, line 152: @jeffreylees Can we explain why this is not recommended?

PR #4850, opened 2017-08-21
Fix rule sample

khorne3:
https://auth0-docs-content-pr-4850.herokuapp.com/docs/rules/current#copy-user-metadata-to-id-token


mpaktiti (approved):
💯

PR #4855, opened 2017-08-22
3di proposed changes to the Token Renewal article

awie3di:
Submitting a separate pull request for the updated Token Renewal (Angular2+) article

I'm adding a separate pull request for the conflicted files.


mpaktiti (approved):
LGTM

articles/quickstart/spa/angular2/05-token-renewal.md
```diff
@@ -110,7 +112,7 @@ export class AppComponent {
 }
 ```
 
-Since client-side sessions should not persist after the user logs out, call `unscheduleRenewal` in the `logout` method to unschedule the renewal.
+Client-side sessions should not be renewed after the user logs out. Call the `unscheduleRenewal` method in the `logout` method.
```
aaguiarz: I think it reads better to say:

Since client-side sessions should not be renewed after the user logs out, call the `unscheduleRenewal` method in the `logout` method.

@mpaktiti ?

PR #4868, opened 2017-08-22
Add link to user profile docs

khorne3:
https://auth0-docs-content-pr-4868.herokuapp.com/docs/api-auth/tutorials/adoption/scope-custom-claims

### DIFF
--- a/articles/api-auth/tutorials/adoption/scope-custom-claims.md
+++ b/articles/api-auth/tutorials/adoption/scope-custom-claims.md
@@ -11,7 +11,7 @@ The behavior of the `scope` parameter has been changed to conform to the [OIDC s
 Instead of requesting arbitrary application-specific claims, clients can request any of the standard OIDC scopes such as `profile` and `email`, as well as any [scopes supported by the API they want to access](/api-auth/tutorials/adoption/api-tokens).
 
 ::: note
-See  for additional information on the various types of Auth0 User Profiles.
+See [User Profile](/user-profile) for additional information on the various types of Auth0 User Profiles.
 :::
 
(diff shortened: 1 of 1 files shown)


mpaktiti (changes requested):
This note looks out of context. The linked doc does not even have a section on how to set namespaced custom claims using the dashboard. And I also don't get which are `the various types of Auth0 User Profiles` (they are explained later on in the current doc, but I cannot find this info in the linked doc).

I'm not sure if this negative feedback is actionable. This doc is not supposed to teach you how to set or retrieve custom claims, programmatically or not (even though we may be missing a doc with this content gathered in one place). It's just an adoption guide, describing the differences between one version and another.

The way I see it, it would make more sense if we enhanced the Scopes doc https://auth0.com/docs/scopes with any missing cases (for example, it does not say how to create custom claims via the dashboard), and then add a note in this doc saying for more information on how to do this and that refer to Scopes. Wdyt?

khorne3:
@mpaktiti So I agree with most of your comments, but I do think we need to mention something about user profiles, since we kind of just throw that word in the adoption guide and not mention anything further. I think we should either link to user profile docs so that people know what it is or we can not mention user profiles at all so that we don't casually mention something and then not explain it. This would be fine if we had a user who came in w/ at least a basic understanding of Auth0, but if they landed on this page via search engine results, for example, it might be problematic.

khorne3:
@mpaktiti I've made a note to review the scopes doc and to link at a later point if we think that's best. I think there's some reorganization of this doc that could be done, since I was kind of confused at first. Closing this particular PR for now.

PR #4891, opened 2017-08-28
Update styleguide

khorne3:
This is to add the highlights of the new dashboard-related terminology changes to the styleguide.

Also, is it style guide or styleguide? :)


mpaktiti:
@khorne3 I'm wondering if this content fits better in this doc https://github.com/auth0/docs/blob/master/WORDS.md since it's about when to use which word. What do you think?

khorne3:
@mpaktiti Moved. Thanks!

PR #4921, opened 2017-08-30
Add note on deleting metadata

khorne3:
https://auth0-docs-content-pr-4921.herokuapp.com/docs/metadata/management-api


mpaktiti (approved):
Looks good Katie. I updated also the UI element at `PATCH` to make it more prominent (plus added a link to the new section)

PR #4926, opened 2017-08-31
Fixed typo

Supratentorial:
Should be "...weight in..." not "...weight in..."


mpaktiti:
Oops, good catch! Thanks you for the fix @Supratentorial 👍

PR #4928, opened 2017-08-31
Add note on caching connections

khorne3:
https://auth0-docs-content-pr-4928.herokuapp.com/docs/connections/database/mysql


articles/connections/database/mysql.md
```diff
@@ -87,6 +87,8 @@ function login(email, password, callback) {
 
 This script connects to a **MySQL** database and executes a query to retrieve the first user with `email == user.email`. With the `bcrypt.compareSync` method, it then validates that the passwords match, and if successful, returns an object containing the user profile information including `id`, `nickname`, and `email`. This script assumes that you have a `users` table containing these columns. You can tweak this script in the editor to adjust it to your own requirements.
 
+You can also [cache connections](/rules/current#cache-expensive-resources) such that it will survive individual execution.
```
mpaktiti: No need to add the version, the link works like that: `[cache connections](/rules#cache-expensive-resources)`

That way if the versions change, we don't have to go over the docs and update all these links.

mpaktiti (changes requested):
I added a note for a quick fix at the link. While you are at it can you also rename this doc to something more generic? Like `/connections/database/custom.md`. I think we also had a slack request about this at some point. We will also need a redirect.

mpaktiti (approved):

PR #4956, opened 2017-09-07
Update QuickStart iOS Swift Authorization to use idToken custom claims

cocojoe:
Sample PR https://github.com/auth0-samples/auth0-ios-swift-sample/pull/43

### DIFF
--- a/articles/quickstart/native/ios-swift/05-authorization.md
+++ b/articles/quickstart/native/ios-swift/05-authorization.md
@@ -17,55 +17,59 @@ budicon: 500
 
 Many identity providers will supply access claims, like roles or groups, with the user. You can request these in your token by setting `scope: openid roles` or `scope: openid groups`. However, not every identity provider provides this type of information. Fortunately, Auth0 has an alternative to it, which is creating a rule for assigning different roles to different users.
 
-## Create a Rule to Assign Roles
+## Create A Rule To Assign Roles
 
-To create a rule, just go to the [new rule page](${manage_url}/#/rules/new). You can create it from scratch or use an existing template.  These templates are written by Auth0 team to assist you complete common tasks.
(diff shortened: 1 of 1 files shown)


articles/quickstart/native/ios-swift/05-authorization.md
```diff
@@ -17,55 +17,59 @@ budicon: 500
...
+}
 ```
-if (user.email.indexOf('@example.com') > -1)
-```
+
+The rule will run every time a user attempts to authenticate. If the user has a valid `email` and the email's domain is `admin.com`, the user will be granted the `user` and `admin` roles. On any other case, the user will be granted a simple `user` role. This new claim is saved in the *id_token* under the `https://access.control/roles` name, which you will have to check on your app.
 
 ::: note
-You can set roles other than `admin` and `user` or customize the rule as needed.
+Feel free to customize the conditions and the roles to grant on each of them. You can define as many roles as your project requires, beyond `admin` and `user`. There are some restrictions on the naming of the claims detailed in the [rules documentation](/rules/#hello-world).
```
lbalmaceda: URL is `/rules#hello-world`

articles/quickstart/native/ios-swift/05-authorization.md
```diff
@@ -17,55 +17,59 @@ budicon: 500
...
-        }
-}
+guard
+    let idToken = self.keychain.string(forKey: "id_token"),
+    let jwt = try? decode(jwt: idToken),
+    let roles = jwt.claim(name: "https://access.control/roles").array
+    else { // Couldn't retrieve claim }
+
+    if roles.contains("admin") {
+        // Authorised
```
lbalmaceda: Does the sample do something when these criteria are met? like allowing the user to access a different screen or something?

lbalmaceda (changes requested):

articles/quickstart/native/ios-swift/05-authorization.md
cocojoe: Either works 😄

articles/quickstart/native/ios-swift/05-authorization.md
cocojoe: Yes, I could add that in comments

lbalmaceda (approved):

PR #4963, opened 2017-09-07
Clean up doc

khorne3:


mpaktiti (approved):
Looks good Katie 👍 I added the `next-steps` element at the last paragraph, even though you are not listing next steps. I think it looks better this way, and since the list is short (up to 5 elements) we can use that. Thanks!

PR #4966, opened 2017-09-08
Updating rule for ip ranges + linting doc

jeffreylees:


mpaktiti:
@jeffreylees I'm merging this but bear in mind that the `/protocols#` links are broken. Send another PR to fix them. Thanks!

PR #4979, opened 2017-09-11
Updated Office 365 custom provisioning rule code and minor adjustments

nicosabena:


mpaktiti (changes requested):
Hey Nico, can you shrink the screenshots a bit? Max 1500px width. And before you take the screenshot run this so your browser is resized to the standard size we use.

```
osascript -e 'tell application "Chrome" to set the bounds of the front window to {100, 150, 1200, 900}'
```

If you are too busy let me know and I can fix it myself. Thanks!

PR #4982, opened 2017-09-12
Updates guide to show how to grant scopes to specific roles

jerriep:
https://auth0-docs-content-pr-4982.herokuapp.com/docs/architecture-scenarios/application/spa-api


mpaktiti:
@jerriep Looks good 👍 I think though that we should add also a bullet at the TL;DR, wdyt?

jerriep:
@mpaktiti Updated as requested :)

PR #4994, opened 2017-09-14
Added missing sections of importance between v1 -> v2

cocojoe:


jeffreylees:
I was going to add my Lock iOS docs changes to this PR, + move these new documents to the Auth0.Swift section, add to sidebar, etc, but apparently I don't have permissions to write to the branch? Not sure if we can fix that, or if we should just close this one and I can send a new PR, or? @mpaktiti

articles/connections/passwordless/ios-touch-id-objc.md
```diff
@@ -12,6 +12,8 @@ languages:
 
 ::: warning
 This feature is disabled for new tenants as of June 8th 2017. Any tenant created after that date won't have the necessary legacy [grant types](/clients/client-grant-types) to use Touch ID. This document is offered as reference for older implementations.
+
+Please take a look at an alternative approach to [Touch ID Authentication](/libraries/lock-ios/v2/touchid-authentication).
```
cocojoe: @jeffreylees this might look nicer broken out into `::: note` vs a `::: warning`.  Your call, team docs 😄 Same goes for the other file.

mpaktiti:
@jeffreylees I can see the message `Add more commits by pushing to the changed-update-docs-ios-lock-v2-add-missing-sections branch on auth0/docs.` This doesn't work for you?

jeffreylees:
@mpaktiti yeah, dunno. I see the message, but when I try to commit to that branch using either the CLI or a graphical client, I get permissions errors. I can pull down fine, just not write. 

I always commit my stuff to branches of my own fork, then PR it - it's possible I don't have any write permissions at all on the docs repo, regardless of branch?

cocojoe:
@jeffreylees what is happening here? Could always merge then work on a PR after that?

jeffreylees:
@mpaktiti yes please, let's merge it, as soon as we can, and I'll PR my changes directly after. We can figure out whatever's wrong with my GH permissions after that.

articles/libraries/lock-ios/v2/save-and-refresh-jwt-tokens.md
```diff
@@ -0,0 +1,103 @@
...
+}
+```
+
+Renewing a user's credentials works exactly the same way, if the token has expired. The Credentials Manager will automatically renew, store the new credentials to the Keychain and return them in the closure.
+
+## SimpleKeychain
+
+If you are migrating from v1, you may already be familiar with using [SimpleKeychain](https://github.com/auth0/SimpleKeychain) to handle iOS Keychain access.
+
+First thing we need to do is store any tokens we need, in this case we will store the `access_token` and `refresh_token` in the Keychain after a successful authentication.
```
lbalmaceda: Keep talking to the user: we -> you.

articles/libraries/lock-ios/v2/save-and-refresh-jwt-tokens.md
```diff
@@ -0,0 +1,103 @@
...
+             keychain.setString(accessToken, forKey: "access_token")
+             keychain.setString(refreshToken, forKey: "refresh_token")
+             // You might want to route to a user profile screen at this point
+        }
+}
+```
+
+Once you have those stored, you can at any point request a fresh [Credentials](https://github.com/auth0/Auth0.swift/blob/master/Auth0/Credentials.swift) instance.
+
+### Renewing User Credentials
```
lbalmaceda, line 93: Renewing User's Credentials

articles/libraries/lock-ios/v2/touchid-authentication.md
```diff
@@ -0,0 +1,68 @@
+---
+section: libraries
+toc_title: Touch ID Authentication
+description: How to implement Touch ID authentication with Lock iOS.
+---
+# Lock iOS: Touch ID Authentication
+
+Here's the scenario: You are using `webAuth` to present the HLP (Hosted Login page) for the user to Login. After user authentication you want to store the user's credentials and use the `refreshToken` to renew the user's credentials without having to present the HLP. Instead you want to use Touch ID to authenticate and renew the user's credentials.
```
lbalmaceda: This makes some noise. "After user authentication you want to store the user's credentials and use the `refreshToken` to renew the user's credentials without having to present the HLP."

It seems like we're assuming that the user instead of using the saved credentials, the user will always renew them before use. We should be saying something like "After user authentication, you want to store the credentials and use them as long as they are valid. Once they expire, you would want to renew them using the `refreshToken` in order to avoid presenting the HLP again."   instead.......

articles/libraries/lock-ios/v2/touchid-authentication.md
```diff
@@ -0,0 +1,68 @@
...
+
+You will to be using the [Credentials Manager](https://github.com/auth0/Auth0.swift/blob/master/Auth0/CredentialsManager.swift) utility in [Auth0.swift](https://github.com/auth0/Auth0.swift/) to streamline the management of user credentials and perform the Touch ID authentication.
```
lbalmaceda: you will ~to~ be using

articles/libraries/lock-ios/v2/touchid-authentication.md
```diff
@@ -0,0 +1,68 @@
...
+
+First, import the `Auth0` module:
+
+```swift
+import Auth0
+```
+
+### Credentials Manager
+
+Setup the Credentials Manager and enable Touch ID authentication:
```
lbalmaceda: I'd add. "You can pass the title to show in the touch id prompt screen"

articles/libraries/lock-ios/v2/touchid-authentication.md
```diff
@@ -0,0 +1,68 @@
...
+credentialsManager.credentials { error, credentials in
+    guard error == nil, let credentials = credentials else {
+        // Handle Error
+        // Fallback to Login Screen
+    }
+    // Continue routing the user as authentication was a success
+}
+```
+
+There is no need to store the credentials again as you did with the Login as this is done by the Credentials Manager for you during renewal.
```
lbalmaceda: "to store the *new* credentials again". Emphasis on these just acquired credentials

lbalmaceda (changes requested):
Looks good so far. I have a few comments:
* Don't mention the whole v1 (keychain thing) approach first. Instead, say that if they are using v1 they should be checking THIS (link) section that links them to a section below the newer approach.
* You're not making use of the "hasCredentials" method. I think its important to check for them BEFORE saying "log in, then persist them so you can use the manager".

articles/libraries/lock-ios/v2/save-and-refresh-jwt-tokens.md
cocojoe: You don't need an apostrophe here. If it was `the user` or `a user` then you would.

jeffreylees (approved):
Will send a following PR to alter some stuff, all good as of now.

lbalmaceda (approved):

PR #4997, opened 2017-09-14
Add daccount connection page

erjohnson:
Updating list of identity providers.

[Connect your app to Docomo dAccount](https://auth0-docs-content-pr-4997.herokuapp.com/docs/connections/social/docomo)


mpaktiti (commented):
I removed this text from the intro: "This doc refers to the client steps to connect your client. If you are looking to manage authentication in your application, see [Next Steps](#next-steps) below."

I suppose you copied it from some other social doc, so can you please send a PR removing it from all docs? I added a toc so the user can see the next steps without having to scroll down.

PR #4998, opened 2017-09-15
Auth0.js parameter updates

jeffreylees:
Feedback from a customer noted that we have required parameters in `authorize()` - wrongly so, they're required by the `authorize` endpoint, not as parameters to the auth0.js method (if they're not provided, the SDK defaults them). Also made some suggestions about the below paragraph, which just needed removed.

Lastly, separate request to document `leeway` parameter + explain what error might cause you to need to set a `leeway` to start with.


mpaktiti (approved):
Looks good. I fixed one more link.

PR #5000, opened 2017-09-15
Correct misspelling of "users"

d60402:


mpaktiti:
Thanks for the fix @d60402 ! 👍

PR #5001, opened 2017-09-15
Updating Access Token doc

jeffreylees:
Re-arranging access-token doc in an attempt to make it first introduce access tokens as opaque tokens which are not JWTs by necessity (although often at Auth0) and then further discuss the JWT access tokens used at Auth0 for custom APIs.


articles/tokens/access-token.md
```diff
@@ -6,15 +6,25 @@ toc: true
...
 
-Auth0 generates access tokens in [JSON Web Token (JWT)](/jwt) format, an industry standard. JWTs contain three parts: a header, a set of claims, and a signature:
-- The header contains metadata about the type of token and the cryptographic algorithms used to secure its contents.
-- The set of claims contains verifiable security statements such as the identity of the user and the permissions they are allowed.
-- The signature is used to validate that the token is trustworthy and has not been tampered with.
+## Access Tokens at Auth0
+
+Auth0 currently generates access tokens in two formats:
+
+* As opaque strings, when `/userinfo` is the audience (no custom API audience, no openid scope).
```
aaguiarz: I think we need to say:

- As opaque strings, when '$domain/userinfo' is the audience.
- As a JWT when a custom API is specified in the audience

and then clarify something like:

If you ask for a custom API audience and use the 'openid' scope, and the API and the Client have the same signing algorithm, you will get a JWT token that is valid for both the /userinfo endpoint and the API endpoint

articles/tokens/access-token.md
```diff
@@ -107,9 +114,39 @@ For more information on the namespaced format of custom claims, refer to [User p
...
+Auth0 creates access tokens in JWT format for custom APIs. JWTs contain three parts: a header, a set of claims, and a signature:
+
+* The header contains metadata about the type of token and the cryptographic algorithms used to secure its contents.
+* The set of claims contains verifiable security statements such as the identity of the user and the permissions they are allowed.
+* The signature is used to validate that the token is trustworthy and has not been tampered with.
+
+### Authorize Access Tokens
 
-Once a client has obtained an access token, it will include that token as a credential when making API requests.
+Once a client has obtained an access token for a custom API, it will include that token as a credential when making API requests.
```
aaguiarz: I'd remove the "for a custom API", this happens for any API

articles/tokens/access-token.md
```diff
@@ -6,15 +6,25 @@ toc: true
 
 ## Overview
 
-The Access Token, commonly referred to as `access_token` in code samples, is a credential that can be used by a client to access an API. The `access_token` should be used as a `Bearer` credential and transmitted in an HTTP `Authorization` header to the API. Auth0 uses access tokens to protect access to the Auth0 Management API.
+The Access Token, commonly referred to as `access_token` in code samples, is a credential that can be used by a client to access an API. The `access_token` can be any type of token (not necessarily a JSON Web Token) and is meant for the API. The purpose of the `access_token` is to inform the API that the bearer of this token has been authorized to access the API and perform specific actions (as specified by the `scope` that has been granted). The `access_token` should be used as a `Bearer` credential and transmitted in an HTTP `Authorization` header to the API.
```
mpaktiti: I would probably remove this sentence `(not necessarily a JSON Web Token)` since you haven't made the claim that it is a JWT. You could instead type some types in this parenthesis `([JSON Web Tokens](/jwt), opaque strings, whatever-else)`

articles/tokens/access-token.md
mpaktiti: @aaguiarz question: if I set only `userinfo` at my audience, but my client is OIDC-conformant, won't the token be a JWT? (at least that's what we say in some docs, including the pipeline2 adoption guide)

articles/tokens/access-token.md
```diff
@@ -6,15 +6,25 @@ toc: true
...
+
+![Token Signing Algorithm - Client](/media/articles/tokens/tokens-algorithm-client.png)
+
+When setting up an API in the [Dashboard](${manage_url}/#/apis), the signing algorithm can also be specified.
+
+![Token Signing Algorithm - API](/media/articles/tokens/tokens-algorithm-api.png)
 
 ::: note
-The [Auth0 Management API v1](/api/management/v1) (which has been deprecated) uses an opaque token format in which claims are referenced in a separate database, rather than directly in the token.
+While Auth0 currently uses JWTs for many access tokens, the important thing to remember is that the client should not be depending on the access token to be any specific format, but instead should treat the access token as opaque.
```
mpaktiti: I would remove this `While Auth0 currently uses JWTs for many access tokens`

mpaktiti: And maybe use the `warning` UI element since this is something we want to put emphasis on

articles/tokens/access-token.md
```diff
@@ -107,9 +114,39 @@ For more information on the namespaced format of custom claims, refer to [User p
...
+
+The token lifetime can be controlled on a per-API basis. The validity period can be increased or decreased based on the security requirements of each API.
+
+To configure the amount of time a token lives, use the **Token Expiration (Seconds)** field for your API at the [APIs dashboard](${manage_url}/#/apis). The default value is `24` hours (`86400` seconds).
+
+![Token Expiration - API](/media/articles/tokens/tokens-expiration-api.png)
+
+Once expired, an access token can no longer be used to access an API. In order to obtain access again, a new access token needs to be obtained. This can be done by repeating the OAuth flow used to obtain the initial access token.
+
+In some situations, it is desirable to have permanent, ongoing access to an API without having to repeat an OAuth flow. This is often referred to as `offline access`, and is possible with the use of a [refresh token](/tokens/refresh-token).
```
mpaktiti: Maybe use `offline_access` so the users get accustomed to the scope's name?

articles/tokens/access-token.md
```diff
@@ -107,9 +114,39 @@ For more information on the namespaced format of custom claims, refer to [User p
...
+
+A refresh token is issued from the OAuth 2.0 endpoints along with the access token. When the access token expires, the refresh token can be used to obtain a fresh access token with the same permissions, without further involvement from a user. Note that offline access is enabled as a policy of the API the access token grants access to. If the API does not permit offline access, a refresh token will not be issued. In such circumstances, the OAuth flow must be repeated in order to obtain a new access token.
```
mpaktiti: `Note that offline access is enabled as a policy of the API the access token grants access to. If the API does not permit offline access, a refresh token will not be issued.`

You do not tell them how they can check and change this (field at API settings)

articles/tokens/access-token.md
```diff
@@ -107,9 +114,39 @@ For more information on the namespaced format of custom claims, refer to [User p
...
+
+For more information on refresh tokens and how to use them refer to: [Refresh Token](/tokens/refresh-token).
```
mpaktiti: I'd make this a note

articles/tokens/access-token.md
mpaktiti: plus you are already under the "with Custom APIs" section

articles/tokens/access-token.md
aaguiarz: No, it's not. That's the main confusion. If you specify userinfo as the audience the access_token is never a JWT.

If you specify a custom API audience, and you ask for the 'openid' scope, you get a JWT valid for the custom API and for userinfo.

articles/api-auth/tutorials/verify-access-token.md
```diff
@@ -2,14 +2,14 @@
 description: How an API can verify a bearer JWT access token
```
mpaktiti, line 2: Update the description metadata

articles/tokens/access-token.md
```diff
@@ -6,15 +6,29 @@ toc: true
...
+## Access Tokens at Auth0
 
-::: note
-The [Auth0 Management API v1](/api/management/v1) (which has been deprecated) uses an opaque token format in which claims are referenced in a separate database, rather than directly in the token.
+Auth0 currently generates access tokens in two formats:
+
+* As opaque strings, when `${account.namespace}/userinfo` is the audience.
+* As a [JSON Web Token (JWT)](/jwt) when a custom API is specified as the audience.
+
+JWTs are an industry standard, and are the format access tokens are generated in by Auth0 when a custom API is the audience. When a custom API audience is specified along with an `openid` scope, an access token is generated that will be valid for both the `/userinfo` endpoint and for the custom API.
```
mpaktiti: The first sentence of this paragraph (JWTs are an industry standard...) says the same thing as the previous bullet, I'd remove it

articles/tokens/access-token.md
```diff
@@ -6,15 +6,29 @@ toc: true
...
-- The set of claims contains verifiable security statements such as the identity of the user and the permissions they are allowed.
-- The signature is used to validate that the token is trustworthy and has not been tampered with.
+## Access Tokens at Auth0
 
-::: note
-The [Auth0 Management API v1](/api/management/v1) (which has been deprecated) uses an opaque token format in which claims are referenced in a separate database, rather than directly in the token.
+Auth0 currently generates access tokens in two formats:
+
+* As opaque strings, when `${account.namespace}/userinfo` is the audience.
+* As a [JSON Web Token (JWT)](/jwt) when a custom API is specified as the audience.
```
mpaktiti: In these bullets we mention audience and the user might not be familiar with it. I would add a note underneath saying that audience is a parameter set during the authorization (link to https://auth0.com/docs/api/authentication#authorize-client) and that it holds the unique identifier of the target API you want to access.

mpaktiti (changes requested):

articles/api-auth/tutorials/verify-access-token.md
jeffreylees: To what? That still seems correct for this document.

PR #5007, opened 2017-09-16
Fix typos

d60402:


mpaktiti:
Thanks for fixing that @d60402 ! 👍

PR #5008, opened 2017-09-16
Update 01-authorization.md

bentedder:
The block of code for the jwt method actually needs to have the key `aud` instead of `audience` for it to function properly.

I've attempted 3 times to re-do this PR to avoid the change on line 96. Not sure why it keeps getting pulled in.

### DIFF
--- a/articles/quickstart/backend/nodejs/01-authorization.md
+++ b/articles/quickstart/backend/nodejs/01-authorization.md
@@ -59,7 +59,7 @@ const checkJwt = jwt({
   }),
 
   // Validate the audience and the issuer.
-  audience: '${apiIdentifier}',
+  aud: '${apiIdentifier}',
   issuer: `https://${account.namespace}/`,
   algorithms: ['RS256']
 });


aaguiarz:
Actually, it needs to be 'audience'. If you specify 'aud' the audience will not be checked, so it will always succeed.

Can you share the code you are using to get the access_token?

bentedder:
Hmmm. You're right, it is just always succeeding. The code I'm using in js (react) is:
```
auth0 = new auth0.WebAuth({
  domain: 'myapp.auth0.com',
  clientID: 'xyzxyzxz',
  redirectUri: 'http://localhost:3000/lcb',
  audience: 'localhostAPI',
  responseType: 'token id_token',
  scope: 'openid profile'
});
```
When I use that coupled with this in node.js, I get an `UnauthorizedError: jwt audience invalid. expected: localhostAPI` error
```
var jwt = require('express-jwt');
var jwksRsa = require('jwks-rsa');

var checkJwt = jwt({
  secret: jwksRsa.expressJwtSecret({
    cache: true,
    rateLimit: true,
    jwksRequestsPerMinute: 50,
    jwksUri: 'https://myapp.auth0.com/.well-known/jwks.json'
  }),
  audience: 'localhostAPI',
  issuer: 'https://myapp.auth0.com/',
  algorithms: ['RS256']
});
```
The reason I went with remove audience and replacing with `aud` was via a few forum posts, some of which I'll want to go back and add corrections to if that's the case :)

bentedder:
You know, I'm not so sure actually, I was digging in to the jsonwebtoken code and see the following: https://github.com/auth0/node-jsonwebtoken/blob/master/verify.js#L132

Doesn't that say it's looking for an `aud` prop instead of an `audience` prop?

aaguiarz:
That code seems to be comparing the 'audience' option in the jwt() function parameters with the 'aud' key in the token, so it's consistent with the behavior I described, isn't it?

If you use the jwt.io chrome extension to look at the access_token you get, does it have the localhostAPI audience?

bentedder:
good call on the chrome extension. but the decoded access_token from the parseHash function is:
```
{
  "iss": "https://XYZ.auth0.com/",
  "sub": "auth0|XYZ",
  "aud": [
    "localhostAPI",
    "https://XYZ.auth0.com/userinfo"
  ],
  "azp": "XYZ",
  "exp": 1505921272,
  "iat": 1505914072,
  "scope": "openid profile"
}
```
looks like the `aud` param is an array. I didn't expect that. i wonder if it's because i've also requested the 'profile' scope.

aaguiarz:
Yes, when you use your API identifier as the audiene and request an openid scope, you get a JWT you can use to call both your API and the /userinfo endpoint, so you get two audiences in an array.

We are updating the access_token docs right now to explain that better.

bentedder:
Ah. that makes sense. For now then yeah it makes sense to close this PR. Thanks for the info.

PR #5011, opened 2017-09-18
Replace instances of app with client

khorne3:
The logout doc still referred to app instead of client (I also made minor edits to the doc).

https://auth0-docs-content-pr-5011.herokuapp.com/docs/logout#set-the-allowed-logout-urls-at-the-client-level


mpaktiti (approved):
Looks good Katie 👍  I updated also the AuthN API endpoint that linked to this paragraph

PR #5017, opened 2017-09-19
3di proposed changes to the Android quickstart

awie3di:


articles/quickstart/native/android/_includes/_auth0.md
```diff
@@ -19,10 +19,12 @@ dependencies {
 ```
 
 ::: note
-You can check for the latest version on the repository [Readme](https://github.com/auth0/auth0.android#installation), in [Maven](http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22auth0%22%20g%3A%22com.auth0.android%22), or in [JCenter](https://bintray.com/auth0/android/auth0).
+You can check for the latest version on the [repository Readme](https://github.com/auth0/auth0.android#installation), in [Maven](http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22auth0%22%20g%3A%22com.auth0.android%22), or in [JCenter](https://bintray.com/auth0/android/auth0).
```
awie3di, line 22: @aaguiarz I wasn't sure of what is the latest version mentioned here. Could you help make this note more precise?

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -18,24 +18,25 @@ This tutorial will show you how to login and maintain an active session with Aut
...
+* `refreshToken`: Refresh token that can be used to request new tokens without signing in again. To learn more, see the [refresh token documentation](/tokens/refresh-token/current).
+* `tokenType`: The type of tokens issued by the server.
+* `expiresIn`: The amount of seconds before the tokens expire.
+* `expiresAt`: The date when the tokens expire.
+* `scope`: The scope that was granted to a user. This information is shown only if the granted token is different than the requested one.
 
-The Tokens are the objects used to prove your identity against the Auth0 APIs. Read more about them [here](https://auth0.com/docs/tokens).
+The tokens are the objects used to prove your identity against the Auth0 APIs. Read more about them in the [tokens documentation](https://auth0.com/docs/tokens).
 
+## Before You Start
```
awie3di: There are two general types of headers in the quickstarts: 
* Steps in the process
* Descriptions 

To introduce some separation between the two types, how about: 
* Using imperative for the steps: "Do X", "Configure Y"
* Using the "About X" form for descriptions? 

That way, the readers will have some idea of what to expect from the section following the header. 

@aaguiarz

articles/quickstart/native/_includes/_callback-url-introduction.md
```diff
@@ -1,3 +1,7 @@
-Callback URLs are the URLs that Auth0 invokes after the authentication process. Auth0 routes your application back to this URL and appends additional parameters to it, including a token. Since callback URLs can be manipulated, you will need to add your application's URL to your client's `Allowed Callback URLs` for security. This will enable Auth0 to recognize these URLs as valid. If omitted, authentication will not be successful.
+A callback URL is a URL in your application where Auth0 redirects the user after they have been authenticated. Auth0 adds parameters to the callback URL, inluding a token.
```
lbalmaceda: "redirects -the user- after -they- have been authenticated". Singular or plural?

typo ~inluding~ -> including. Maybe instead of "including a token" we can say something different. It's true, a token is always returned (auth_token) but you can get more of them (id_token, refresh_token) and even data regarding those tokens (scope, expires_in).

articles/quickstart/native/_includes/_callback-url-introduction.md
```diff
@@ -1,3 +1,7 @@
-Callback URLs are the URLs that Auth0 invokes after the authentication process. Auth0 routes your application back to this URL and appends additional parameters to it, including a token. Since callback URLs can be manipulated, you will need to add your application's URL to your client's `Allowed Callback URLs` for security. This will enable Auth0 to recognize these URLs as valid. If omitted, authentication will not be successful.
+A callback URL is a URL in your application where Auth0 redirects the user after they have been authenticated. Auth0 adds parameters to the callback URL, inluding a token. 
+
+Callback URLs can be manipulated. For security, add your application's URL to the **Allowed Callback URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings).
```
lbalmaceda: Manipulated how or by who? i.e. Attackers may manipulate the URL and redirect the traffic to them.

articles/quickstart/native/android/00-login.md
```diff
@@ -5,7 +5,7 @@ seo_alias: android
 budicon: 448
 ---
 
-This tutorial will show you how to integrate the Auth0 Login in your Android project in order to present the login box.
+This tutorial shows you how to integrate the Auth0 Login in your Android project and present the login box.
```
lbalmaceda: If "Auth0 Login" is a product I would say: "how to integrate Auth0 Login" or I would avoid saying Login at all: "how to integrate Auth0" as you later mention the login box.

articles/quickstart/native/android/00-login.md
```diff
@@ -23,5 +23,5 @@ This tutorial will show you how to integrate the Auth0 Login in your Android pro
 <%= include('_includes/_login') %>
 
 ::: note
-Please see [Browser-Based vs. Native Login Flows on Mobile Devices](/tutorials/browser-based-vs-native-experience-on-mobile) for information on choosing between the two types of login flows.
+Follow the [Browser-Based vs. Native Login Flows on Mobile Devices](/tutorials/browser-based-vs-native-experience-on-mobile) tutorial to learn how to choose between the two types of login flows.
```
lbalmaceda: This one linked is not a tutorial but rather an article.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -18,24 +18,25 @@ This tutorial will show you how to login and maintain an active session with Aut
   ]
 }) %>__
 
-For this, you will need to handle the user's `Credentials`. Let's take a look at this class, which is composed of five objects:
+You need the `Credentials` class to handle users' credentials. The class is composed of seven objects:
```
lbalmaceda: users' -> user's if referring to a single user. I would avoid the amount as it might change in the future and it's easier to update.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -18,24 +18,25 @@ This tutorial will show you how to login and maintain an active session with Aut
...
-* `expiresIn`: The amount in seconds in which the tokens will be deemed invalid.
-* `expiresAt`: The Date in which the tokens will be deemed invalid.
-* `scope`: The granted scope, if different from the requested one.
+* `accessToken`: Access token used by the Auth0 API. To learn more, see the [access token documentation](/tokens/access-token).
+* `idToken`: Identity token that proves the identity of the user. To learn more, see the [ID token documentation](/tokens/id-token).
+* `refreshToken`: Refresh token that can be used to request new tokens without signing in again. To learn more, see the [refresh token documentation](/tokens/refresh-token/current).
+* `tokenType`: The type of tokens issued by the server.
+* `expiresIn`: The amount of seconds before the tokens expire.
+* `expiresAt`: The date when the tokens expire.
+* `scope`: The scope that was granted to a user. This information is shown only if the granted token is different than the requested one.
```
lbalmaceda: "This information is shown only if the granted token **scope** is different than the requested one" or "This information is shown only if the granted **scope** is different than the requested one"

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -49,9 +50,13 @@ WebAuthProvider.init(auth0)
                 .start(LoginActivity.this, callback);
 ```
 
-## Save The User's Credentials
+## Save the User's Credentials
```
lbalmaceda, line 53: I think the Docs team suggested us to use word caps on titles/sections. If you don't want to keep them in "short words" then "Before You Start" above should be "Before you Start"

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -76,14 +81,13 @@ private final AuthCallback callback = new AuthCallback() {
...
+User credentials are stored in [Private mode](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE) in the seed project in the `SharedPreferences` file. You can achieve this with the `CredentialsManager`class. For details, check the implementation in the project code. There are better and more secure ways to store tokens, but we will not cover them in this tutorial.
 :::
 
-## At Startup: Check Token Existence
+## Check for Tokens when the Application Starts
 
-The main purpose of storing this token is to save users from having to re-enter their login credentials when relaunching the app. Once the app has launched, we need to check for the existence of an `access_token` to see if we can automatically log the user in and redirect the user straight into the app’s main flow, skipping the login screen.
+To save your users the effort of logging in every time they open your app, use their access tokens. You can check for the access token when a user starts the app. If you find the token, you can automatically log the user in and direct them straight to your app's main flow.
```
lbalmaceda, line 89: i would use redirect instead of direct

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -96,13 +100,16 @@ if (accessToken == null) {
...
 ```
 
-## Validate an Existing Token
+## Validate the Existing Token
 
-If the `access_token` exists, we need to check whether it’s still valid. To do so we can:
-* Check that the elapsed time since the token was obtained is lesser than the `expires_in` value received in with the credentials. For this to work we'll have to save the current time whenever we receive and store a new pair of credentials.
+If the access token exists, the next step is to check if it is valid. 
+You can choose between two options: 
+* Ssave the time when the user receives a new pair of credentials. Then, check if the time since the user got the access token exceeds the time defined in the `expires_in` value.
```
lbalmaceda: typo: ssave -> save
The concept is a bit wrong. When you receive a token you also receive the `expires_in` value. Save both the date/time at which it was requested and the expires_in value. When you need to use the token, check that the elapsed time since first saved is no longer than expires_in. If that's the case, the token is expired and needs to be renewed.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -96,13 +100,16 @@ if (accessToken == null) {
...
 * Call the Auth0 Authentication API and check the response.
 
-We will explain the latter approach by calling the `/userinfo` endpoint.
+::: note 
+This tutorial shows how to call the Auth0 Authentication API with the `/userinfo` endpoint.
```
lbalmaceda: This tutorial shows how to verify a Token by calling the Auth0 Authentication API on the `/userinfo` endpoint.

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -49,11 +51,14 @@ AuthenticationAPIClient authClient = new AuthenticationAPIClient(auth0);
...
 
 ::: note
-It's suggested that you add both the Auth0 `domain` and `clientId` to the `strings.xml` file rather than hardcode them.
+Do not hardcode the Auth0 `domain` and `clientId` values. We recommend you add them to the `strings.xml` file.
 :::
 
-Next, use the `access_token` to obtain the user id with the `AuthenticationAPIClient`. Although the call returns a `UserProfile` instance, this is a basic OIDC conformant profile and the only guaranteed claim is the `sub` which contains the user's id, but depending on the requested scope the claims returned may vary. With the `sub` value call the [Management API](https://auth0.com/docs/api/management/v2#!/Users) and get a complete user profile back.
+Use the access token and the `AuthenticationAPIClient` client to obtain the user's ID. 
+
+When you call the `AuthenticationAPIClient` client, it returns an instance of the `UserProfile` profile. This profile is a basic, OIDC-conformant profile, which guarantees only the `sub` claim. The `sub` claim contains the user's ID. Depending on the scope you requested, the remaining claims returned by the `UserProfile` are different.
```
lbalmaceda: "When you call the `AuthenticationAPIClient` client" -> You don't "call" the client, you call an API or its endpoints. This paragraph is missing which endpoint gets called (/userinfo). Also it's not just a "basic oidc conformant", that would be the case of requesting just the `openid` scope. The more scope you request, the larger the profile will be. Maybe we can link to this doc https://auth0.com/docs/scopes/current#example-asking-for-standard-claims

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -49,11 +51,14 @@ AuthenticationAPIClient authClient = new AuthenticationAPIClient(auth0);
...
+Do not hardcode the Auth0 `domain` and `clientId` values. We recommend you add them to the `strings.xml` file.
 :::
 
-Next, use the `access_token` to obtain the user id with the `AuthenticationAPIClient`. Although the call returns a `UserProfile` instance, this is a basic OIDC conformant profile and the only guaranteed claim is the `sub` which contains the user's id, but depending on the requested scope the claims returned may vary. With the `sub` value call the [Management API](https://auth0.com/docs/api/management/v2#!/Users) and get a complete user profile back.
+Use the access token and the `AuthenticationAPIClient` client to obtain the user's ID. 
+
+When you call the `AuthenticationAPIClient` client, it returns an instance of the `UserProfile` profile. This profile is a basic, OIDC-conformant profile, which guarantees only the `sub` claim. The `sub` claim contains the user's ID. Depending on the scope you requested, the remaining claims returned by the `UserProfile` are different. 
 
+When you get the `sub` value, call the [Management API](https://auth0.com/docs/api/management/v2#!/Users).
```
lbalmaceda: This line should be after the "user info snippet". Also I think the content is the same as line 83. "Use the `UsersAPIClient` client and the user's ID to get the full user profile"

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -94,12 +99,11 @@ usersClient.getProfile(userId)
         });
 ```
 
+## Access the Data Inside the `UserProfile` Profile
```
lbalmaceda: by skipping the `UserProfile` class you avoid saying profile twice. 

"## Access the Data Inside the received Profile"

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -94,12 +99,11 @@ usersClient.getProfile(userId)
...
 
-## Access The Data Inside The UserProfile
+### Default information
 
-##### I. DEFAULT INFO
-
-At this point, you already have access to the `UserProfile`.
+At this point, you already have access to the `UserProfile` profile.
```
lbalmaceda: Maybe we can replace "profile" with "instance".

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -110,48 +114,51 @@ profile.getEmail();
...
+### Additional information
 
-Besides the defaults, you can handle more information that is contained within any of the following `map`:
+You can handle additional information within any of the following maps:
 
-##### A. USER METADATA
+#### A. User metadata
 
-The `userMetadata` map contains fields related to the user profile that can be added from the client-side (e.g. when editing the profile). This tutorial explains how to achieve this. You can access its fields as follows:
+The `userMetadata` map contains user profile fields that can be added from the client-side session (e.g. when a user edits their profile). You can access this information in the following way:
```
lbalmaceda: "session" is a confusing word here as it might refer to a "previous state" (like a cart on a store's web page). Better if avoided.

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -110,48 +114,51 @@ profile.getEmail();
...
 
 ```java
 String country = (String) profile.getUserMetadata().get("country");
 ```
 
 ::: note
-The strings you use for subscripting the user_metadata map, and the variable types you handle, are up to you.
+You can choose the strings for subscripting the `user_metadata` map and the variable types you handle.
```
lbalmaceda: Maybe change a bit the wording to say something like "You can choose the name of the keys and its values types"

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -110,48 +114,51 @@ profile.getEmail();
...
+#### B. App metadata
 
-The `appMetadata` map contains fields that are usually added via a Rule or Hook, which is read-only for the native platform.
+The `appMetadata` map contains fields that are usually added with a [Rule](/rule) or a [Hook](/hooks). For native platforms, this information is read-only.
 
-##### C. EXTRA INFO
+#### C. Extra information
 
-The `extraInfo` map contains additional values stored in Auth0 but not mapped to a `UserProfile` getter method. That information is read-only for the native platform.
+The `extraInfo` map contains additional values that are stored in Auth0 but not mapped to the `UserProfile` getter method. For native platforms, this information is read-only.
```
lbalmaceda: the `UserProfile` class has "getters methods" to obtain specific claims, like "getUserID". The `extraInfo` map gathers all the claims that don't have an associated getter method. That said, I would keep the "to **a** `UserProfile` getter method."

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -18,22 +18,28 @@ This tutorial will show you how to link two different accounts for the same user
...
-* You're familiar with the concepts of `userId` and `idToken`. You can find info about them in the [Session Handling](/quickstart/native/android/03-session-handling) and [User Profile](/quickstart/native/android/04-user-profile) tutorials.
+::: panel Complete the previous tutorials
+Before you continue with this tutorial, make sure that you have completed the previous tutorials. This tutorial assumes that:
+* You have integrated [Auth0](https://github.com/auth0/Auth0.Android) as a dependency in your project. 
+* You are familiar with the `WebAuthProvider` class. To learn more, see the [Login](/quickstart/native/android/00-login) and the [Session Handling](/quickstart/native/android/03-session-handling) tutorials.
+* You are familiar with the concepts of `userId` and `idToken`. You can find info about them in the [Session Handling](/quickstart/native/android/03-session-handling) and the [User Profile](/quickstart/native/android/04-user-profile) tutorial.
```
lbalmaceda: tutorial -> tutorials

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -18,22 +18,28 @@ This tutorial will show you how to link two different accounts for the same user
...
 
-Here's the scenario: Your logged-in user wants to link one (or multiple) accounts to the account they are logged in with.
+Your users may want to link their other accounts to the account they are logged in to. 
+
+::: note
+Use the [Auth0 Android](https://github.com/auth0/Auth0.Android) library.
+:::
 
-To do this, we will use the Auth0 library as we did in the [Login](/quickstart/native/android/00-login) tutorial. In this case, we will send as an extra a boolean value to indicate that this is a secondary login, along with the `userId` obtained from the first login.
+To allow your users to link their accounts, in your requests, you need to send an additional boolean value. The boolean value tells the application that it is a secondary login, with the `userID` value obtained after the first login.
```
lbalmaceda: I would rewrite and clarify this: This boolean value is not "sent in a request" (people might think it's a HTTP request). The sample has 2 activities and to move from one to the other uses the android `Intent` class, which accepts key-value data. One of them is telling the "login activity" that it's going to be created in the "linking accounts mode" (telling nothing means it just wants to log in, typically at the start of the app), and the other one is telling which is the user id that it got that wants to merge with whatever userId is obtained next.

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -52,7 +58,7 @@ boolean linkSessions = getIntent().getExtras().getBoolean(Constants.LINK_ACCOUNT
 String userId = getIntent().getExtras().getString(Constants.PRIMARY_USER_ID);
 ```
 
-Then, in the login response we decide if we advance to the `MainActivity` as usual or return to the already instantiated one:
+In the login response, decide whether you want to move to  `MainActivity` as usual, or return to the already instantiated one:
```
lbalmaceda: I think that by "the already instantiated one" we meant "the current activity". It's confusing, I would just say:

"In the login response check the linkSessions boolean value to know if you need to show the MainActivity or continue to link the accounts."

Also, there are two whitespaces before the `MainActivity`

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -69,9 +75,11 @@ public void onAuthentication(Credentials credentials) {
 }
 ```
 
-## Link an Account
+## Link the Accounts
 
-Now we can link the accounts. You have a main user along with another account you want to link to that user. All we need is the `id` of the logged-in user and the `id_token` for the two accounts: the one we had previously saved and the one that we just received in the login response.
+Now, you can link the accounts. To do this, you need the logged-in user's ID and the ID tokens for the two accounts: 
+* The saved account the user initially logged in to
+* The second account received in the login response
```
lbalmaceda: make clear which response we mean: "The second account received in the **last** login response"

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -93,9 +101,9 @@ private void performLink(String secondaryIdToken) {
           });
 ```
 
-## Retrieve Linked Accounts
+## Retrieve the Linked Accounts
 
-The `AuthenticationAPIClient#userInfo` response doesn't include the identities array, but still you need to use it to obtain the user `id`. Then, by calling the `UsersAPIClient#getProfile` method you can obtain a user's full profile, which includes the linked accounts as an array of `UserIdentities`.
+To get the user's ID, use the `AuthenticationAPIClient#userInfo` response. Then, call the `UsersAPIClient#getProfile` method to obtain the user's full profile. The profile includes the linked accounts as the `UserIdentities` array.
```
lbalmaceda: I would rephrase this to avoid mentioning the "class#method". Maybe:
"Use the user ID (((((I don't mention from where I obtained it because it was used in the section above))))) to call the UsersAPIClient's `getProfile` method and obtain the full user's profile, which will now include the linked accounts in the `UserIdentities` array.

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -115,12 +123,17 @@ usersClient.getProfile(userInfo.getId())
 ```
 
 ::: note
-For more information, check the [UserIdentity.java](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/result/UserIdentity.java) class.
+For more information, check the [UserIdentity.java class documentation](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/result/UserIdentity.java).
```
lbalmaceda, line 118: remove `.java`

articles/quickstart/native/android/06-calling-apis.md
```diff
@@ -18,26 +18,40 @@ This tutorial demonstrates how to use a previously saved token to authenticate y
   ]
 }) %>__
 
-## Before Starting
+## Before You Start
 
-You should already know how to handle the `Credentials` object, as explained in the [Session Management](/quickstart/native/android/03-session-handling) tutorial.
+::: panel Complete the previous tutorials
+Before you continue with this tutorial, make sure that you have completed the previous tutorials. This tutorial assumes that:
+* You have completered the [Session Handling](/quickstart/native/android/03-session-handling tutorial and you know how to handle the `Credentials` object.
```
lbalmaceda: typo: completered -> completed

articles/quickstart/native/android/06-calling-apis.md
```diff
@@ -18,26 +18,40 @@ This tutorial demonstrates how to use a previously saved token to authenticate y
...
+* You have set up a backend application as API. To learn how to do it, follow one of the [backend tutorials](https://auth0.com/docs/quickstart/backend). 
+:::
 
-This sample assumes that you have already setup a backend application as API. If you haven't done so, you can follow any backend quickstart defined [here](https://auth0.com/docs/quickstart/backend). Then obtain the endpoint in which you're expecting the user to be first authenticated and declare it as a constant in the current class.
+After you set up an API, obtain the endpoint where your users authenticate. Declare the endpoint as constant in the current class:
```
lbalmaceda: "as **a** constant"

articles/quickstart/native/android/06-calling-apis.md
```diff
@@ -78,4 +88,6 @@ client.newCall(request).enqueue(new Callback() {
 });
 ```
 
-From here, check that the request was made and that the response came back as expected. You will need to configure your server-side to protect your API endpoints with the secret key for our Auth0 application. As in this example we're using the Auth0's issued `access_token`, you can use this same token to call Auth0 API's.
+Check if the request was made and if the response that came back was what you expected. 
+
+You need to configure your server side to protect your API endpoints with the secret key for our Auth0 application. In this example, you can use the user's access token issued by Auth0 to call Auth0 APIs.
```
lbalmaceda: I think we refer to "Auth0 applications" as "Auth0 clients" previously. Also, I would change: our Auth0 -> your Auth0

articles/quickstart/native/android/07-authorization.md
```diff
@@ -18,28 +18,34 @@ This step demonstrates how to use Auth0 to create access roles for your users. W
...
+Edit the following line from the default script to match the conditions that fit your needs:
 
 ```java
 if (user.email.indexOf('@example.com') > -1)
 ```
-to match the condition that fits your needs.
 
-By default, it says that if the user email contains `@example.com` then they will be given an `admin` role, otherwise a regular `user` role.
+The default rules for assigning access roles are:
+* If the user's email contains @example.com, the user gets the admin roles anything else, the user gets the regular user role
```
lbalmaceda: If the user's email contains `@example.com`, the user gets the admin roles. Anything else, the user gets the regular user role.

articles/quickstart/native/android/07-authorization.md
```diff
@@ -67,9 +73,9 @@ if (roles.contains("admin")) {
 ```
 
 ::: note
-Notice that you'll find the `roles` information within the `appMetadata` HashMap and not in the `userMetadata`. Application metadata cannot be modified by users, whereas User metadata can be.
+Application metadata cannot be modified by users but the user metadata can be modified. Because of that, the information about user roles is in the `appMetadata` HashMap, not in the `userMetadata` HashMap.
```
lbalmaceda: I would use:
"Application metadata cannot be modified by users but the user metadata can. Because of that, information about user roles is stored in the `appMetadata` Map"

articles/quickstart/native/android/_includes/_auth0.md
lbalmaceda: @awie3di We use `compile 'com.auth0.android:auth0:1.+'` in the articles to use the "latest major" and avoid coming here to update it on every release. If something related to the setup changes then we come and fix it. By adding this "you can check for the latest.." paragraph, people can check which is the "explicit" latest version so they don't use `1.+`

articles/quickstart/native/android/_includes/_auth0.md
```diff
@@ -19,10 +19,12 @@ dependencies {
...
 :::
 
-Now add the _Manifest Placeholders_, required by the SDK to define internally an **intent-filter** to capture the authentication callback. You do that by adding the next line:
+Add manifest placeholders required by the SDK. The placeholders define internally an `intent-filter` to capture the authentication callback URL.
```
lbalmaceda: placeholders are used to generate the intent-filter. (the library generates it). I would use:
"Add manifest placeholders required by the SDK. Placeholders are used internally to define an `intent-filter` that captures the authentication callback URL."

articles/quickstart/native/android/_includes/_login.md
```diff
@@ -1,11 +1,30 @@
...
+In the `login` method, create a new instance of the `Auth0` object to hold user credentials. 
 
-After calling `WebAuthProvider#start` the browser will launch and show **Lock**, and the final result will be received in the callback that you pass.
+You can use a constructor that receives an Android Context if you have added the following String resources: 
+* `R.string.com_auth0_client_id`
+* `R.string.com_auth0_domain`
+
+If you prefer to hardcode the resources, use the constructor that receives both strings. Then, use the `WebAuthProvider` class to authenticate with any connection you enabled on your client in the  [Auth0 dashboard](${manage_url}/#/).
+
+You need to make sure you get a reponse compliant with the OpenID Connect protocol. You can choose between two options:
```
lbalmaceda: typo: reponse -> response

articles/quickstart/native/android/_includes/_login.md
```diff
@@ -1,11 +1,30 @@
...
+
+* Set the audience
+* Turn on the **OIDC conformant** switch in your Auth0 dashboard
+  		  
+::: note
+To turn on the **OIDC conformant** switch, in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings), click on **Show Advanced Settings** > **OAuth**. To learn more, read the [net flows documentation](/api-auth/intro#how-to-use-the-new-flows).
+:::
+
+After you call the `WebAuthProvider#start` function, the browser launches and shows the **Lock** widget. The callback URL contains the final result of the authentication process.
```
lbalmaceda: "After you call the `WebAuthProvider#start` function, the browser launches and shows the **Lock** widget. Once the user authenticates, the callback URL will be called with the final result of the authentication process. "

articles/quickstart/native/android/_includes/_login.md
```diff
@@ -38,13 +57,23 @@ private void login() {
 
 ## Capture the Result
 
-You'll first need to whitelist the **Callback URL** in the "Allowed Callback URLs" section of the [Client settings](${manage_url}/#/clients) by adding the URL below. Remember to replace `YOUR_APP_PACKAGE_NAME` with your actual application's package name, available in the `app/build.gradle` file as the `applicationId` attribute:
+Whitelist the callback URL for your app in the **Allowed Callback URLs** section in [Client settings](${manage_url}/#/clients). In that section, enter the following URL:
```
lbalmaceda, line 64: Replace Client Settings URL with `${manage_url}/#/applications/${account.clientId}/settings`

articles/quickstart/native/android/_includes/_login.md
```diff
@@ -70,12 +99,7 @@ The browser will redirect to your application with the authentication result. Be
...
 ```
 
-There are many options to customize the authentication using the `WebAuthProvider` builder. Make sure to check them [here](/libraries/auth0-android#implementing-web-based-auth).
+There are many options to customize the authentication with the `WebAuthProvider` builder. You can read about them in the [Auth0 SDK for Android documentation](/libraries/auth0-android).
 <div class="phone-mockup">
   <img src="/media/articles/native-platforms/android/login-android.png" alt="Mobile example screenshot" />
-</div>
-
-
-## Hosted Login Page vs Embedded Login
```
lbalmaceda: @aaguiarz do we want to remove this section? Is now mentioned here in the "start" step, which IMO can get more attention than before (which might be bad) https://github.com/auth0/docs/pull/5017/files#diff-4bdb4aedb0f37d8c3116bbdfcff71d65R6

articles/quickstart/native/android/getting-started.md
```diff
@@ -4,22 +4,22 @@ description: Short Introduction to the Auth0 Android Quickstarts.
...
 
 ## Dependencies
 
-Each tutorial will require you to use the [Auth0.Android](https://github.com/auth0/Auth0.Android) library. This is a toolkit that lets you communicate with many of the basic [Auth0 API](https://auth0.com/docs/api) functions in a neat way.
+Each tutorial uses the [Auth0.Android](https://github.com/auth0/Auth0.Android) library. It is a toolkit that lets you communicate with many of the basic [Auth0 API](https://auth0.com/docs/api) functions in a neat way.
 
 ## Create a Client
 
-If you haven't already done so, create a new client application in your [Auth0 dashboard](${manage_url}/#/applications/${account.clientId}/settings) and choose **Native** for the Type.
+Create a new client application in your [Auth0 dashboard](${manage_url}/#/applications/${account.clientId}/settings). For **Type**, select **Native**.
```
lbalmaceda: Replace the Auth0 Dashboard URL to `${manage_url}/#/clients` as this one needs to be the clients/create clients page

articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -115,12 +123,17 @@ usersClient.getProfile(userInfo.getId())
...
-### Unlink An Account
+### Unlink the Accounts
+
+To unlink the accounts, you need to specify the following: 
+* user ID for the main account
+* user ID for the linked acocunt
+* the provider name
 
-The unlink process is similar to the linking one, the only difference being that you need to specify both the two `user id``s and the `provider name` to unlink the connections. Additionally, the token used to instantiate the `UsersAPIClient` must be the `id_token` of the main identity.
+To instantiate the `UsersAPIClient` client, use the ID token for the  main account.
```
lbalmaceda: There are two whitespaces before the "main account"

articles/quickstart/native/android/06-calling-apis.md
```diff
@@ -50,13 +64,9 @@ Request request = new Request.Builder()
...
 ```
 
-::: note
-Notice that how you configure your authorization header should match the standards that you're using in your API, this is just an example of what it could look like.
-:::
-
 ## Send the Request
 
-Finally we tell the client to create a new `Call` with the given request, and then invoke `enqueue` to execute the request asynchronously.
+Tell the client to create a new `Call` with the request you created.  Call the `enqueue` function to execute the request asynchronously.
```
lbalmaceda: There are two whitespaces before the "Call"

articles/quickstart/native/android/07-authorization.md
```diff
@@ -18,28 +18,34 @@ This step demonstrates how to use Auth0 to create access roles for your users. W
...
+
+## Create a Rule to Assign Roles
 
-## Create A Rule To Assign Roles
+Create a rule that assigns the following access roles to your user: 
+* An admin role
+* A regular user role
 
-First, you need to create a rule that assigns your users either an `admin` role, or a single `user` role. To do so, go to the [new rule page](${manage_url}/#/rules/new) and select the "*Set Roles To A User*" template, under *Access Control*. Then, replace this line from the default script:
+To assign roles, go to the [New rule](${manage_url}/#/rules/new) page. In the **Access Control** section,  select the **Set roles to a user** template.
```
lbalmaceda: There are two whitespaces before the "select the"

articles/quickstart/native/android/_includes/_callback_urls.md
```diff
@@ -2,12 +2,10 @@
...
-
-This tutorial series make use of a **Callback URL** that looks like this:
+This tutorial series uses a callback URL that looks like this:
 
 ```js
 demo://${account.namespace}/android/YOUR_APP_PACKAGE_NAME/callback
 ```
 
-Remember to replace `YOUR_APP_PACKAGE_NAME` with your actual application's package name, available in the `app/build.gradle` file as the `applicationId` attribute.
+Replace `YOUR_APP_PACKAGE_NAME` with your application's package name, available as the `applicationId` attribute  in the `app/build.gradle` file.
```
lbalmaceda: There are two whitespaces before the "in the app/build"

articles/quickstart/native/android/_includes/_login.md
```diff
@@ -1,11 +1,30 @@
...
+:::
 
+In the `login` method, create a new instance of the `Auth0` object to hold user credentials. 
 
-After calling `WebAuthProvider#start` the browser will launch and show **Lock**, and the final result will be received in the callback that you pass.
+You can use a constructor that receives an Android Context if you have added the following String resources: 
+* `R.string.com_auth0_client_id`
+* `R.string.com_auth0_domain`
+
+If you prefer to hardcode the resources, use the constructor that receives both strings. Then, use the `WebAuthProvider` class to authenticate with any connection you enabled on your client in the  [Auth0 dashboard](${manage_url}/#/).
```
lbalmaceda: There are two whitespaces before the `Auth0 dashboard`

articles/quickstart/native/android/getting-started.md
```diff
@@ -4,22 +4,22 @@ description: Short Introduction to the Auth0 Android Quickstarts.
 budicon: 715
 ---
 
-This multistep quickstart guide will walk you through managing authentication in your android apps with Auth0.
+This quickstart guide walks you through managing authentication  with Auth0 in your Android apps.
```
lbalmaceda: There are two whitespaces before the "with Auth0"

lbalmaceda (changes requested):

articles/quickstart/native/_includes/_callback-url-introduction.md
aaguiarz: I think we need to use this one https://github.com/auth0/docs/blob/master/articles/_includes/_callback_url.md in the Native ones too.

I'll change it so it communicates some of the information that is in the Native ones.

aaguiarz: Actually, I think that one is OK, we should use that one.

articles/quickstart/native/android/03-session-handling.md
aaguiarz: I think title case you capitalize "You" but not "the".

articles/quickstart/native/android/03-session-handling.md
aaguiarz: I think "direct" them is OK here, because it's a mobile app, so there isn't any "Redirection"

articles/quickstart/native/android/03-session-handling.md
aaguiarz: I think we are not verifying a token by calling the API, so I'd leave the change as is

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -123,26 +130,32 @@ aClient.userInfo(accessToken)
...
-
-We will use the previously saved `refresh_token` to get a new `access_token`. It is recommended that you read and understand the [refresh tokens documentation](/refresh-token) before proceeding. For example, you should remember that even though the refresh token cannot expire and must be securely saved, it can be revoked. Also note that the new pair of credentials will never have additional scope than the requested in the first login.
+## Refresh the User's Session
 
+::: panel
+Before you go further with this tutorial, read the [refresh token documentation](/refresh-token).
+It is important that you remember the following:
+* Even though the refresh token cannot expire and must be securely saved, it can be revoked. 
+* The new pair of credentials will never have a different scope than the scope you requested during the first login.
+:::
```
aaguiarz: I will add 3 lines:

- Refresh tokens must be securely saved
- Even though refresh tokens cannot expire, they can be revoked.
- New tokens will never have a different scope than the scope you requested during the first login.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -123,26 +130,32 @@ aClient.userInfo(accessToken)
...
 
-First instantiate an `AuthenticationAPIClient`:
+Create an instance of the `AuthenticationAPIClient` object:
```
aaguiarz: "Create an AuthenticationAPIClient instance'

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -183,9 +195,11 @@ private void logout() {
 ```
 
 ::: note
-Deleting the user credentials depends on how you have stored them.
+Depending on the way you store users' credentials, you delete them differently. 
 :::
 
-### Optional: Encapsulated session handling
+### Optional: Encapsulate Token Handling
```
aaguiarz: I think "Session" is better here

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -183,9 +195,11 @@ private void logout() {
...
+
+Handling users' sessions is not a straightforward process. You can simplify it by storing token-related information and processes in a class. The class needs to separate the logic for handling users' sessions from the activity.
```
aaguiarz: "The class will separate the logic for handling users' session from the activity"

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -18,11 +18,13 @@ This tutorial will show you how to get and modify the user's profile data in you
...
 
-Be sure that you have completed the [Login](/quickstart/native/android/00-login) and the [Session Handling](/quickstart/native/android/03-session-handling) Quickstarts. You'll need a valid `access_token` and `id_token` to call the API clients.
+::: note
+Before you continue with this tutorial, make sure that you have completed the [Login](/quickstart/native/android/00-login) and the [Session Handling](/quickstart/native/android/03-session-handling) tutorial. To call the API clients, you need a valid access token and ID token.
+:::
 
-Before launching the log in you need to ask for the `openid profile email` scopes in order to get a valid profile in the response. Locate the snippet were you're initializing the `WebAuthProvider` and add the `withScope("openid profile email")` line.
+Before launching the login process, you need to make sure you get a valid profile in the response. To do that, ask for the `openid profile email` scope. Find the snippet in which you initialize the `WebAuthProvider` class. To that snippet, add the line `withScope("openid profile email")`.
 
 ```java
```
aaguiarz: instead of 
"you need to make sure you get a valid profile in the response"

say
"you need to make sure you will get a valid profile from the authorization server"

articles/quickstart/native/android/04-user-profile.md
aaguiarz: Maybe

"
To get the user information:

- Instantiate the AuhenticationAPIClient using the 'access_token'. 
- Call the AuthenticationAPIClient's userInfo method, it returns an instance of the UserProfile profile.  The profile will be OIDC conformant and its contents will depend on the [scopes] (https://auth0.com/docs/scopes/current) you requested when logging. 

- To return the full Auth0 User Profile you can use the [Management API](https://auth0.com/docs/api/management/v2#!/Users).

articles/quickstart/native/android/04-user-profile.md
aaguiarz: Yes, removing 'session' there is fine.

articles/quickstart/native/android/05-linking-accounts.md
aaguiarz: "To allow your users to link their accounts you need to specify an additional boolean value which tells the application that it is a secondary login, in addition to the `userID` value obtained after the first login."

articles/quickstart/native/android/_includes/_login.md
aaguiarz: Yes, we should remove the section, it is already mentioned. We are showing it in the start steps in all SPA quickstarts, so it's fine.

articles/quickstart/native/android/_includes/_auth0.md
awie3di: @lbalmaceda, is it something that your readers know? Or can we somehow indicate that in the note? Not sure if I understood your comment, if you could help me clarify it for the first-time readers: "You can check for the latest version of X on the ..." Thanks :)

articles/quickstart/native/_includes/_callback-url-introduction.md
awie3di: @aaguiarz just to make sure, should I replace the **articles/quickstart/native/_includes/_callback-url-introduction.md** snippet with the **articles/_includes/_callback_url.md** one with no additional changes?

articles/quickstart/native/android/03-session-handling.md
awie3di: An "instance" is usually of something. In the previous quickstarts, I usually tried to clarify *of what* is the created instance. Is AuthenticationAPIClient not an object? (sorry, I tried to guess it from the context and some of the code but it's not always clear to me)

articles/quickstart/native/android/04-user-profile.md
awie3di: I will remove the word 'session', however, I'm not sure how to rephrase the sentence. 
"the client-side" works only if there is a noun after. Would "that can be added when a user edits their profile" be OK? As I'm not sure if just  "from the client's side" would be clear enough. @aaguiarz, could you advise?

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -110,48 +114,51 @@ profile.getEmail();
...
-The `userMetadata` map contains fields related to the user profile that can be added from the client-side (e.g. when editing the profile). This tutorial explains how to achieve this. You can access its fields as follows:
+The `userMetadata` map contains user profile fields that can be added from the client-side session (e.g. when a user edits their profile). You can access this information in the following way:
 
 ```java
 String country = (String) profile.getUserMetadata().get("country");
 ```
 
 ::: note
-The strings you use for subscripting the user_metadata map, and the variable types you handle, are up to you.
+You can choose the names and value types of the keys you use for subscripting the `user_metadata` map.
```
awie3di: @lbalmaceda Can you take a look at the revised note? Not sure if I correctly understood what you meant

articles/quickstart/native/android/05-linking-accounts.md
awie3di: @lbalmaceda, is this a style requirement? UserIdentity.java is the exact name of the element.

articles/quickstart/native/android/04-user-profile.md
awie3di: An "instance" is usually of something. In the previous quickstarts, I usually tried to clarify **of what** is the created instance. Is there some other word I could use to replace "profile"?

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -49,11 +51,15 @@ AuthenticationAPIClient authClient = new AuthenticationAPIClient(auth0);
...
 
 ::: note
-It's suggested that you add both the Auth0 `domain` and `clientId` to the `strings.xml` file rather than hardcode them.
+Do not hardcode the Auth0 `domain` and `clientId` values. We recommend you add them to the `strings.xml` file.
 :::
 
-Next, use the `access_token` to obtain the user id with the `AuthenticationAPIClient`. Although the call returns a `UserProfile` instance, this is a basic OIDC conformant profile and the only guaranteed claim is the `sub` which contains the user's id, but depending on the requested scope the claims returned may vary. With the `sub` value call the [Management API](https://auth0.com/docs/api/management/v2#!/Users) and get a complete user profile back.
+To get the user's information:
 
+1. Use the user's access token to create an instance of the `AuhenticationAPIClient` client.
```
awie3di: @aaguiarz, could you take a look at the revised instructions to see if they're ok?

articles/quickstart/native/android/04-user-profile.md
lbalmaceda: A map/dictionary has a `KEY` and a `VALUE`. The Key is always a String (this is how you want to call it) and the Value can be of many types like a String, Integer, Double, Boolean, Object. i.e.:

```json
profile: {
  "user_metadata":{
    "favorite_quote": "Light in Nature creates the movement of colors.",
    "siblings": 2,
    "money_in_bank": 999.34,
    "is_alive": true,
    "car": {
      "color": "black",
      "wheels": 4,
      "brand": "toyota"
    }
  },
  //More Stuff
}
```

The `user_metadata` is itself a Key in the user's profile, that has a nested Object (more key-values) as the Value. What we want to tell in the above note is that the user can customize the `user_metadata` contents however they want.

articles/quickstart/native/android/04-user-profile.md
lbalmaceda: The client itself doesn't require an `access_token` to be instantiated. What we meant was that we're going to use an instance of the API client to call one of its methods (userInfo) that receives an `access_token` and returns the OIDC-conformant profile of the user associated with that token.

articles/quickstart/native/android/04-user-profile.md
lbalmaceda: In this case, it's an instance of "UserProfile" (the class). I don't think there's another word, but removing that word will take you back from where you started.

articles/quickstart/native/android/03-session-handling.md
lbalmaceda: 1 and 2 can be combined. (2 is the reason why you should do 1).

articles/quickstart/native/android/03-session-handling.md
lbalmaceda: `AuthenticationAPIClient` is the main class that exposes Auth0's authentication API endpoints. In a programmers context, calling "AuthenticationAPIClient client = new AuthenticationAPIClient()" would mean creating a new instance of that class and assigning it to the "client" variable. That's what we mean by "instance". Some of the methods we call return something of our interest, like a pair of credentials. In that specific case, note we're talking about "Credentials" with the capital C ([link](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/result/Credentials.java)). Those are classes that are instantiated by our SDKs and the user can use later.

articles/quickstart/native/android/_includes/_auth0.md
lbalmaceda: Using version `x.+` means use major `X` but latest minor/patch. The problem with fixing it to `x.y` is that each time we release a new minor/patch we need to come here and update the file, otherwise the users will copy paste the version number and keep using the old documented version. This is fine as we follow semver and we don't break stuff in each minor/patch. If that happens, then we will come here and edit the file to say "use at least _this_ new version". By adding a paragraph saying that the explicit last version can be found in the repo url we save from manually editing the file on each release. (we can't automate this yet).

articles/quickstart/native/android/05-linking-accounts.md
lbalmaceda: By saying "{NAME} class" you're already implying that it's a java file. I think it's redundant to keep it, but that's just my opinion.

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -49,11 +51,13 @@ AuthenticationAPIClient authClient = new AuthenticationAPIClient(auth0);
 ```
 
 ::: note
-It's suggested that you add both the Auth0 `domain` and `clientId` to the `strings.xml` file rather than hardcode them.
+Do not hardcode the Auth0 `domain` and `clientId` values. We recommend you add them to the `strings.xml` file.
```
lbalmaceda, line 54: I'd remove this note as we already covered this suggestion in the getting-started and login tutorials.

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -110,48 +115,51 @@ profile.getEmail();
 profile.getPictureURL();
 ```
 
-::: note
-Remember that you can't modify the UI inside the onSuccess() method, as it works in a second thread. To solve this, you can persist the data, create a task in the UI thread or create a handler to receive that information.
+::: panel Modifying the UI
+You cannot modify the UI inside the `onSuccess()` method because the method works in a second thread. To solve this issue, you can choose between three options:
```
lbalmaceda, line 119: I'd not say "modify" as you sometimes just _interact_ with it, like changing a inner state rather than a visual component. i.e. calling `view.setEnabled(true)` will fail as well. The suggestion applies both for this paragraph and the title.

lbalmaceda (changes requested):

PR #5018, opened 2017-09-19
Update aditional-parameter description

mjlescano:
Sometimes, it may be misleading to list `aditional-parameter` as a parameter itself.

In my case I tried to add, for example, `?...aditional-parameter=login_hint=asd@asd.com`, instead of directly adding it to the request as `?...login_hint=asd@asd.com`.

Here's a case that someone had the same misconception: https://community.auth0.com/questions/2984/additional-parameter-on-authorize-does-not-work


mpaktiti:
Hi @mjlescano , thank you for fixing that and apologies for taking so long to review and merge it! 👍

PR #5026, opened 2017-09-21
Update state parameter as required

hlibco:
Not sending the state will make the request to fail


mpaktiti:
Thank you @hlibco ! 👍

PR #5027, opened 2017-09-21
Lock-Android docs updates

jeffreylees:
Lining up versioned documents to match each other more closely. One of the biggest changes was combining the native social auth docs for v2 - it makes for a rather long form doc, but that's ok in my book - the versioning matches up better, and to be honest the whole mess is deprecated anyway, so the more consolidated it is, the better.

Also a linting and copyediting improvements across the board.


articles/libraries/lock-android/v1/configuration.md
```diff
@@ -1,15 +1,15 @@
 ---
 description: Configuration options and methods for Lock for Android v1
 ---
-## Lock v1 Configuration
+# Lock v1 Configuration
```
mpaktiti: Why are the titles different? v1 is `Lock v1 Configuration` while v2 is `Lock Configuration Options`.
I'd expect to see `Lock Configuration Options` in both titles since the version is displayed directly after.

articles/libraries/lock-android/v1/internationalization.md
```diff
@@ -0,0 +1,37 @@
+---
+section: libraries
+toc_title: Internationalization
+title: Internationalization
+description: Internationalization support in Lock for Android
+---
+# Internationalization in Lock for Android v1
```
mpaktiti: Here the titles are aligned but the version is not needed. 

Also, at the previous title was `Lock v1 Configuration` while this is `Internationalization in Lock for Android v1` (they are both for Android though).

Pick one format and use it in all titles.

articles/libraries/lock-android/v1/internationalization.md
```diff
@@ -0,0 +1,37 @@
+---
+section: libraries
+toc_title: Internationalization
```
mpaktiti: Can you tell me what this metadata does? I have seen it in some docs, mainly about libraries, (without the `toc` metadata, like this one https://github.com/auth0/docs/blob/master/articles/libraries/lock-ios/v1/logging.md) but I don't get what it does.

articles/libraries/lock-android/v1/native-social-authentication.md
```diff
@@ -2,7 +2,7 @@
 toc_title: Native Social Authentication
 description: How to implement native social authentication with Lock Android
 ---
-# Lock Android: Native Social Authentication
+# Native Social Authentication
```
mpaktiti: Again pointing out the thing with the headers. Here you don't have either a version or any mention of `Lock Android`.

articles/libraries/lock-android/v1/passwordless-magic-link.md
```diff
@@ -2,14 +2,16 @@
 toc_title: Passwordless with Magic Link
 description: Passwordless with Magic Link with Lock Android
 ---
-# Lock Android: Passwordless with Magic Link
+# Passwordless with Lock Android
```
mpaktiti: Headers do not match.

Also, some of these docs are quite long but there is no toc.

articles/libraries/lock-android/v2/custom-authentication-providers.md
```diff
@@ -3,17 +3,23 @@ section: libraries
 toc_title: Custom Authentication Providers
 description: Implementing custom authentication providers
 ---
-
 # Custom Authentication Providers
```
mpaktiti: This does not have a v1 version so it leads to the main doc https://auth0-docs-content-pr-5027.herokuapp.com/docs/libraries/lock-android/v1

articles/libraries/lock-android/v2/custom-authentication-providers.md
```diff
@@ -3,17 +3,23 @@ section: libraries
...
 ## The AuthProvider class
+
 Create a class that implements the `AuthProvider` interface and override its methods.
 
-You can also use any _Native_ implementation already provided by Auth0. Currently available are [Google](/libraries/lock-android/native-provider-google) and [Facebook](/libraries/lock-android/native-provider-facebook).
+You can also use any _Native_ implementation already provided by Auth0. The [native social authentication](/libraries/lock-android/v2/native-social-authentication) providers currently available are Google and Facebook.
+
+::: warning
+The native social providers rely on a deprecated grant type. Clients created after June 8th 2017 won't be able to use this feature.
+We recommend using browser-based flows, as explained in [Authentication with Auth0 Hosted Login Page](/libraries/auth0-android#authentication-with-auth0-hosted-login-page).
```
mpaktiti: Avoid using whitespace inside notes/warnings/panel, the end result doesn't look good.

In this case, it's ignored either way
image

articles/libraries/lock-android/v2/custom-fields.md
```diff
@@ -3,18 +3,18 @@ section: libraries
 toc_title: Custom Fields at Signup
 description: Adding additional fields to signups with Lock for Android
 ---
-
 # Lock for Android Custom Fields at Signup
```
mpaktiti: Doesn't have a matching v1 doc

articles/libraries/lock-android/v2/custom-oauth-connections.md
```diff
@@ -2,7 +2,6 @@
 section: libraries
 description: Lock for Android - Custom OAuth Connections
 ---
-
 # Custom OAuth Connections
```
mpaktiti: Doesn't have a matching v1 doc

articles/libraries/lock-android/v2/custom-theming.md
```diff
@@ -3,12 +3,12 @@ section: libraries
 toc_title: Custom Theming
 description: Customizing the Lock for Android UI
 ---
-
 # Lock for Android Custom Theming
```
mpaktiti: Doesn't have a matching v1 doc

articles/libraries/lock-android/v2/keystore.md
```diff
@@ -4,7 +4,6 @@ toc_title: Android Development Keystores and Key Hashes
 title: Android Development Keystores and Key Hashes
 description: Instructions on acquiring development keystores/key hashes during Android app development.
 ---
-
 # Android Development Keystores and Key Hashes
```
mpaktiti, line 7: Doesn't have a matching v1 doc

mpaktiti (changes requested):
Some comments, mainly on aligning headers for this library's docs

jeffreylees:
Yeah, we are probably not creating a lot of v1 content to match what there is in v2 - this is an incremental improvement, not a total fix for the situation.

The custom-theming one might could be created from what is already there.

Will pick a heading format and alter them to match each other, and add TOCs.

articles/libraries/lock-android/v1/internationalization.md
jeffreylees: It was for the sort of "related docs" section used on some older docs - see the `/libraries/lock-ios/v1/index.md` - that's the only place I see still using it, while I'm in here I think I'll remove that, add a list of "related topics" or next steps, and we can strip that metadata item out anywhere it is.

jeffreylees:
Updated title to all have a matching format, without versions in them.

Updated metadata to remove all `toc_title`, and to include `title` in all of them. These do include the version, for search results purposes. 

Removed error-messages doc from v1, as we have `/libraries/error-messages` now that is more exhaustive. Also removed an unnecessary redirect.

jeffreylees:
I'd like to locate `keystore` and `migration-guide` outside of the versioned folders, but it appears that locating items inside of `lock-android` but not inside of `v1` or `v2` doesn't work, seems to mess everything up. Can't navigate to items outside of versioned folders from the internal nav sidebar, even when sidebar links are changed, but can navigate to them by direct URL. Once on such a page, the entire set of sidebar links don't appear to work.

mpaktiti:
@jeffreylees Yes, you can't do that. You would have to move this outside the `lock-android` folder. Are you going to push more changes or should I review this?

jeffreylees:
I might have one more commit later - I can take the label off and ping you when it's good to be reviewed again so that you you'll know.

jeffreylees:
@mpaktiti ready for your review

PR #5032, opened 2017-09-21
Update extensions.md

sabues:
Linking to centralized complete certificates characteristics


mpaktiti:
Hey @sabues, I updated the URL, removing the `https://auth0.com/docs` part (we use relative URLs in docs, so we only include everything after `https://auth0.com/docs`). Thanks!

PR #5034, opened 2017-09-21
[edit]: updated per recent changes

ndelgado:


sabues (commented):
I would replace
```This mirrors how the cloud handles webtask domains and enables safely using extensions in multi-tenant environments```
with  
```This enables safely using extensions in multi-tenant environments in the same manner as in our multi-tenant cloud offering``` or something similar.

PR #5040, opened 2017-09-24
Fixes to hooks examples

thameera:
Fixes the following:
* The error message in pre-reg hook doesn't match with message in code
* Misleading Slack URL in post-reg hook example
* `cb()` function in post-reg hooks don't take any arguments, but the
code in example does


mpaktiti:
Thank you @thameera ! 👍

PR #5054, opened 2017-09-26
Update Auth API user metadata description

khorne3:
Fixing description and example of user metadata so that it displays the correct requirements for that field.

https://auth0-docs-content-pr-5054.herokuapp.com/docs/api/authentication

### DIFF
--- a/articles/api/authentication/_sign-up.md
+++ b/articles/api/authentication/_sign-up.md
@@ -10,7 +10,7 @@ Content-Type: 'application/json'
   "email": "EMAIL",
   "password": "PASSWORD",
   "connection": "CONNECTION",
-  "user_metadata": { plan: 'silver', team_id: `a111` }
+  "user_metadata": { plan: 'silver', team_id: 'a111' }
 }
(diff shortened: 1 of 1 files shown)


mpaktiti (commented):
Hey Katie, here https://github.com/auth0/docs/pull/5054/files#diff-a6fe6efcdca017ba19de251d590255a0R13 you use `a111` while here https://github.com/auth0/docs/pull/5054/files#diff-a6fe6efcdca017ba19de251d590255a0R38 you use 'a111'. I suppose the fix was because team_id is a string, and the first case should be updated to single quotes?

khorne3:
@mpaktiti I think there were two things:

* I changed 111 to a111 since the latter makes more sense as a strong
* I removed the single quotes (now, from both) since they're objects

Sorry for the confusion!

mpaktiti:
@khorne3 I pushed a commit to have `team_id: 'a111'` in all 3 snippets. Have a look and let me know if sth is wrong. Thnx!

PR #5061, opened 2017-09-26
Fixed Delegated Admin Filter Hook docs

amweiss:
Search syntax fails with colon syntax and has to use equals syntax.

I didn't go through wordy for this, but can if needed.

### DIFF
--- a/articles/extensions/delegated-admin/hooks.md
+++ b/articles/extensions/delegated-admin/hooks.md
@@ -123,7 +123,7 @@ function(ctx, callback) {
   }
 
   // Return the lucene query.
-  return callback(null, 'app_metadata.department:"' + department + '"');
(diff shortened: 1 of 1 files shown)


khorne3:
Syntax is correct as is. `app_metadata` is [no longer searchable](https://auth0.com/docs/api/management/v2/user-search#searchable-fields); will update docs to reflect changes to the extension.

amweiss:
The equals syntax searching worked fine for `app_metadata` so you may want to have someone look into that if it shouldn't be searchable anymore.

khorne3:
@amweiss Was your tenant created prior to 9/1/17? If so, that's correct.

amweiss:
@khorne3 nope it was after 9/1, it was closer to 9/20. I still have 13 days left on my trial.

PR #5062, opened 2017-09-27
Lock web updates

jeffreylees:


articles/libraries/lock/v10/auth0js.md
```diff
@@ -49,8 +45,8 @@ If you need further detail about usage, check out the [Auth0.js v8 Reference](/l
 If you included the Lock script from the Auth0 CDN, you will need to also include the auth0.js script before Lock:
 
 ```html
-<script src="https://cdn.auth0.com/w2/auth0-7.6.1.min.js"></script>
-<script src="https://cdn.auth0.com/js/lock/10.9.0/lock.min.js"></script>
+<script src="${auth0js_url}"></script>
+<script src="${auth0js_urlv8}"></script>
```
mpaktiti: This includes auth0.js twice

articles/libraries/lock/v10/auth0js.md
```diff
@@ -1,16 +1,12 @@
...
 
-## Changes with newer versions
-
-If you try to use auth0.js along with Lock 10, you will not be able to call `getClient()`. Instead, you should include its dependency and instantiate the object.
-
-In the past, a version of auth0.js was included automatically with Lock; at this point, you must include auth0.js yourself, allowing you to choose the supported version of auth0.js that best suits the needs of your application - version 7, or version 8. Note that the two versions have different CDN paths, as well as slightly different instantiations.
+For simple uses, Lock is all that is necessary. However, while using Lock, if more customization is required in an application than Lock allows, functionality from the Auth0.js SDK can be used alongside Lock to meet those needs. An example might be using Lock to handle signups and logins, while using auth0.js to [manage users](/libraries/auth0js#user-management) (read and update user metadata, link user accounts together, and similar tasks).
 
 ## Using auth0.js v8
```
mpaktiti: Can you update this section, so that you use this Before/After UI? I think it will simplify the doc. The captions would be something like `auth0.js v7/auth0.js v8`

articles/libraries/lock/v10/auth0js.md
```diff
@@ -19,8 +15,8 @@ In the past, a version of auth0.js was included automatically with Lock; at this
 If you included the Lock script from the Auth0 CDN, you will need to also include the auth0.js script before Lock:
```
mpaktiti: We say `you will need to also include the auth0.js script before Lock`. Is this an actual requirement? That auth0.js should be included before Lock, otherwise sth won't work? If so I'm thinking maybe we need to make it more prominent.

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
 section: libraries
 description: Details about Popup Mode with Lock V10.
 ---
-# Popup Mode
+# Authentication Modes
```
mpaktiti: Most of the Lock docs start with `Lock: `. So in order to keep them consistent this should be `Lock: Authentication Modes` 

(tbh I'm not that sure about that `:` character. Looks like sth we do in lack of better navigation, like breadcrumbs. Don't change it for now, but we should look into this at some point)

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
...
 
-## The Default: Redirect Mode
+## Redirect Mode
```
mpaktiti: I would like to see a couple of sentences before this paragraph, explaining what this doc is about

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
 section: libraries
 description: Details about Popup Mode with Lock V10.
```
mpaktiti: `description` needs updating

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
...
+If after you click on the IdP button (Facebook for example), a popup (new tab or window) is opened, it means you are using popup mode. In that popup, you'll see that Facebook page is displayed. Once you successfully login to Facebook, the popup will be closed and your web app will recognize that the user has been authenticated. The web app has **never been redirected to any other page**.
 
 ::: warning
-There is a known bug that prevents popup mode from functioning properly in Android or Firefox on iOS, and in Internet Explorer under certain circumstances. As such we recommend either only using redirect mode or detecting these special cases and selectively enabling redirect mode. See more info [here](https://ask.auth0.com/t/popup-login-window-is-not-closed-after-authentication/2843).
+There is a known bug that prevents popup mode from functioning properly in Android or Firefox on iOS, and in Internet Explorer under certain circumstances. As such we recommend either only using redirect mode or detecting these special cases and selectively enabling redirect mode.
 :::
 
 ![Widget Popup](/media/articles/libraries/lock/v10/widget-popup.gif)
```
mpaktiti: I love the gif but Lock is VERY outdated. Can you create and add a new one?

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
...
 description: Details about Popup Mode with Lock V10.
 ---
-# Popup Mode
+# Authentication Modes
 
-## The Default: Redirect Mode
+## Redirect Mode
 
-If after you click on the IdP button (Facebook for example), the web app you built gets redirected to Facebook, it means you're using Redirect Mode. Redirect Mode is the default with Lock 10, and is the recommended mode for almost all use cases. Once you successfully login to Facebook, Facebook will redirect you back to your app (through Auth0).
+If after you click on the IdP button (Facebook for example), the web app you built gets redirected to Facebook, it means you're using redirect mode. Redirect Mode is the default with Lock 10, and is the recommended mode for almost all use cases. Once you successfully login to Facebook, Facebook will redirect you back to your app (through Auth0). The majority of examples or samples in the reference documentation employ redirect mode, as it is the default mode for Lock 10.
```
mpaktiti: Add GIF for redirect mode

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
...
```
mpaktiti: I think you should re-write this to simplify the sentences a bit. You can start with describing the behavior (`When you use redirect mode.... this happens`). The example you should use ideally should match the gif (so if the gif clicks on Facebook then use Facebook in the example).

mpaktiti: The very last part of the last sentence (`as it is the default mode for Lock 10`) just repeats what we have already told them (`Redirect Mode is the default with Lock 10`).

articles/libraries/lock/v10/authentication-modes.md
```diff
@@ -2,23 +2,23 @@
...
 
 ::: warning
-There is a known bug that prevents popup mode from functioning properly in Android or Firefox on iOS, and in Internet Explorer under certain circumstances. As such we recommend either only using redirect mode or detecting these special cases and selectively enabling redirect mode. See more info [here](https://ask.auth0.com/t/popup-login-window-is-not-closed-after-authentication/2843).
+There is a known bug that prevents popup mode from functioning properly in Android or Firefox on iOS, and in Internet Explorer under certain circumstances. As such we recommend either only using redirect mode or detecting these special cases and selectively enabling redirect mode.
 :::
 
 ![Widget Popup](/media/articles/libraries/lock/v10/widget-popup.gif)
 
-Implementing Lock with Popup Mode is again a simple change of an option from its default.
+Implementing Lock with Popup Mode is again a simple change of the `redirect` option from its default.
```
mpaktiti: Nice. I would leave only the `Auth0Lock` part (remove the `lock.on` since from what I can see it has no differences with the redirect mode)

mpaktiti: Can you copyedit the paragraph at `## Database connections and popup mode`? Some of the english need fixing, and there is a `cross-origin requests` link pointing to https://auth0-docs-content-pr-5062.herokuapp.com/docs/api/authentication#resource-owner I also think the note at the very end is not at the proper place, it should be earlier in the doc

articles/libraries/lock/v10/configuration.md
```diff
@@ -96,6 +97,16 @@ var lock = new Auth0Lock('clientID', 'account.auth0.com', options);
 
 ## Display Options
 
+### allowAutoComplete {Boolean}
+
+Determines whether or not the the email or username inputs will allow autocomplete (`<input autocomplete />`). Defaults to `false`.
```
mpaktiti: typo: the the email

config/sidebar.yml
```diff
@@ -502,7 +502,7 @@ libraries:
         url: "/libraries/lock/v10/auth0js"
 
       - title: "Popup Mode"
```
mpaktiti: You should update the title to match the new one

articles/libraries/lock/v9/authentication-modes.md
```diff
@@ -1,6 +1,6 @@
 ---
 section: libraries
-description: Lock V9 doc on the different types of authentication modes.
+description: Lock v9 doc on the different types of authentication modes.
 ---
 # Lock: Authentication Modes
```
mpaktiti: typo: openned

articles/libraries/lock/v9/authentication-modes.md
```diff
@@ -1,6 +1,6 @@
 ---
 section: libraries
-description: Lock V9 doc on the different types of authentication modes.
+description: Lock v9 doc on the different types of authentication modes.
```
mpaktiti, line 3: add toc

articles/libraries/lock/v10/authentication-modes.md
mpaktiti: The same comment applies to v9 https://auth0-docs-content-pr-5062.herokuapp.com/docs/libraries/lock/v9/authentication-modes#database-connections-and-popup-mode

articles/libraries/lock/v9/authentication-modes.md
```diff
@@ -14,20 +14,61 @@ If after you click on the IdP button (Facebook for example), the Webapp you buil
 
 ![Widget redirect](/media/articles/libraries/lock/v9/WidgetRedirect.gif)
 
-You can [click here](/libraries/lock/v9/types-of-applications#redirect-mode) to learn how to implement Redirect mode with Single Page Apps, or you can [click here](/libraries/lock/v9/types-of-applications#redirect-mode-1) to learn how to implement it with Regular WebApps.
+### Redirect Mode Implementation Examples
```
mpaktiti: I would remove this and name the sub-paragraphs to sth like `### Example: Redirect Mode in SPA` or `### How to enable Redirect Mode in SPA`

Same comment applies to all paragraphs/sub-paragraphs examples in this doc

articles/libraries/lock/v9/authentication-modes.md
```diff
@@ -14,20 +14,61 @@ If after you click on the IdP button (Facebook for example), the Webapp you buil
 
 ![Widget redirect](/media/articles/libraries/lock/v9/WidgetRedirect.gif)
```
mpaktiti: I would like to see also these GIFs updated. I understand it's an outdated version, but since you will do the job for v10 I don't think it will be so much more effort to switch to v9 and get a gif as well. Let me know if otherwise.

mpaktiti (changes requested):

articles/libraries/lock/v10/auth0js.md
jeffreylees: It seems like the code picker UI widget is intended for use with a single block of code with a particular syntax (defaulting to code highlighting for HTML).

Here we have paragraphs, subheadings, HTML code blocks, JS code blocks... so far I haven't found a way to make that work inside this UI element, or an example of that elsewhere. 

If you don't know how to do this either, but you really want it used here, I can try and refactor the content to not need the headings, different code blocks, etc?

articles/libraries/lock/v10/authentication-modes.md
jeffreylees: Makes sense. I really like the idea of breadcrumbs in the docs. I know a lot of people don't like them (heck, I don't really like them) but it could be really valuable in sections where the sidebar is not exhaustive or the docs didn't merit being included in the sidebar.

jeffreylees:
@mpaktiti I've pushed changes for all of your feedback save for:

- https://github.com/auth0/docs/pull/5062#discussion_r141643342 the code picker widget in this case, not sure how to make that work or that it's worth the trouble TBH
- GIFs. I removed references to the old GIFs. I will look at how to do new ones tomorrow if time permits - to be frank I'm not a big fan of GIFs in the docs as they're impossible to maintain, requiring setting up a demo app and recording a new GIF every time there are changes - but I will also admit they are nice, and good for a visual example, yeah? Anyway, will look into it for both versions. If you'd rather finish this one up, I can do another PR tomorrow or next week to add in the GIFs (and to be honest, if I end up doing them, I might see if there are any more we might could use in the Lock docs?).

Also I see there's a conflict now, interesting since I've rebased and all claims to be synced up. Will look at that tomorrow too. It's probably from me adding `audience` earlier.

articles/libraries/lock/v10/auth0js.md
mpaktiti: @jeffreylees Only the snippets would be in the picker, not all the content. For example you would have the same text for including (`If you included the Lock script from the Auth0 CDN, you will need to also include the auth0.js script before Lock:`)

and then you would have the UI picker with the different snippets.

mpaktiti:
@jeffreylees I replied to the comment about UI picker. You can address the GIFs in another PR, just add a trello card and prioritize it, if possible.

articles/libraries/lock/v10/auth0js.md
```diff
@@ -1,75 +1,77 @@
...
 
-## Changes with newer versions
-
-If you try to use auth0.js along with Lock 10, you will not be able to call `getClient()`. Instead, you should include its dependency and instantiate the object.
-
-In the past, a version of auth0.js was included automatically with Lock; at this point, you must include auth0.js yourself, allowing you to choose the supported version of auth0.js that best suits the needs of your application - version 7, or version 8. Note that the two versions have different CDN paths, as well as slightly different instantiations.
+For simple uses, Lock is all that is necessary. However, while using Lock, if more customization is required in an application than Lock allows, functionality from the Auth0.js SDK can be used alongside Lock to meet those needs. An example might be using Lock to handle signups and logins, while using auth0.js to [manage users](/libraries/auth0js#user-management) (read and update user metadata, link user accounts together, and similar tasks).
 
 ## Using auth0.js v8
```
mpaktiti: This header is no longer needed since the paragraph includes both v7 and v8.

mpaktiti: Also, the first snippet is `Using Auth0.js v8/Using Auth0.js v7`, while the second one is `Auth0.js v8/Auth0.js v7`. Use one consistently.

articles/libraries/lock/v10/auth0js.md
```diff
@@ -1,75 +1,77 @@
...
+	<div class="languages-bar">
+	  <ul>
+	    <li><a href="#cdn-v8" data-toggle="tab">Using Auth0.js v8</a></li>
+	    <li><a href="#cdn-v7" data-toggle="tab">Using Auth0.js v7</a></li>
+	  </ul>
+	</div>
+	<div class="tab-content">
+	  <div id="cdn-v8" class="tab-pane active">
+	  <pre class="hljs html"><code>
+	&lt;script src="${auth0js_urlv8}"&gt;&lt;/script&gt;
```
mpaktiti: There is some whitespace that should be removed. Not sure if you can manage this just by removing the spaces before the code, but try it.
image

articles/libraries/lock/v10/auth0js.md
mpaktiti: @jeffreylees Did you check if this is an actual requirement?

mpaktiti (changes requested):
Looks good Jeff 👍  Just a few final things and we're good to go. Btw don't forget to create a card for the GIFs if you haven't already.

jeffreylees:
I don't think the order is a requirement. I dredged up an old conversation with Luis discussing the use of the two together, the current NPM verbiage is actually his wording, but yeah. As far as by CDN goes, you'll need both (as you aren't importing dependencies of dependencies like you are with NPM). The reason you don't have to explicitly install auth0.js with NPM is that it's already a dependency of Lock.

Made the other changes, and reworded ^ to make it less assertive about order, and about having to include it via CDN.

mpaktiti (approved):
Perfect 💯

PR #5065, opened 2017-09-27
Custom DB / users import: document new data validations

mpaktiti:
Document new (optional) data validations when importing users from custom databases https://github.com/auth0/auth0-users/pull/908

This is enabled with `strategy_version=2` https://github.com/auth0/auth0-users/pull/912/files

### DIFF
--- a/articles/connections/database/migrating.md
+++ b/articles/connections/database/migrating.md
@@ -59,7 +59,7 @@ You can enable these validations for a connection using the [Update a connection
   "queryString": [],
   "postData": {
     "mimeType": "application/json",
-    "text": "{ \"options\": { \"strategy_version\": 2 } }"
+    "text": "{ \"options\": { \"strategy_version\": 2, \"nextOptionParam\": \"...\" } }"
   },
   "headersSize": -1,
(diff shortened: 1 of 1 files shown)


articles/connections/database/migrating.md
```diff
@@ -1,37 +1,76 @@
...
+  "httpVersion": "HTTP/1.1",
+  "cookies": [],
+  "headers": [
+    { "name": "Authorization", "value": "Bearer {access-token}" }, 
+    { "name": "Content-Type", "value": "application/json"}
+  ],
+  "queryString": [],
+  "postData": {
+    "mimeType": "application/json",
+    "text": "{ \"options\": { \"strategy_version\": 2 } }"
```
glena: can we add something here to make it clear that there are missing attributes in the options object to avoid people overriding if they dont read further?

something like `"text": "{ \"options\": { \"strategy_version\": 2, ... } }"` (not sure how do you use to handle this cases in other docs)

maybe even a comment pointing them to read the note about this.

You know, people will copy paste and run it without reading the following notes

PR #5075, opened 2017-09-28
Fix terminal commands for easier copying

erjohnson:
Minor fix to some terminal commands for easier copy/pasting.

[Install the AD/LDAP Connector on Non-Microsoft Platforms](https://auth0-docs-content-pr-5075.herokuapp.com/docs/connector/install-other-platforms)


mpaktiti (approved):
💯

PR #5082, opened 2017-10-02
Fixing MFA protocol links and other minor adjustments

jeffreylees:

### DIFF
--- a/articles/multifactor-authentication/custom/custom-landing.md
+++ b/articles/multifactor-authentication/custom/custom-landing.md
@@ -91,7 +91,7 @@ function (user, context, callback) {
 
 ## Use a Custom MFA Service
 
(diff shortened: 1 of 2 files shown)


articles/multifactor-authentication/custom/custom-landing.md
```diff
@@ -91,7 +91,7 @@ function (user, context, callback) {
 
 ## Use a Custom MFA Service
 
-If you are using an MFA provider that does not have Auth0 built-in support or if you are using a service you have created, you can use the [redirect](/protocols#redirect-protocol-in-rules) protocol for the integration.
+If you are using an MFA provider that does not have Auth0 built-in support or if you are using a service you have created, you can use the [redirect](/rules/current/redirect) protocol for the integration.
```
mpaktiti: It's better if you don't use the version at the URL (unless you wanna go explicitly to the older version). Within the next year the rules/legacy will be removed, and if we have versions hardcoded we will have to update the docs or add redirects.

So use `/rules/redirect` instead. I'll make the updates in this PR.

PR #5083, opened 2017-10-03
Add web_message response_mode

mpaktiti:


articles/protocols/oauth2/index.md
```diff
@@ -156,6 +156,8 @@ The [OAuth 2.0 Multiple Response Type Encoding Practices](https://openid.net/spe
   </html>
   ```
 
+- `web_message`: This response mode is defined by the [OAuth 2.0 Web Message Response Mode specification](https://tools.ietf.org/html/draft-sakimura-oauth-wmrm-00). It uses HTML5 Web Messaging instead of the redirect for the Authorization Response from the Authorization Endpoint. Basically it enables your single-page application to perform a sign-in flow in an iFrame or a popup window, and get the tokens in the parent page securely, and without leaving the context of that page. To use this response mode you have to register your app's URL at the __Allowed Web Origins__ field of your [Client's settings](${manage_url}/#/clients/${account.clientId}/settings).
```
jhiner: > Basically it enables your single-page application to perform a sign-in flow in an iFrame or a popup window, and get the tokens in the parent page securely, and without leaving the context of that page. 

I would remove this and basically say instead:

> This is particularly useful when using [Silent Authentication](https://auth0.com/docs/api-auth/tutorials/silent-authentication) with `prompt=none`

mpaktiti:
Related to https://github.com/auth0/manage/pull/3255

luisrudge (approved):

PR #5087, opened 2017-10-04
QS React Native, expand configuration notes, move ordering of callback section

cocojoe:


articles/quickstart/native/react-native/00-login.md
```diff
@@ -16,8 +16,6 @@ budicon: 448
 
 <%= include('_includes/_intro') %>
 
-<%= include('_includes/_config') %>
-
```
lbalmaceda: This was here because if you see the [final document](https://auth0-docs-content-pr-5087.herokuapp.com/docs/quickstart/native/react-native) there's a "Create client" step, which should be followed by getting the client_id/domain values and setting up the callback url, because you're already there in the client dashboard. I'd refactor the document to keep it that way. The rest looks good 👍

lbalmaceda (changes requested):

cocojoe:
@lbalmaceda have a look once new deployment is up, have expanded it a bit more, the audience for react-native is often non-mobile people so adding extra info.
If you feel it needs anything else let me know. I've not mentioned how to create a project as this should at least be covered by the officail getting started 😄

articles/quickstart/native/react-native/_includes/_login.md
```diff
@@ -76,6 +88,10 @@ You can then register the identifier by adding the following snippet:
 ```
 
 ::: note
+The value org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier) is the default for apps created with React Native CLI, you may have a different value.
+:::
+
+::: note
 For further reference on linking, check [the official documentation](https://facebook.github.io/react-native/docs/linking.html)
```
lbalmaceda: This note should be in the "Link the native module" section. I'd also add a sentence on that section telling that "libraries like this depend on other native platform's libraries such as Android and iOS, which requires us to first link those dependencies to be able to use the native code". Or something like that. It feels so lonely that section with just a header and a snippet... Also I don't think "native dependencies" is a thing you see on every RN library, and although well documented on their site, we should just introduce or mention them. 

Another thing. After the snippet, you go straight to the "how to login using auth0" text which IMO should be the first thing in Configuration section or at least have a separate header because it's kinda lost there below the Install snippets

lbalmaceda (changes requested):

articles/quickstart/native/react-native/_includes/_login.md
cocojoe: The https://facebook.github.io/react-native/docs/linking.html isn't anything to do with native module linking. It's a section about handling universal links etc
I might just remove it from the QS, it's not something the avg person cares about anyway, it's still in the README though for more detail.

The `link the native module` is about linking the module itself into the react-native project.

I'll check other QS formatting regarding the header.

lbalmaceda (approved):

PR #5095, opened 2017-10-05
Adding links / references to cross-origin authentication document

jeffreylees:

### DIFF
--- a/articles/hosted-pages/login/index.md
+++ b/articles/hosted-pages/login/index.md
@@ -10,7 +10,7 @@ crews: crew-2
 
 Auth0's Hosted Login Page is the most secure way to easily authenticate users for your applications. The Hosted Login Page is easily customizable right from the [Dashboard](${manage_url}). By default, the Hosted Login Page uses Auth0's [Lock Widget](/libraries/lock) to authenticate your users, but the code of the Hosted Login Page can be customized to replace Lock with the Lock Passwordless widget, or an entirely custom UI can be built in its place, using the [Auth0.js SDK](/libraries/auth0js) for authentication.
 
-The Hosted Login Page is the safest way to authenticate your users with Auth0. Because the Hosted Login Page does not involve embedding authentication in an application, [cross-origin authentication](/cross-origin-authentication) can be avoided.
+If you cannot use the Hosted Login Page, you can embed the Lock widget or a custom login form in your application using [cross-origin authentication](/cross-origin-authentication), but be sure to read about its limitations before choosing to do so.
 
 ![Hosted Login Page](/media/articles/hosted-pages/hlp-lock.png)
 
--- a/articles/hosted-pages/login/index.md
+++ b/articles/hosted-pages/login/index.md
@@ -10,6 +10,8 @@ crews: crew-2
 
 Auth0's Hosted Login Page is the most secure way to easily authenticate users for your applications. The Hosted Login Page is easily customizable right from the [Dashboard](${manage_url}). By default, the Hosted Login Page uses Auth0's [Lock Widget](/libraries/lock) to authenticate your users, but the code of the Hosted Login Page can be customized to replace Lock with the Lock Passwordless widget, or an entirely custom UI can be built in its place, using the [Auth0.js SDK](/libraries/auth0js) for authentication.
 
+The Hosted Login Page is the safest way to authenticate your users with Auth0. Because the Hosted Login Page does not involve embedding authentication in an application, [cross-origin authentication](/cross-origin-authentication) can be avoided.
+
 ![Hosted Login Page](/media/articles/hosted-pages/hlp-lock.png)
 
 ### How Does the Hosted Login Page Work
--- a/articles/libraries/auth0js/v8/index.md
+++ b/articles/libraries/auth0js/v8/index.md
@@ -483,6 +483,13 @@ If attempting to set up a password reset functionality, you'll use the `changePa
 
 The user will then receive an email which will contain a link that they can follow to reset their password.
 
+## Cross-Origin Authentication
+
+Using auth0.js within your application, rather than using the [Hosted Login Page](/hosted-pages/login), requires [cross-origin authentication](/cross-origin-authentication). In order to use embedded auth0.js via cross-origin authentication, you must do the following:
+
+* Set the [audience](/libraries/lock/v10/configuration#audience-string-) option
+* In the client settings area of the [Dashboard]($manage_url}), set OIDC Conformant and Cross Origin Authentication to true.
+
 ## User management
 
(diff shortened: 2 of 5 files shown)


articles/libraries/auth0js/v8/index.md
```diff
@@ -483,6 +483,13 @@ If attempting to set up a password reset functionality, you'll use the `changePa
 
 The user will then receive an email which will contain a link that they can follow to reset their password.
 
+## Cross-Origin Authentication
+
+Using auth0.js within your application, rather than using the [Hosted Login Page](/hosted-pages/login), requires [cross-origin authentication](/cross-origin-authentication). In order to use embedded auth0.js via cross-origin authentication, you must do the following:
+
+* Set the [audience](/libraries/lock/v10/configuration#audience-string-) option
+* In the client settings area of the [Dashboard]($manage_url}), in the **Advanced Settings** menu, under the **OAuth** tab, turn on the OIDC Conformant and Cross Origin Authentication settings.
```
mpaktiti: These are fields so they should be formatted as bold: OIDC Conformant, Cross Origin Authentication

I'd also add a screenshots of these fields.

articles/hosted-pages/login/index.md
```diff
@@ -10,6 +10,8 @@ crews: crew-2
 
 Auth0's Hosted Login Page is the most secure way to easily authenticate users for your applications. The Hosted Login Page is easily customizable right from the [Dashboard](${manage_url}). By default, the Hosted Login Page uses Auth0's [Lock Widget](/libraries/lock) to authenticate your users, but the code of the Hosted Login Page can be customized to replace Lock with the Lock Passwordless widget, or an entirely custom UI can be built in its place, using the [Auth0.js SDK](/libraries/auth0js) for authentication.
 
+The Hosted Login Page is the safest way to authenticate your users with Auth0. Because the Hosted Login Page does not involve embedding authentication in an application, [cross-origin authentication](/cross-origin-authentication) can be avoided.
```
mpaktiti: The first sentence repeats what we say in the first sentence of the previous paragraph.

Also, I don't get the way this link is presented. It's like you already talked about a problem called `cross-origin authentication` and now you tell them that this way they can avoid them. Isn't the purpose of this link to show this alternative option? Why not use something like `If you do not want to use centralized authentication, you can embed the Lock widget or a custom login form in your application using [cross-origin authentication](/cross-origin-authentication)`

articles/libraries/lock/v10/index.md
```diff
@@ -115,6 +115,14 @@ document.getElementById('btn-login').addEventListener('click', function() {
 });
 ```
 
+## Cross-Origin Authentication
+
+Using Lock within your application, rather than using the [Hosted Login Page](/hosted-pages/login), requires [cross-origin authentication](/cross-origin-authentication). In order to use embedded Lock via cross-origin authentication, you must do the following:
```
mpaktiti: `Embedding Lock in your application`

articles/libraries/lock/v10/index.md
```diff
@@ -115,6 +115,14 @@ document.getElementById('btn-login').addEventListener('click', function() {
...
+
+* Set the [oidcconformant](/libraries/lock/v10/configuration#oidcconformant-boolean-) option to true
+* Set the [audience](/libraries/lock/v10/configuration#audience-string-) option
+* In the client settings area of the [Dashboard]($manage_url}), in the **Advanced Settings** menu, under the **OAuth** tab, turn on the OIDC Conformant and Cross Origin Authentication settings.
```
mpaktiti: These are fields so they should be formatted as bold: OIDC Conformant, Cross Origin Authentication

I'd also add a screenshots of these fields.

articles/sso/current/index.md
```diff
@@ -23,27 +23,29 @@ In the case of SSO with Auth0, the *Central Service* is the Auth0 Authorization
...
+1. Auth0 will check to see whether there is an existing SSO cookie.
+1. Because this is the first time the user visits this Hosted Login page, and no SSO cookie is present, they may be presented with username and password fields and also possibly some Social Identity Providers such as LinkedIn, GitHub, etc. (The exact layout of the Lock screen will depend on the [Identity Providers](/identityproviders) you have configured.
 
     ![](/media/articles/sso/single-sign-on/lock-no-sso-cookie.png)
 
-4. Once the user has logged in, Auth0 will set an SSO cookie
-5. Auth0 will also redirect back to your web application and will return an `id_token` containing the identity of the user.
+1. Once the user has logged in, Auth0 will set an SSO cookie
+1. Auth0 will also redirect back to your web application and will return an `id_token` containing the identity 1f the user.
```
mpaktiti: `1f`

articles/sso/current/index.md
```diff
@@ -23,27 +23,29 @@ In the case of SSO with Auth0, the *Central Service* is the Auth0 Authorization
...
 
 Now let's look at flow when the user returns to your website for a subsequent visit:
 
 1. Your application will redirect the user to the Auth0 Hosted Login page where they can sign in.
-2. Auth0 will check to see whether there is an existing SSO cookie.
-3. This time Auth0 finds an SSO cookie and instead of displaying the normal Lock screen with the username and password fields, it will display a Lock screen which indicates that we know you the user is, as they have already logged in before. They can simply confirm that they want to log in with that same account.
+1. Auth0 will check to see whether there is an existing SSO cookie.
+1. This time Auth0 finds an SSO cookie and (if using Lock) instead of displaying the normal Lock screen with the username and password fields, it will display a Lock screen which indicates that we know you the user is, as they have already logged in before. They can simply confirm that they want to log in with that same account.
```
mpaktiti: This doesn't read well: `This time Auth0 finds an SSO cookie and (if using Lock) instead of displaying`

mpaktiti (changes requested):

articles/hosted-pages/login/index.md
jeffreylees: Since we're saying HLP is definitively the best and safest option, I was thinking we might want to also show readers the method we're comparing it to (safest vs what?). If that looks bad, and the alternative is something like what you've suggested above, I think I might rather backtrack and not add anything at all here - that reads like COA is a recommended alternative option to me, whereas we would rather recommend that COA wasn't used, wherever possible.

I suppose we could also opt for a slightly different wording, like 

`If you cannot use centralized authentication, you can embed the Lock widget or a custom login form in your application using [cross-origin authentication](/cross-origin-authentication), but be sure to read about its limitations before choosing to do so.` or something similar, that doesn't assume foreknowledge of COA but makes it clear that this isn't the path we'd like them to take.

What do you think?

jeffreylees: Just went with the latter rewording - let me know if you are not ok with this

PR #5106, opened 2017-10-05
Update part-3.md

obogobo:
change `method.request.header.Authorization` to just plain `Authorization` in Token Source, API Gateway fails to invoke custom authorizer if method.request.header is specified


khorne3:
Looks good. 👍

mpaktiti:
Thank you @obogobo and @khorne3 ! 👍

PR #5119, opened 2017-10-09
Remove trailing quotation in auth0 parameters.

stcastle:
When the web page renders with the user's local clientId and domain, the trailing quotation mark remains. If copied into the resource file in Android, the application will successfully run, but the auth0 server will not recognize the clientId because of the trailing quotation mark. This is misleading, and I believe it's a mistake.


lbalmaceda (approved):
Thanks. It's definitely a mistake.

PR #5122, opened 2017-10-10
Further updates to COA documentation

jeffreylees:


mpaktiti:
Hey Jeff, I removed the `zoomable` element since we are supposed to use it only on complex diagram images with `.svg` format

jeffreylees:
Haha, yeah, I saw that in the guidance when I looked it up (not sure I had seen zoomable before) but I just found it kind of hard to read the text without being able to zoom :grin: but my eyes are for shit so that's probably the real problem.

PR #5134, opened 2017-10-12
Change Node.js to JS

jerriep:
Based on customer feedback, labelling the code samples as `Node.js` is confusing since it seem to imply that this gets executed on the server. But this is should be executed on the client instead.

Have labelled code samples (where possible) as JavaScript instead


mpaktiti:
Currently using a workaround, to be solved by https://github.com/auth0/auth0-docs/issues/1512

PR #5135, opened 2017-10-12
Added Link to Custom Login Form

cocojoe:


lbalmaceda (approved):
The HLP is introduced in the "link module" section along with a screenshot of what should be the output of running the QS. And you mention two times "Auth0 hosted login page" in line 96, maybe there's a way to rephrase that second sentence. But besides that, looks ok. Check deployment [here](https://auth0-docs-content-pr-5135.herokuapp.com/docs/quickstart/native/react-native/00-login).

cocojoe:
This is a standard copy section I took from SPA QS, although I removed the word `Lock` and replaced with Custom Login Form. I can tighten it up as it has been previously mentioned.

lbalmaceda:
Looks better now IMO.

PR #5138, opened 2017-10-13
Migrate QS iOS Swift User Session guide to use credentials manager

cocojoe:
Remove redundant sections
Link the older SimpleKeychain methods


lbalmaceda (approved):
LGTM

PR #5140, opened 2017-10-13
Reorganize stuff in /tutorials

khorne3:
Found new homes (some of which are newly created) for everything in /tutorials


mpaktiti:
@khorne3 Can you resolve the conflicts please?

khorne3:
@mpaktiti Rebased. Thank you.

articles/clients/how-to-update-client-secret.md
```diff
@@ -5,7 +5,7 @@ crews: crew-2
```
mpaktiti, line 5: Hey Katie, this doc needs some love. There are things wrong and others missing, for example:
- no reference of how to get a Mgmt APIv2 token. It's also missing from the har https://auth0.com/docs/clients/how-to-update-client-secret#using-the-management-api-to-set-the-new-client-secret
- refers to Base64-encoded client secrets

Also the client_secret does not look good at headers/titles, it's better if it was Client Secret. Can you add a new trello card in order to review this?

mpaktiti (commented):
This looks good Katie. As a second step (=new trello card) we can build indexes for some of the categories and list the docs there. For example we could add a "Monitoring" index page and a "Support" index page. Wdyt? Wanna take this over?

PR #5141, opened 2017-10-13
prevent_sign_ups appears to be a typo

kcculhwch:
It looks like `prevent_sign_up` is the param that is actually passed into the uri, not `prevent_sign_ups`.


jeffreylees (approved):
The URL as written in the doc and the one in the screenshot are different, so I'm good with the change to make them match up.

It's essentially a custom param, so either works as long as you're putting it into the URL that way (and if you're hitting the page via an SDK and the `authorize` endpoint, neither may).

But the doc should be internally consistent - thanks for spotting that!

mpaktiti:
Thank you for the contribution @kcculhwch ! 👍

PR #5146, opened 2017-10-15
Minor library fixes

jeffreylees:
Docs-Request

- `allowAutoComplete` is not the name of the option as represented in the code - no capital C.
- Bad links in Auth0.js for replicating their own manual parsing. One goes to the wrong line in the `index.js` for auth0.js; the other goes to 404. Removed both links altogether and just linking to the file `index.js` - better to recommend they CMD+F than to constantly be pointing to the wrong line number as the file changes.


mpaktiti:
@jeffreylees missing review app

PR #5150, opened 2017-10-16
Update reference to correct sample

khorne3:
Found one reference on the `current` SSO docs pointing to legacy samples, so I updated the link.

https://auth0-docs-content-pr-5150.herokuapp.com/docs/sso/current#checking-the-user-s-sso-status-from-the-client

### DIFF
--- a/articles/sso/current/index.md
+++ b/articles/sso/current/index.md
@@ -71,12 +71,10 @@ Whenever you need to determine the user's SSO status, you'll need to check the f
 
 If you don't have a valid `accessToken`, the user is *not* logged in. However, they may be logged in via SSO to another associated application -- you can determine if this is the case or not by calling the `renewAuth` method of the auth0.js library, which will attempt to silently authenticate the user within an iframe. Whether the authentication is successful or not indicates whether the user has an active SSO cookie.
 
-For more detailed information on how to implement this, please refer to the following:
-
-* [Client-Side SSO (Single Page Apps)](/sso/current/single-page-apps-sso)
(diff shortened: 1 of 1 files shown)


articles/sso/current/index.md
```diff
@@ -71,8 +71,7 @@ For more detailed information on how to implement this, please refer to the foll
 * [Client-Side SSO (Single Page Apps)](/sso/current/single-page-apps-sso)
 
 ::: note
-Please see the [Auth0 SSO Sample](https://github.com/auth0/auth0-sso-sample) repo for an example of SSO with both Single Page Apps and Regular Web Apps.
-:::
+Please see the [Auth0 OIDC SSO Sample](https://github.com/auth0-samples/oidc-sso-sample) repo for an example of how to implement OIDC-compliant SSO.
```
mpaktiti: Forgot some markup:

image

Keep in mind to always check the review app, especially when you are editing visual elements like notes and panels.

articles/sso/current/index.md
```diff
@@ -71,8 +71,7 @@ For more detailed information on how to implement this, please refer to the foll
 * [Client-Side SSO (Single Page Apps)](/sso/current/single-page-apps-sso)
```
mpaktiti: You need to also update the text before the sample, since it still describes the old sample (uses the `getSSOData()` function, which is legacy)

mpaktiti (changes requested):

khorne3:
@mpaktiti  I pushed the commits, so it should work better now. :)

mpaktiti (approved):

PR #5154, opened 2017-10-17
Add instructions on finding Azure AD for Office 365 account

erjohnson:
Add instructions on finding the Azure AD which backs an Office 365 account, and clarify that creating a new Azure AD is not required (if you already have an Office 365 account).

Perhaps this blurb should be in a panel? Or a subsection?

[Connect your app to Microsoft Azure Active Directory](https://auth0-docs-content-pr-5154.herokuapp.com/docs/connections/enterprise/azure-active-directory/v2)


mpaktiti:
@erjohnson Yes, you could use a panel. Feel free to send another PR if you think it looks better that way.

::: panel Use your Office 365 account
...
:::

PR #5162, opened 2017-10-18
Lock 11 Reference Docs version

jeffreylees:


articles/libraries/lock/v11/migration-guide.md
```diff
@@ -0,0 +1,73 @@
...
+
+You can enable COA using the Dashboard, for details refer to [Configure Your Client for Cross-Origin Authentication](/cross-origin-authentication#configure-your-client-for-cross-origin-authentication).
+
+::: note
+Cross Origin Authentication has some limitations, before you enable it for your app make sure that you are aware of them. For details refer to [Cross Origin Authentication](/cross-origin-authentication).
+:::
+
+### Single Sign On using IP Ranges
+
+Up to Lock 10, you could configure IP ranges in an Active Directory/LDAP connection and use them to allow integrated Windows Authentication if the user attempting to log in was within that range. Lock would display a button enabling SSO for the user.
```
aaguiarz: "Up to Lock 10, if you configured IP ranges in an Active Directory/LDAP connection, Lock would notice you are connecting from that IP range andredirect to the Active Directory login page'.

Lock 11 will no longer do that. It will prompt for user/password in Lock itself and authenticate with the AD directory. 

You can achieve the Lock 10 behavior by using the Hosted Login Page'

articles/libraries/lock/v11/migration-guide.md
```diff
@@ -0,0 +1,79 @@
...
+section: libraries
+title: Lock v10 to v11 Migration Guide
+description: How to migrate from Lock v10 to Lock v11
+toc: true
+---
+# Migration Guide: Lock v10 to v11
+
+This document lists all the changes you should be aware of between versions 10 and 11 of Lock. It includes information on what is changing and why, details on new or deprecated features, and instructions on how you can migrate your implementation.
+
+If you are using Lock in embedded mode, you are encouraged to use the latest, and more secure, version of the widget, but before you update your code, make sure that you have reviewed this document and made any necessary changes in your implementation.
```
aaguiarz: Let's suggest to use the HLP here too.

articles/libraries/lock/v11/migration-guide.md
```diff
@@ -0,0 +1,71 @@
...
+```js
+{
+  "error":"invalid_request",
+  "error_description":"Connection strategy not supported."
+}
+```
+
+## Default Values
+
+Both Lock 11 and Auth0.js 9 will default the `scope` parameter to `'oidc profile email'`. This is to make `getSSOData` and the "Last Logged in With" window work correctly.
```
jhiner: I think you mean `openid profile email` scope

articles/libraries/lock/v11/index.md
```diff
@@ -0,0 +1,142 @@
...
+title: Lock 11 for Web
+description: A widget that provides a frictionless login and signup experience for your web apps.
+img: media/articles/libraries/lock-web.png
+---
+# Lock 11 for Web
+
+Lock is an embeddable login form, [configurable to your needs](/libraries/lock/v11/configuration) and ready for use on web apps. It enables you to easily add social identity providers to Lock, allowing your users to login seamlessly using any provider they want.
+
+::: panel Hosted Login Page
+Lock 11 is a version of Lock tailored for embedded login in your applications. If you use Auth0's [Hosted Login Page](/hosted-pages/login), Lock 11 is not necessary and does not work inside the Hosted Login Page. The Hosted Login Page remains the simplest and most secure method by which to authenticate users for your applications. Lock 11 is designed to make cross-origin authentication work smoothly for when you cannot use the Hosted Login Page and need to embed a login in your application.
```
nicosabena: I'm not sure about the "Lock 11 is not necessary" part, it's a little confusing.
How about "If you use Auth0's [Hosted Login Page](/hosted-pages/login), do not upgrade to Lock 11. Instead, keep using the latest version of Lock 10."
This might event warrant some graphic.
Want Lock or Auth0.js embedded in your app? -> Use Lock 11 or Auth0.js 9
Want Lock or Auth0.js in the hosted login page (a.k.a. centralized login) -> Use Lock v10 or Auth0.js 8

articles/libraries/lock/v11/index.md
```diff
@@ -0,0 +1,142 @@
...
+<meta name="viewport" content="width=device-width, initial-scale=1"/>
+```
+
+### Bundling Dependencies
+
+If you are using browserify or webpack to build your project and bundle its dependencies, after installing the `auth0-lock` module, you will need to bundle it with all its dependencies. Examples are available for [Browserify](https://github.com/auth0/lock/tree/master/examples/bundling/browserify) and [webpack](https://github.com/auth0/lock/tree/master/examples/bundling/webpack).
+
+### Cross-Origin Authentication
+
+Embedding Lock within your application, rather than using the [Hosted Login Page](/hosted-pages/login), requires [cross-origin authentication](/cross-origin-authentication). In order to use embedded Lock via cross-origin authentication, you ensure that the client settings area of the [Dashboard]($manage_url}), in the **Advanced Settings** menu, under the **OAuth** tab, ensure that the **OIDC Conformant** and **Cross Origin Authentication** settings are turned on.
```
nicosabena: CO authentication works with both OIDC-Conformant and non-OIDC-Conformant apps, and the **Cross-Origin Authentication** toggle won't be present anymore.
We can note, however, that 
> Embedding Lock within your application requires [cross-origin authentication](/cross-origin-authentication), which is only enabled for [first-party clients](/link-to-relevant-doc).

articles/libraries/lock/v11/migration-guide.md
```diff
@@ -0,0 +1,71 @@
...
+}
+```
+
+## Default Values
+
+Both Lock 11 and Auth0.js 9 will default the `scope` parameter to `'openid profile email'`. This is to make `getSSOData` and the "Last Logged in With" window work correctly.
+
+### getProfile
+
+In Lock 10 this function received a string parameter with an ID Token. If you choose to use it in Lock 11, it needs to receive an Access Token. You’ll need to update your code to change the parameter sent.
```
nicosabena: I would add that `getProfile` is now effectively a synonym for getUserInfo.

articles/libraries/lock/v11/sending-authentication-parameters.md
```diff
@@ -0,0 +1,71 @@
...
+var options = {
+  auth: {
+    params: {scope: 'openid email user_metadata app_metadata picture'},
+  }
+};
+```
+
+There are different values supported for scope. Keep in mind that JWTs are sent on every API request, so it is desirable to keep them as small as possible:
+
+* `scope: 'openid'`: _(default)_ It will return not only the `access_token`, but also an `id_token` which is a JSON Web Token (JWT). The JWT will only contain the user ID (`sub` claim).
```
nicosabena: On other documents for Lock v11 and Auth0.js 9, we mention that the default scope is now `openid profile email`.

articles/libraries/lock/v11/sending-authentication-parameters.md
```diff
@@ -0,0 +1,71 @@
...
+* `scope: 'openid profile'`: will return all the user attributes in the token.
+* `scope: 'openid {attr1} {attr2} {attrN}'`: If you want only specific user attributes to be part of the `id_token` (For example: `scope: 'openid name email picture'`). When selecting specific attributes, the attributes chosen are from those available in the user's profile, which will vary from application to application.
```
nicosabena: This is not quite true for OIDC-Conformant apps. We might want to avoid repeating what is (or should be) already specified in `/scopes`.

articles/libraries/lock/v11/sending-authentication-parameters.md
```diff
@@ -0,0 +1,71 @@
...
+
+The default `scope` value in Lock 11 is `openid profile email`.
+
+For more information about scopes, see the [scopes documentation page](/scopes).
+
+#### Example: retrieve a token with the profile data
```
nicosabena: Careful with this. We only "retrieve a token with the profile data" if responseType="id_token". Otherwise, we are asking for a token that can be used to retrieve the user's profile data.

articles/libraries/lock/v11/sending-authentication-parameters.md
```diff
@@ -0,0 +1,71 @@
...
+
+If you want to do this using Lock widget version 11, you should add the `scope` parameter. For example in AngularJS you would use the initializing method of `authProvider`:
+
```
nicosabena: Not sure if the AngujarJS provider mention is useful or confusing. Shouldn't we stick to Lock's API here?

articles/libraries/lock/v11/sending-authentication-parameters.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+::: note
+There is also a `connectionScopes` configuration option for Lock 11, which allows you to specify scopes on any specific connection. This will be useful if you want to initially start with a set of scopes (defined on the dashboard), but later on request additional permissions or attributes. Read more about it on the [Lock Configuration Options](/libraries/lock/v11/configuration#connectionscopes-object-) page.
+:::
+
+### state {string}
+
+The `state` parameter is an arbitrary state value that will be mantained across redirects. It is useful to mitigate [XSRF attacks](http://en.wikipedia.org/wiki/Cross-site_request_forgery) and for any contextual information, [such as a return url](/tutorials/redirecting-users), that you might need after the authentication process is finished.
+
```
nicosabena: I would add that if the user doesn't provide a `state`, Lock will automatically generate one.

nicosabena: We should also add the `nonce` parameter (used to mitigate replay attacks) and mention that Lock will generate one automatically if not provided.

PR #5169, opened 2017-10-19
[edit]: updated per feedback from Bob and GermanM.

ndelgado:


germanmatosas:
> Other cloud environments may also be support but will need to be reviewed.

I don't like the "support" (or supported) word here, I prefer "other public cloud service providers will need to be reviewed"

PR #5186, opened 2017-10-23
Update Tutorials.md

khorne3:
Remove recursive link.

https://auth0-docs-content-pr-5186.herokuapp.com/docs/tutorials

### DIFF
--- a/articles/tutorials/index.md
+++ b/articles/tutorials/index.md
@@ -1,7 +1,0 @@
----
-url: /tutorials
----
-
-# Quickstarts
-
-Find the Quickstart most appropriate for the platform and language you're using by going to the [Docs home page](/) and using the nagivator at the top of the page.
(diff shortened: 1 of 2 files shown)


khorne3:
@mpaktiti I thought about redirecting to /quickstarts but that goes straight to the home page. So, I left it as is, and removed the recursive link. It looked a bit strange for tutorials to redirect with no mention why. What do you think?

mpaktiti:
@khorne3 Tbh I don't see any point in this page. I think we should remove it altogether. Wdyt?

khorne3:
Sure. I see merits in both sides, but I can go ahead and remove it.

mpaktiti:
@khorne3 Go ahead and remove it, we can always restructure it in the future if we need it. I doubt we have any links towards this page but double check please. Thanks!

khorne3:
We might, but I'll redirect to the Docs home page. I originally thought to redirect to /quickstarts, but that goes to /, so I'll just skip the intermediate step.

PR #5193, opened 2017-10-24
Add indexes

khorne3:


articles/dashboard/index.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Management Dashboard
```
mpaktiti: Let's use `Dashboard` instead of `Management Dashboard` ([reference](https://github.com/auth0/docs/blob/master/STYLEGUIDE.md#the-dashboard))

Comment applies to the whole doc

articles/dashboard/index.md
```diff
@@ -0,0 +1,25 @@
...
+  <div data-name="example" class="topic-page-badge"></div>
+  <h1>Management Dashboard</h1>
+  <p>
+    When working in the Management Dashboard, you may find the following articles helpful.
+  </p>
+</div>
+
+<ul class="topic-links">
+  <li>
+    <i class="icon icon-budicon-715"></i><a href="/dashboard/creating-users-in-the-management-portal">Create Users in the Management Dashboard</a>
```
mpaktiti: Add a short description after each title. Something like the doc's `description` metadata should be ok.

articles/dashboard/index.md
```diff
@@ -0,0 +1,25 @@
...
+  </li>
+  <li>
+    <i class="icon icon-budicon-715"></i><a href="/dashboard/dashboard-tenant-settings">Manage Dashboard Administrators</a>
```
mpaktiti: Link throws 404

articles/monitoring/index.md
```diff
@@ -0,0 +1,34 @@
...
+  <div data-name="example" class="topic-page-badge"></div>
+  <h1>Monitor Your Auth0 Implementation</h1>
+  <p>
+    Learn how to monitor your Auth0 implementation and track your Auth0 usage, as well as how to send logs to your choice of logging suite.
+  </p>
+</div>
+
+<ul class="topic-links">
+  <li>
+    <i class="icon icon-budicon-715"></i><a href="/monitoring/how-to-monitor-auth0">Create Users in the Management Dashboard</a>
```
mpaktiti: Add a short description after each title

articles/security/index.md
```diff
@@ -17,7 +17,8 @@ description: Learn about various topics regarding how to handle sensitive data w
 
 <%= include('../_includes/_topic-links', { links: [
   'security/store-tokens',
-  'tokens'
+  'tokens',
+  'security/blacklisting-attributes'
```
mpaktiti: It's good that you added this article in the index, but maybe it should be on its own and not grouped under Tokens.

articles/users/index.md
```diff
@@ -0,0 +1,19 @@
...
+<!-- markdownlint-disable MD041 MD002 -->
+<div class="topic-page-header">
+  <div data-name="example" class="topic-page-badge"></div>
+  <h1>Users</h1>
+  <p>
+    Learn about working with users in Auth0.
+  </p>
+</div>
+
+<%= include('../_includes/_topic-links', { links: [
```
mpaktiti: You can add also all of the following I think:

image

I'm also thinking that this index could be the landing page when you click Articles > User Management. Wdyt?

mpaktiti: Add also this doc, it's a new one that Jerrie wrote and got deployed yesterday https://auth0.com/docs/connections/database/migrating-okta

mpaktiti (changes requested):

articles/dashboard/index.md
mpaktiti: You forgot to change this one (replace Management Dashboard with Dashboard)

articles/dashboard/index.md
mpaktiti: Missing description. Like this it does not look good, we might as well have a list of bullets:

image

articles/dashboard/index.md
mpaktiti: Same here, not fixed

articles/monitoring/index.md
mpaktiti: Missing description

articles/security/index.md
mpaktiti: Not fixed

articles/users/index.md
mpaktiti: Looks good but the list is long. How about grouping them?

For example:
- User Profile
  - User Profile
  - Structure of the User Profile
  - User Picture
  - Metadata
  - etc
- Import/Export
  - Bulk Import Users to Auth0
  - Migrate Users from Okta to Auth0
- etc

mpaktiti (changes requested):
Hey, not sure what went wrong here but you resubmitted this PR for review while you had addressed only one of the review comments. Some git mess-up I guess?

PR #5197, opened 2017-10-24
Add android Credential Manager docs

lbalmaceda:

### DIFF
--- a/articles/libraries/auth0-android/save-and-refresh-tokens.md
+++ b/articles/libraries/auth0-android/save-and-refresh-tokens.md
@@ -1,6 +1,7 @@
 ---
 section: libraries
 description: Keeping your user logged in with Auth0.Android
+toc: true
 ---
 # Auth0.Android Saving and Renewing Tokens
 
--- a/articles/libraries/auth0-android/save-and-refresh-tokens.md
+++ b/articles/libraries/auth0-android/save-and-refresh-tokens.md
@@ -53,7 +53,7 @@ manager.clearCredentials();
 //Show login screen
 ```
 
-### Obtain Credentials
+### Retrieving Credentials
 
 Because the credentials may need to be refreshed against Auth0 Servers, this method is asynchronous. Pass a callback implementation where you'd like to receive the credentials. Credentials returned by this method upon success are always valid.
 
@@ -72,7 +72,7 @@ manager.getCredentials(new BaseCallback<Credentials, CredentialsManagerException
 ```
 
 ::: note
-If the `accessToken` has expired, the manager will automatically use the `refreshToken` and renew the credentials for you.
+If the `accessToken` has expired, the manager will automatically use the `refreshToken` and renew the credentials for you. New credentials will be stored for future access.
 :::
 
 ### Save new Credentials
@@ -108,7 +108,7 @@ private static final int RC_UNLOCK_AUTHENTICATION = 123;
 boolean available = manager.requireAuthentication(this, RC_UNLOCK_AUTHENTICATION, getString(R.string.unlock_authentication_title), getString(R.string.unlock_authentication_description));
 ```
 
-If the feature was enabled, the manager will prompt the user to authenticate using the configured Lock Screen.  The result of this call will be obtained in the `onActivityResult` method of the activity passed before as first parameter. If the feature was not enabled, Lock Screen authentication will be skipped.
+If the feature was enabled, the manager will prompt the user to authenticate using the configured Lock Screen. The result of this call will be obtained in the `onActivityResult` method of the activity passed before as first parameter. If the feature was not enabled, Lock Screen authentication will be skipped.
 
 After checking that the received request code matches the one used in the configuration step, redirect the received parameters to the manager to finish the authentication. The credentials will be yield to the original callback.
 
--- a/articles/libraries/auth0-android/save-and-refresh-tokens.md
+++ b/articles/libraries/auth0-android/save-and-refresh-tokens.md
@@ -1,0 +1,117 @@
+---
+section: libraries
+description: Keeping your user logged in with Auth0.Android
+---
+
+# Auth0.Android Saving and Renewing Tokens
+
+When an authentication is performed with the `offline_access` scope included, it will return a [refresh token](/refresh-token) that can be used to request a new user token, without authenticating the user again.
+
+## Credentials Manager
+
+[Auth0.Android](https://github.com/auth0/Auth0.Android) provides a utility class to streamline the process of storing and renewing credentials. You can access the `accessToken` or `idToken` properties from the [Credentials](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/result/Credentials.java) instance. This is the preferred method to manage user credentials.
+
+First, add the library dependency to your build.gradle file:
+
+```gradle
+compile 'com.auth0.android:auth0:1.+'
+```
+
+Next, decide which class to use depending on your Android SDK target version:
+- If you require APIs lower than 21 (Lollipop) create a new instance of `CredentialsManager`. This class makes use of the `Context.MODE_PRIVATE` to store the credentials in a key-value preferences file, accessible only to your app. The data is stored in plaintext.
+- If you require APIs equal or greater than 21 (Lollipop) create a new instance of `SecureCredentialsManager`. This class has the same methods as the one above, but encrypts the data before storing it using a combination of _RSA_ and _AES_ algorithms along with the use of [Android KeyStore](https://developer.android.com/reference/java/security/KeyStore.html).
+
+
+## Using the CredentialsManager class
+
+### Setup
+Create a new instance by passing an `AuthenticationAPIClient` and a `Storage` implementation.
+
+```java
+Auth0 auth0 = new Auth0(this);
+AuthenticationAPIClient apiClient = new AuthenticationAPIClient(auth0);
+CredentialsManager manager = new CredentialsManager(apiClient, new SharedPreferencesStorage(this));
+```
+
+### Current State of the Authentication
+By asking for credentials existence upon launch and clearing them on logout, the manager can be used to manage the state of the authentication inside your app.
+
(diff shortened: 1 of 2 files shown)


articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
+---
+section: libraries
+description: Keeping your user logged in with Auth0.Android
+---
+
+# Auth0.Android Saving and Renewing Tokens
+
+When an authentication is performed with the `offline_access` scope included, it will return a [refresh token](/refresh-token) that can be used to request a new user token, without authenticating the user again.
```
cocojoe: without the user having to perform authentication again.

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+[Auth0.Android](https://github.com/auth0/Auth0.Android) provides a utility class to streamline the process of storing and renewing credentials. You can access the `accessToken` or `idToken` properties from the [Credentials](https://github.com/auth0/Auth0.Android/blob/master/auth0/src/main/java/com/auth0/android/result/Credentials.java) instance. This is the preferred method to manage user credentials.
+
+First, add the library dependency to your build.gradle file:
+
+```gradle
+compile 'com.auth0.android:auth0:1.+'
+```
+
+Next, decide which class to use depending on your Android SDK target version:
+- If you require APIs lower than 21 (Lollipop) create a new instance of `CredentialsManager`. This class makes use of the `Context.MODE_PRIVATE` to store the credentials in a key-value preferences file, accessible only to your app. The data is stored in plaintext.
```
cocojoe: Might be better to have the API as a header here

### API <21
blah

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+Create a new instance by passing an `AuthenticationAPIClient` and a `Storage` implementation.
+
+```java
+Auth0 auth0 = new Auth0(this);
+AuthenticationAPIClient apiClient = new AuthenticationAPIClient(auth0);
+CredentialsManager manager = new CredentialsManager(apiClient, new SharedPreferencesStorage(this));
+```
+
+### Current State of the Authentication
+By asking for credentials existence upon launch and clearing them on logout, the manager can be used to manage the state of the authentication inside your app.
```
cocojoe: Not sure you need this sentence, makes it sound a little confusing.

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+
+When you want to log out the user from your app, remove the stored credentials and direct them to the login screen.
+
+```java
+manager.clearCredentials();
+//Show login screen
+```
+
+### Obtain Credentials
+Because the credentials may need to be refreshed against Auth0 Servers, this method is a-synchronic. Pass a callback implementation where you'd like to receive the credentials. Credentials returned by this method upon success are always valid.
```
cocojoe: Do you need to talk about `a-synchronic` would just use `async` if you have to.
If it uses a callback isn't it implied to be async anyway?

Also didn't you suggest to me this should be called *Retrieving Credentials* 😄

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+
+    @Override
+    public void onFailure(CredentialsManagerException error) {
+        //No credentials were previously saved or they couldn't be refreshed
+    }
+});
+```
+
+::: note
+If the credentials were renewed, the manager will automatically save the new value for you.
```
cocojoe: Maybe expand a little here.

If the accessToken has expired, the manager will automatically use the refreshToken and renew the credentials for you.

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+:::
+
+### Save new Credentials
+You can save the credentials obtained upon log in in the manager. If they contain a valid `refresh_token` they will be refreshed automatically if they expire, in other case, they will be deemed invalid.
```
cocojoe: Don't think you need to talk about refresh token here, the section is about 'saving'.  The refresh stuff can be covered in the previous note? Can mention it also saves it automatically (assuming it does)

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+Auth0 auth0 = new Auth0(this);
+AuthenticationAPIClient apiClient = new AuthenticationAPIClient(auth0);
+SecureCredentialsManager manager = new SecureCredentialsManager(this, apiClient, new SharedPreferencesStorage(this));
+```
+
+The methods to obtain, save, check for existence and clearing the credentials are the ones explained before.
+
+
+### Pre-Authenticate the User
+This class can require the user to authenticate before being able to obtain the credentials by using the device's configured Lock Screen. If the Lock Screen Security is set to something different than PIN, Pattern, Password or Fingerprint, this feature won't be available. You need to call the method below to enable the authentication. Pass a valid `Activity` context, a request code, and 2 optional Strings to use as title and description for the Lock Screen.
```
cocojoe: This class provides optional functionality for additional authentication using the device's configured Lock Screen.

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+
+```java
+private static final int RC_UNLOCK_AUTHENTICATION = 123;
+
+//Called from an Activity
+boolean available = manager.requireAuthentication(this, RC_UNLOCK_AUTHENTICATION, getString(R.string.unlock_authentication_title), getString(R.string.unlock_authentication_description));
+```
+
+If the feature could be successfully enabled, when the manager is asked for the existing credentials it will automatically prompt the user to authenticate using the configured Lock Screen. The result of this call will be obtained in the `onActivityResult` method of the activity passed before as first parameter. If the feature couldn't be enabled, the authentication step is skipped.
```
cocojoe: If the features was enabled, the manager will prompt the user to authenticate using the configured Lock Screen.

If the feature was not enabled, Lock Screen authentication will be skipped.

cocojoe (changes requested):
So we going to ignore that your comments have no space between the `//` and the start of the comment?

articles/libraries/auth0-android/save-and-refresh-tokens.md
jeffreylees: And this word, as far as I know, is `asynchronous` if you aren't just using `async`.

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+### Current State of the Authentication
+By asking for credentials existence upon launch and clearing them on logout, the manager can be used to manage the state of the authentication inside your app.
+
+Stored credentials are considered valid if they have not expired or can be refreshed. Check if a user has already logged in:
+
+```java
+boolean loggedIn = manager.hasValidCredentials();
+```
+
+When you want to log out the user from your app, remove the stored credentials and direct them to the login screen.
```
jeffreylees: >When you want to log the user out of your app

Sounds better

articles/libraries/auth0-android/save-and-refresh-tokens.md
```diff
@@ -0,0 +1,117 @@
...
+```
+
+Next, decide which class to use depending on your Android SDK target version:
+- If you require APIs lower than 21 (Lollipop) create a new instance of `CredentialsManager`. This class makes use of the `Context.MODE_PRIVATE` to store the credentials in a key-value preferences file, accessible only to your app. The data is stored in plaintext.
+- If you require APIs equal or greater than 21 (Lollipop) create a new instance of `SecureCredentialsManager`. This class has the same methods as the one above, but encrypts the data before storing it using a combination of _RSA_ and _AES_ algorithms along with the use of [Android KeyStore](https://developer.android.com/reference/java/security/KeyStore.html).
+
+
+## Using the CredentialsManager class
+
+### Setup
```
jeffreylees: You need new lines after headings all through

jeffreylees (approved):
I did a few minor copyediting changes, added it to the `sidebar.yml`, and also I think I did most of @cocojoe feedback while I was doing it.

cocojoe:
@jeffreylees is the problem deploying related to the PR or something else ?

jeffreylees:
Heroku builds are broken again I think I read that earlier in Slack - I'll check the log just to make sure.

articles/libraries/auth0-android/save-and-refresh-tokens.md
lbalmaceda: I prefer asynchronous. Yes, it's implied but I think it's good to have a bit more text there just to intro the usage, don't you? I'll change it to "retrieving credentials".

articles/libraries/auth0-android/save-and-refresh-tokens.md
lbalmaceda: I think it's the other way around. We should talk about the refresh logic in the "retrieve" rather than in the save. Save doesn't do refresh, and will accept credentials without refresh_token just fine. I'll add in the previous step something to indicate that we store the refreshed credentials.

lbalmaceda:
Thank you guys. I'll wait for this to deploy to check the current output.

jeffreylees:
Heroku builds are back:

https://auth0-docs-content-pr-5197.herokuapp.com/docs/libraries/auth0-android/save-and-refresh-tokens

lbalmaceda:
@jeffreylees looks good overall.  @cocojoe Any last suggestions?

cocojoe (approved):
LGTM

PR #5199, opened 2017-10-25
fixed golang example: added missing slash / at issuer url

tglander:
without the slash, authentication fails with the error "invalid issuer claim (iss)". WIth the slash, it works.
The python example also has the trailing slash.


aaguiarz (approved):

aaguiarz:
Fixed, thanks!

PR #5233, opened 2017-11-02
3di changes to the WebApp quickstarts: PHP, ASP.NET and ASP.NET Core

awie3di:

### DIFF
--- a/articles/quickstart/webapp/aspnet-core/_includes/_login.md
+++ b/articles/quickstart/webapp/aspnet-core/_includes/_login.md
@@ -161,7 +161,7 @@ Add the `Login` and `Logout` actions to `AccountController`.
 
 To add the `Login` action, call `ChallengeAsync` and pass "Auth0" as the authentication scheme. This will invoke the OIDC authentication handler you registered in the `ConfigureServices` method.
 
-After the OIDC middleware signs the user in, the user is also automatically signed in to the cookie middleware. This allows the user to be authenticated later. 
+After the OIDC middleware signs the user in, the user is also automatically signed in to the cookie middleware. This allows the user to be authenticated on subsequent requests. 
 
 For the `Logout` action, you need to sign the user out of both middlewares.
 
--- a/articles/quickstart/webapp/aspnet-core/_includes/_login.md
+++ b/articles/quickstart/webapp/aspnet-core/_includes/_login.md
@@ -250,7 +250,7 @@ public void ConfigureServices(IServiceCollection services)
 
 ## Add the Log In and Log Out Buttons
 
-Add the **Log In** and **Log Out** buttons to the navigation bar. In the `/Views/Shared/_Layout.cshtml` file, in the navigation bar section, add code that displays the **Log Out** button when the user is authenticated and the **Log In** button in any other case. The buttons link to the `Logout` and `Login` actions in the `AccountController`:  
+Add the **Log In** and **Log Out** buttons to the navigation bar. In the `/Views/Shared/_Layout.cshtml` file, in the navigation bar section, add code that displays the **Log Out** button when the user is authenticated and the **Log In** button if not. The buttons link to the `Logout` and `Login` actions in the `AccountController`:  
 
 ```html
 <!-- Views/Shared/_Layout.cshtml -->
--- a/articles/quickstart/webapp/aspnet-core/_includes/_login.md
+++ b/articles/quickstart/webapp/aspnet-core/_includes/_login.md
@@ -199,7 +199,7 @@ ASP.NET Core calls `SignOutAsync` for the "Auth0" authentication scheme. You nee
 
 When the application calls `SignOutAsync` for the OIDC middleware, it also calls the `/v2/logout` endpoint of the Auth0 Authentication API. The user is logged out of Auth0.
 
-If you specify the URL for redirecting users in the **Allowed Logout URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings), it is passed to the `returnTo` parameter. 
+If you specify the `returnTo` parameter, the users will be redirected there after they are logged out. Specify the URL for redirecting users in the **Allowed Logout URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings).
 
 In the `Startup.cs` file, update the call to `AddOpenIdConnect` with the following code:
 
--- a/articles/quickstart/webapp/aspnet-core/_includes/_login.md
+++ b/articles/quickstart/webapp/aspnet-core/_includes/_login.md
@@ -8,11 +8,18 @@ You can also create a custom login for prompting the user for their username and
 
 ## Configure OpenID Connect Middleware
 
-The easiest way to enable authentication with Auth0 in your ASP.NET Core application is to use the OpenID Connect middleware. First, go to the `ConfigureServices` method of your `Startup` class and add the authentication services and enable cookie autentication by calling the `AddAuthentication` and `AddCookie` methods respectively. 
+To enable authentication in your ASP.NET Core application, use the OpenID Connect (OIDC) middleware.
+Go to the `ConfigureServices` method of your `Startup` class. To add the authentication services, call the `AddAuthentication` method. To enable cookie authentication, call the `AddCookie` method. 
 
-Next you can add a call to `AddOpenIdConnect` to configure the OpenID Connect authentication handler. Configure the authentication scheme by passing "Auth0" as the `authenticationScheme` parameter. This is important, as you will later on use this value to challenge the OIDC middleware.
+Next, configure the OIDC authentication handler. Add a call to `AddOpenIdConnect`. To configure the authentication scheme, pass "Auth0" as the `authenticationScheme` parameter. You will use this value later to test the OIDC middleware.
 
-Also configure the `ClientId`, `ClientSecret`, `ResponseType` and other parameters as per the code sample below:
+Configure other parameters, such as `ClientId`, `ClientSecret` or `ResponseType`. 
+
+By default, the OIDC middleware requests both the `openid` and `profile` scopes. Because of that, you may get a large ID token in return. We suggest that you ask only for the scopes you need. You can read more about requesting additional scopes in the [User Profile step](/quickstart/webapp/aspnet-core/v2/04-user-profile).
+
+::: note
+In the code sample below, only the `openid` scope is requested.
+:::
 
 ```cs
 // Startup.cs
@@ -80,9 +87,7 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
-Note in the code sample above that the scopes are cleared and only the `openid` scope is requested. By default the OIDC middleware will request both the `openid` and the `profile` scopes and can result in a large `id_token` being returned. It is suggested that you be more explicit about the scopes you want to be returned and not ask for the entire profile to be returned. Requesting additional scopes is discussed later in the [User Profile step](/quickstart/webapp/aspnet-core/v2/04-user-profile). 
-
-Next, in the `Configure` method of the `Startup` class, ensure that you add a call to `UseAuthentication` to add the authentication middleware:
+Next, add the authentication middleware. In the `Configure` method of the `Startup` class, call the `UseAuthentication` method.
 
 ```csharp
 // Startup.cs
@@ -113,11 +118,11 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 }
 ```
 
-### Obtaining an Access Token for calling an API
+## Obtain an Access Token for Calling an API
 
-You may want to call an API from your MVC application, in which case you need to obtain an `access_token` which was issued for the particular API you want to call. In this case you will need to pass an extra `audience` parameter containing the API Identifier to the Auth0 authorization endpoint. 
+If you want to call an API from your MVC application, you need to obtain an access token issued for the API you want to call. To obtain the token, to the Auth0 authorization endpoint, pass an additional `audience` parameter containing the API identifier.
 
-If you want to do this, simply handle the `OnRedirectToIdentityProvider` event when configuring the `OpenIdConnectOptions` object, and add the `audience` parameter to the `ProtocolMessage`
+In the configuration for the `OpenIdConnectOptions` object, handle the `OnRedirectToIdentityProvider` event and add the `audience` parameter to `ProtocolMessage`.
 
 ```csharp
 // Startup.cs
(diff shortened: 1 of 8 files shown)


articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -8,11 +8,18 @@ You can also create a custom login for prompting the user for their username and
...
 
-The easiest way to enable authentication with Auth0 in your ASP.NET Core application is to use the OpenID Connect middleware. First, go to the `ConfigureServices` method of your `Startup` class and add the authentication services and enable cookie autentication by calling the `AddAuthentication` and `AddCookie` methods respectively. 
+To enable authentication in your ASP.NET Core application, use the OpenID Connect (OIDC) middleware.
+Go to the `ConfigureServices` method of your `Startup` class. To add the authentication services, call the `AddAuthentication` method. To enable cookie authentication, call the `AddCookie` method. 
 
-Next you can add a call to `AddOpenIdConnect` to configure the OpenID Connect authentication handler. Configure the authentication scheme by passing "Auth0" as the `authenticationScheme` parameter. This is important, as you will later on use this value to challenge the OIDC middleware.
+Next, configure the OIDC authentication handler. Add a call to `AddOpenIdConnect`. To configure the authentication scheme, pass "Auth0" as the `authenticationScheme` parameter. You will use this value later to test the OIDC middleware.
```
jerriep: > pass "Auth0" as the...

Should it not be 

> passing "Auth0" as the...

jerriep: > You will use this value later to test the...

No. **test** is not the correct word. It should be **challenge** as it originally was. You **challenge** authentication middleware, not **test** it.

So it should read:

> You will use this value later to challenge the...

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -113,11 +118,11 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 }
 ```
 
-### Obtaining an Access Token for calling an API
+## Obtain an Access Token for Calling an API
 
-You may want to call an API from your MVC application, in which case you need to obtain an `access_token` which was issued for the particular API you want to call. In this case you will need to pass an extra `audience` parameter containing the API Identifier to the Auth0 authorization endpoint. 
+If you want to call an API from your MVC application, you need to obtain an access token issued for the API you want to call. To obtain the token, to the Auth0 authorization endpoint, pass an additional `audience` parameter containing the API identifier.
```
jerriep: >  To obtain the token, to the Auth0 authorization endpoint, pass an additional `audience` parameter containing the API identifier.

This reads very strangely. It should be

>  To obtain the token, pass an additional `audience` parameter containing the API identifier to the Auth0 authorization endpoint.

Because the `audience` parameter is passed to the endpoint. The way you are wording it, it sounds as though you are obtaining an access token to the endpoint. Which is not correct.

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -146,19 +151,21 @@ public void ConfigureServices(IServiceCollection services)
...
+## Add the `Login` and `Logout` Methods
 
-## Add Login and Logout Methods
+Add the `Login` and `Logout` actions to `AccountController`. 
 
-Next, you will need to add `Login` and `Logout` actions to the `AccountController`. 
+To add the `Login` action, call `ChallengeAsync` and pass "Auth0" as the authentication scheme. This will invoke the OIDC authentication handler you registered in the `ConfigureServices` method.
 
-For the Login, you will need to call `ChallengeAsync`, passing "Auth0" as the authentication scheme which will be challenged. This will invoke the OIDC authentication handler you registered in the `ConfigureServices` method.
+After the OIDC middleware signs the user in, the user is also automatically signed in to the cookie middleware. This allows the user to be authenticated later.
```
jerriep: > This allows the user to be authenticated later

Please keep this as 

> This allows the user to be authenticated on subsequent requests.

The way you worded it makes it sounds like when they come back at a later stage. But this is not the case. It should be made clear the all subsequent requests which are part of the current session will be authenticated by the cookie middleware.

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -188,9 +195,13 @@ public class AccountController : Controller
 }
 ```
 
-At this point ASP.NET Core will call `SignOutAsync` for the "Auth0" authentication scheme (i.e. the OIDC authentication handler), but the OIDC middleware does not know what the actual Logout URL is it should call to log the user out of Auth0. To do this you should handle the `OnRedirectToIdentityProviderForSignOut` event when registering the OIDC authentication handler.
+ASP.NET Core calls `SignOutAsync` for the "Auth0" authentication scheme. You need to provide the OIDC middleware with the URL for logging the user out of Auth0. To get the URL, handle the `OnRedirectToIdentityProviderForSignOut` event when you register the OIDC authentication handler.
```
jerriep: > To get the URL, 

should be

> To set the URL, 

You **set** the URL, not **get** it

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -188,9 +195,13 @@ public class AccountController : Controller
...
+
+When the application calls `SignOutAsync` for the OIDC middleware, it also calls the `/v2/logout` endpoint of the Auth0 Authentication API. The user is logged out of Auth0.
```
jerriep: > When the application calls `SignOutAsync` for the OIDC middleware, it also calls the `/v2/logout` endpoint of the Auth0 Authentication API. The user is logged out of Auth0.

Why are you breaking this up into two sentences? This does not read nicely. Please make it 

> When the application calls `SignOutAsync` for the OIDC middleware, it also calls the `/v2/logout` endpoint of the Auth0 Authentication API, logging the user out of Auth0.

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -188,9 +195,13 @@ public class AccountController : Controller
...
 
-So back in the `Startup.cs` file, update the call to `AddOpenIdConnect` with the following code:
+If you specify the URL for redirecting users in the **Allowed Logout URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings), it is passed to the `returnTo` parameter.
```
jerriep: > If you specify the URL for redirecting users in the **Allowed Logout URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings), it is passed to the `returnTo` parameter. 

No. This completely changes the meaning of this sentence. It should be 

> If you specify the `returnTo` parameter, the user will be redirected there after being logged out. It is important to specify this URL in the **Allowed Logout URLs** field in your [Client Settings](${manage_url}/#/applications/${account.clientId}/settings)

articles/quickstart/webapp/aspnet-core/_includes/_login.md
```diff
@@ -237,13 +248,9 @@ public void ConfigureServices(IServiceCollection services)
...
 
-This will ensure that when `SignOutAsync` is called for the OIDC Middleware, that the `/v2/logout` endpoint of the Auth0 Authentication API is called to log the user out of Auth0. 
-
-It will also pass along the Redirect URL (when specified) in the `returnTo` parameter. You must therefore ensure that you have specified this URL in the **Allowed Logout URLs** for your Client in the Auth0 Dashboard.
-
-## Add Login and Logout Links
+## Add the Log In and Log Out Buttons
 
-Lastly, add Login and Logout links to the navigation bar. To do that, head over to `/Views/Shared/_Layout.cshtml` and add code to the navigation bar section which displays a Logout link when the user is authenticated, otherwise a Login link. These will link to the `Logout` and `Login` actions of the `AccountController` respectively:  
+Add the **Log In** and **Log Out** buttons to the navigation bar. In the `/Views/Shared/_Layout.cshtml` file, in the navigation bar section, add code that displays the **Log Out** button when the user is authenticated and the **Log In** button in any other case. The buttons link to the `Logout` and `Login` actions in the `AccountController`:
```
jerriep: > ...and the **Log In** button in any other case.

There are only 2 cases: either authenticated or not. So I think rather change this to 

> and the **Log In** button if not.

articles/quickstart/webapp/aspnet-core/v2/00-intro.md
```diff
@@ -5,26 +5,27 @@ description: Introduction to the quickstart guide and configuring the environmen
...
-::: panel System Requirements
+::: panel System requirements
 This tutorial and seed project have been tested with the following:
-
 * .NET Core SDK 2.0
 * .NET Core 2.0
 * ASP.NET Core 2.0
 
-This tutorial can be completed with the command line tools and your code editor of choice, or alternatively you can use Microsoft Visual Studio 2017 Update 3. For more details on how to use .NET Core on your platform, please see [the .NET Core Website](https://www.microsoft.com/net/core).
+To complete this tutorial, you can use command line tools and any code editor. Alternatively, you can use Microsoft Visual Studio 2017 Update 3. For more details on how to use .NET Core on your platform, read the [.NET Core](https://www.microsoft.com/net/core) Microsoft documentation.
```
jerriep: > read the [.NET Core](https://www.microsoft.com/net/core) Microsoft documentation.

remove the word **Microsoft** from this sentence

articles/quickstart/webapp/aspnet-core/v2/02-storing-tokens.md
```diff
@@ -16,15 +16,15 @@ budicon: 280
   ]
 }) %>
 
-The OIDC middleware in ASP.NET Core will automatically Decode the ID Token returned from Auth0 and will automatically add the claims contained in the ID Token as claims on the `ClaimsIdentity`.
+The OIDC middleware in ASP.NET Core automatically decodes the ID token returned from Auth0 and adds the claims from the ID token as claims in `ClaimsIdentity`.
```
jerriep: > as claims in `ClaimsIdentity`.

add **the**

> as claims in the `ClaimsIdentity`.

articles/quickstart/webapp/aspnet-core/v2/03-user-profile.md
```diff
@@ -16,15 +16,15 @@ budicon: 292
   ]
 }) %>
 
-## Getting the Profile
+## Get the Profile
 
-As mentioned in the previous step, the OIDC middleware will automatically extract the user's information from the ID Token and add it as claims to the `ClaimsIdentity`.
+The OIDC middleware extracts the user's information from the ID token and adds it as a claim to the `ClaimsIdentity`.
```
jerriep: > The OIDC middleware extracts the user's information from the ID token and adds it as a claim to the `ClaimsIdentity`.

It adds multiple **claims**. Change to 

> The OIDC middleware extracts the user's information from the ID token and adds it as claims to the `ClaimsIdentity`.

articles/quickstart/webapp/aspnet-core/v2/03-user-profile.md
```diff
@@ -39,7 +39,7 @@ public class UserProfileViewModel
 }
 ```
 
-Add a new `Profile` action to the `AccountController` and extract the relevant claims and add them to a new instance of `UserProfileViewModel` which is then passed to the view. Be sure to dectorate the action with the `[Authorize]` attribute so only authenticated users can access the action:
+Add a new `Profile` action to the `AccountController` and extract the relevant claims. Add the claims to a new instance of `UserProfileViewModel` passed to the view. Add the `[Authorize]` attribute to the action so only authenticated users can access the action:
```
jerriep, line 42: > Add the claims to a new instance of `UserProfileViewModel` passed to the view.

Change to

> Add the claims to a new instance of `UserProfileViewModel` which is passed to the view.

articles/quickstart/webapp/aspnet-core/v2/03-user-profile.md
```diff
@@ -56,11 +56,11 @@ public IActionResult Profile()
 }
 ```
 
-The `User.Identity.Name` property will look for a claim of type `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name` on the user object. Auth0 passes the name of the user in the `name` claim of the `id_token`, but this does not get automatically matched the the `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name` claim type, and therefore `User.Identity.Name` will return null.
+The `User.Identity.Name` property looks for a claim of a type `http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name` on the user object. Auth0 passes the name of the user in the `name` claim of the ID token. The user's name is not matched to this claim type, which means that `User.Identity.Name` will return null.
```
jerriep: No. You are completely changing the meaning of this paragraph. Please change back to the way it was. Or if you need to change it, make sure it keeps its meaning. What you have written here does not make sense!

articles/quickstart/webapp/aspnet-core/v2/04-authorization.md
```diff
@@ -16,15 +16,21 @@ budicon: 546
...
 
-### Create a Rule to assign roles
+## Create a Rule to Assign Roles
 
-First, we will create a rule that assigns our users either an `admin` role, or a `user` role based on the email domain. To do so, go to the [new rule page](${manage_url}/#/rules/new) and create an empty rule. Then use the following code for your rule:
+Create a rule that assigns the following access roles to your user:
+* An admin role
+* A regular user role
+
+To aasign roles, go to the [New rule page](${manage_url}/#/rules/new). In the **Access Control** section, create an empty rule.
```
jerriep: > To aasign

Should be 

> To **assign**

articles/quickstart/webapp/aspnet-core/v2/04-authorization.md
```diff
@@ -97,9 +107,9 @@ public void ConfigureServices(IServiceCollection services)
 }
 ```
 
-At this point the you have integrated with the the [Role based authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles) mechanism of ASP.NET Core, which means that your can ensure that a user belongs to a particular role by simply decorating your controller actions with the `[Authorize(Roles = ?)]` attribute.
+Because you have integrated your application with the [Role based authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles) mechanism, you can check which role to assign to a user. Add the `[Authorize(Roles = ?)]` attribute to your controller action.
```
jerriep: Once again, you have completely changed the meaning of the original paragraph.

> you can check which role to assign to a user

is wrong! You are checking which roles are are assigned to a user, not which role to assign to them.

Change to something like

> You can use the [Role based authorization](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/roles) mechanism to ensure that only users who belong to a specific role can access certain actions. To do this, add the `[Authorize(Roles = ?)]` attribute to your controller action.

jerriep (changes requested):
@awie3di It seems that many of the sentences were "fixed" by some sort of grammar checker. But this is problematic, since it changed the actual meaning of many things. 

Please see my comments

jerriep (approved):

PR #5238, opened 2017-11-03
add password grant notes for android articles

lbalmaceda:
Add warning panels to enable Password Grant Type on embedded Lock.

### DIFF
--- a/articles/libraries/auth0-android/index.md
+++ b/articles/libraries/auth0-android/index.md
@@ -282,6 +282,10 @@ To ensure an Open ID Connect compliant response you must either request an `audi
 
 ### Login with database connection
 
+::: panel-warning Database authentication on Native Platforms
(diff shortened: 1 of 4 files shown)


articles/libraries/auth0-android/index.md
```diff
@@ -282,6 +282,10 @@ To ensure an Open ID Connect compliant response you must either request an `audi
 
 ### Login with database connection
 
+::: panel-warning Database authentication on Native Platforms
```
jeffreylees, line 285: Is there a reason you went with a `panel` instead of a `warning` element for this one since you did with the others below? You're probably at about the size of warning where you could use either, but I'd use one consistently since it's the same message.

jeffreylees (changes requested):

articles/libraries/auth0-android/index.md
lbalmaceda: I did it for consistency, because some lines below "### Passwordless login" section has the same panel-warning for a similar message.

jeffreylees: Fair enough. Typically the rule of thumb is `warning` or `note` are for short-form content like 4 lines or less, `panel-foo` are for longer form, or for things broken up into more than one paragraph.

In any case, this seems fine for now, was just checking.

jeffreylees (approved):

aaguiarz (approved):

PR #5243, opened 2017-11-07
Fix syntax error in SSOCircle doc code

thameera:


mpaktiti:
Thank you @thameera ! 👍

PR #5252, opened 2017-11-09
Add Landing Page to /compliance

khorne3:
https://auth0-docs-content-pr-5252.herokuapp.com/docs/compliance


mpaktiti:
@khorne3 I'm merging this but I think that the section headers are confusing because they do look like a link but they do not work. If I was a user I'd assume that something is broken in the page. 

My 2c, add the actual links at the headers, ie link `Privacy Shield` to https://www.privacyshield.gov/participant?id=a2zt000000001VJAAY&status=Active, `HIPAA` to https://www.hhs.gov/hipaa/index.html and the `SOC 2` to a relevant link

PR #5258, opened 2017-11-11
Add missing information about Allowed Logout URL

netchkin:
When following this tutorial https://auth0.com/docs/quickstart/webapp/aspnet-core/v2/01-login the application I built did not logout properly due to returnTo parameter not being allowed for logout. I found solution here: https://auth0.com/docs/logout#set-the-allowed-logout-urls-at-the-client-level and thought that we could prevent users from facing this issue when following the .NET core MVC tutorial


netchkin:
I am sorry, I found out that the next step actually specifies the need to add the url to the list of Allowed Logout URLs. While I still think it should be more visible, I am not sure if this edit is the right way to do it, hence closing.

PR #5259, opened 2017-11-12
API changes and JSON compatibility

wspringer:
I couldn't get it to work the v1 API, but when changing the URL to use the v2 endpoint (and specifying the method to be POST) it worked. Also, JSON is supposed to use double quotes instead of single quotes.


mpaktiti:
Thank you @wspringer! I will push one more change to this (`-H 'Authorization: Bearer YOUR_GLOBAL_CLIENT_ACCESS_TOKEN'` should use a [Management APIv2 access token](https://auth0.com/docs/api/management/v2/tokens)) and merge.

Happy holidays!

PR #5266, opened 2017-11-14
Correct Express version

export-mike:
14 doesn't exist?


mpaktiti:
Oops, thank you for fixing that @export-mike ! 👍 

cc @aaguiarz

PR #5267, opened 2017-11-14
Update index.md

sallareznov:


mpaktiti:
Thanks for fixing that @sallareznov ! 👍

PR #5274, opened 2017-11-16
Add simple example for daemonizing with systemd

sylnsr:


mpaktiti:
Looks good, thank you @sylnsr ! 👍

PR #5276, opened 2017-11-16
Auth0-to-Auth0 SAML: use 'Switch tenant' option

thameera:
Also added missing space


mpaktiti:
Thank you @thameera ! 👍

PR #5277, opened 2017-11-16
Dashboard index: Add new entry for Logs

mpaktiti:


articles/dashboard/index.md
```diff
@@ -15,14 +15,18 @@ classes: topic-page
 <ul class="topic-links">
   <li>
     <i class="icon icon-budicon-715"></i><a href="/dashboard/creating-users-in-the-management-portal">Create Users in the Dashboard</a>
+    <p>How to create end users using the Dashboard</p>
```
mpaktiti, line 18: @khorne3 For future reference, the `p` element should be within the `li`. See the diff below.

Nested within:

image

When put after the li:
image

PR #5281, opened 2017-11-16
Update android QS

lbalmaceda:
This PR:

- Use secure `CredentialsManager` to handle credentials
- Call /userinfo to obtain the profile data required given an access token, since we don't verify the id_tokens locally in our SDKs yet.
- Create instances of the `UsersAPIClient` using an access token instead of an id token.

### DIFF
--- a/articles/quickstart/native/android/03-session-handling.md
+++ b/articles/quickstart/native/android/03-session-handling.md
@@ -21,7 +21,7 @@ This tutorial shows you how to let users log in and maintain an active session w
 You need the `Credentials` class to handle users' credentials. The class is composed of these elements:
 
 * `accessToken`: Access Token used by the Auth0 API. To learn more, see the [Access Token documentation](/tokens/access-token).
-* `idToken`: Identity token that proves the identity of the user. To learn more, see the [ID Token documentation](/tokens/id-token).
+* `idToken`: Identity Token that proves the identity of the user. To learn more, see the [ID Token documentation](/tokens/id-token).
 * `refreshToken`: Refresh Token that can be used to request new tokens without signing in again. To learn more, see the [Refresh Token documentation](/tokens/refresh-token/current).
 * `tokenType`: The type of tokens issued by the server.
 * `expiresIn`: The number of seconds before the tokens expire.
@@ -52,11 +52,11 @@ WebAuthProvider.init(auth0)
 
 ## Check for Tokens when the Application Starts
 
-::: panel Learn about refresh tokens
+::: panel Learn about Refresh Tokens
 Before you go further with this tutorial, read the [refresh token documentation](/refresh-token).
 It is important that you remember the following:
-* Refresh tokens must be securely saved.
-* Even though refresh tokens cannot expire, they can be revoked.
+* Refresh Tokens must be securely saved.
+* Even though Refresh Tokens cannot expire, they can be revoked.
 * New tokens will have the same scope as was originally requested during the first authentication.
 :::
 
--- a/articles/quickstart/native/android/03-session-handling.md
+++ b/articles/quickstart/native/android/03-session-handling.md
@@ -57,13 +57,13 @@ Before you go further with this tutorial, read the [refresh token documentation]
 It is important that you remember the following:
 * Refresh tokens must be securely saved.
 * Even though refresh tokens cannot expire, they can be revoked.
-* New tokens will never have a different scope than the scope you requested during the first login.
+* New tokens will have the same scope as was originally requested during the first authentication.
 :::
 
 You can simplify the way you handle user sessions using a Credential Manager class, which knows how to securely store, retrieve and renew credentials obtained from Auth0. Two classes are provided in the SDK to help you achieve this. Further read on how they work and their implementation differences is available in the [Saving and Renewing Tokens](/libraries/auth0-android/save-and-refresh-tokens.md) article. For this series of tutorials we're going to use the `SecureCredentialsManager` class as it encrypts the credentials before storing them in a private SharedPreferences file.
 
 
-Create a new instance of the Credentials Manager. When you run the application you'd first check if there are any previous saved credentials, in order to skip the login screen:
+Create a new instance of the Credentials Manager. When you run the application, you should check if there are any previously stored credentials. You can use these credentials to bypass the login screen:
 
 ```java
 // app/src/main/java/com/auth0/samples/LoginActivity.java
@@ -84,13 +84,13 @@ Create a new instance of the Credentials Manager. When you run the application y
 ```
 
 ::: note
-Ideally a single class should have the knowledge of handling credentials, but you can share this instance across activities or create a new one every time is required as long as the Storage strategy persists the data in the same location. Check the `LoginActivity` class to understand how to achieve this in a single class.
+Ideally a single class should manage the handling of credentials. You can share this instance across activities or create a new one every time is required as long as the Storage strategy persists the data in the same location. Check the `LoginActivity` class to understand how to achieve this in a single class.
 :::
 
 
 ## Save the User's Credentials
 
-After a successful login response save the user's credentials on the Credentials Manager. Use the `saveCredentials` method.
+After a successful login response, you can store the user's credentials using the `saveCredentials` method.
 
 ```java
 // app/src/main/java/com/auth0/samples/LoginActivity.java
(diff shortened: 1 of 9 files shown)


articles/quickstart/native/android/05-linking-accounts.md
```diff
@@ -116,33 +139,20 @@ The process to link accounts requires the logged-in user's ID and the id tokens
...
-
-      @Override
-      public void onFailure(CredentialsManagerException error) {
-          //Credentials probably expired. Remove them and login again
-          credentialsManager.clearCredentials();
-          finish();
-      }
-  });
+    UsersAPIClient client = new UsersAPIClient(auth0, getIntent().getExtras().getString(KEY_ID_TOKEN));
+    String primaryUserId = getIntent().getExtras().getString(KEY_PRIMARY_USER_ID);
```
aaguiarz: The Management API endpoints now receive an access_token as a bearer and not an id_token, how is this working in the Android SDK?

aaguiarz (commented):
Added a comment

articles/quickstart/native/android/05-linking-accounts.md
lbalmaceda: Both our native SDKs use the same endpoints with an id_token, and they both work ATM. No need to hold this PR because of something that hasn't been announced yet. 👍

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -47,21 +47,55 @@ WebAuthProvider.init(auth0)
...
 
-## Save the User's Credentials
+## Check for Tokens when the Application Starts
+
+::: panel Learn about refresh tokens
+Before you go further with this tutorial, read the [refresh token documentation](/refresh-token).
+It is important that you remember the following:
+* Refresh tokens must be securely saved.
+* Even though refresh tokens cannot expire, they can be revoked.
+* New tokens will never have a different scope than the scope you requested during the first login.
```
cocojoe: New tokens will have the same scope as was originally requested during the first authentication

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -47,21 +47,55 @@ WebAuthProvider.init(auth0)
...
+:::
+
+You can simplify the way you handle user sessions using a Credential Manager class, which knows how to securely store, retrieve and renew credentials obtained from Auth0. Two classes are provided in the SDK to help you achieve this. Further read on how they work and their implementation differences is available in the [Saving and Renewing Tokens](/libraries/auth0-android/save-and-refresh-tokens.md) article. For this series of tutorials we're going to use the `SecureCredentialsManager` class as it encrypts the credentials before storing them in a private SharedPreferences file.
+
+
+Create a new instance of the Credentials Manager. When you run the application you'd first check if there are any previous saved credentials, in order to skip the login screen:
```
cocojoe: ~When you run the application you'd first check~  -> When you run the application, you should check if there are any previously stored credentials.  You can use these credentials to bypass the login screen.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -47,21 +47,55 @@ WebAuthProvider.init(auth0)
...
+    // Prompt Login screen.
+  } else {
+    // Obtain credentials and move to the next activity
+  }
+```
 
-::: warning
-Make sure you use a secure method.
+::: note
+Ideally a single class should have the knowledge of handling credentials, but you can share this instance across activities or create a new one every time is required as long as the Storage strategy persists the data in the same location. Check the `LoginActivity` class to understand how to achieve this in a single class.
```
cocojoe: ~Ideally a single class should have the knowledge of handling credentials~, -> +Ideally a single class should manage the handling of credentials.  You can share

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -47,21 +47,55 @@ WebAuthProvider.init(auth0)
...
 :::
 
+
+## Save the User's Credentials
+
+After a successful login response save the user's credentials on the Credentials Manager. Use the `saveCredentials` method.
```
cocojoe: ~After a successful login response save the user's credentials on the Credentials Manager. Use the `saveCredentials` method.~
After a successful login response, you can save the user's credentials using the `saveCredentials` method.

I prefer the word store instead of save.  However, as the method is called `saveCredentials` it is okay.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -74,133 +108,67 @@ private final AuthCallback callback = new AuthCallback() {
...
-        saveCredentials(credentials);
+        credentialsManager.saveCredentials(credentials);
         //...
     }
 };
 ```
 
 ::: note
-User credentials are stored in [Private mode](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE) in the seed project in the `SharedPreferences` file. You can achieve this with the `CredentialsManager` class. For details, check the implementation in the project code. There are better and more secure ways to store tokens, but we will not cover them in this tutorial.
+The Storage implementation given to the Credentials Manager in the seed project uses a SharedPreferences file to store the user credentials in [Private mode](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE). Change this behavior by implementing a custom Storage.
```
cocojoe: ~Change this behavior by implementing a custom Storage.~
You can modify this behavior by implementing a custom Storage ??

What is Storage defined as, a class, an object etc ? Sounds like this should be added after.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -74,133 +108,67 @@ private final AuthCallback callback = new AuthCallback() {
...
+The Storage implementation given to the Credentials Manager in the seed project uses a SharedPreferences file to store the user credentials in [Private mode](https://developer.android.com/reference/android/content/Context.html#MODE_PRIVATE). Change this behavior by implementing a custom Storage. 
 :::
 
-## Check for Tokens when the Application Starts
-
-To save your users the effort of logging in every time they open your app, use their access tokens. You can check for the access token when a user starts the app. If you find the token, you can automatically log the user in and direct them straight to your app's main flow.  
+## Recover the User's Credentials
 
+Retrieving the credentials from the Credentials Manager is an async process, as credentials may have expired and require to be refreshed. This renewing process is done automatically by the Credentials Manager as long as a valid refresh token is currently stored. A `CredentialsManagerException` exception will raise if the credentials have expired and the network request to refresh them has failed.
```
cocojoe: ~require to be refreshed.  This renewing process is done automatically~
require renewal. The renewal process is performed automatically

~A `CredentialsManagerException` exception will raise if the credentials have expired and the network request to refresh them has failed.~
A `CredentialsManagerException` exception will be raised if the credentials can not be renewed.

articles/quickstart/native/android/03-session-handling.md
```diff
@@ -74,133 +108,67 @@ private final AuthCallback callback = new AuthCallback() {
...
 ::: note
-Depending on the way you store users' credentials, you delete them differently. 
-:::
-
-## Optional: Encapsulate Session Handling
-
-You can simplify the way you handle user sessions by storing token-related information and processes in a class. The class separates the logic for handling user sessions from the activity. 
-
-We recommend that you download the sample project from this tutorial and look at its implementation. Focus on the `CredentialsManager` class, which manages session handling, obtains user credentials from the `SharedPreferences` file and saves them.
+If you are not using our Credentials Manager classes, depending on the way you store users' credentials you delete them differently.
```
cocojoe: ~If you are not using our Credentials Manager classes, depending on the way you store users' credentials you delete them differently. ~
If you are not using our Credentials Manager classes, you are responsible for ensuring that the user's credentials have been removed.

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -37,67 +37,57 @@ WebAuthProvider.init(auth0)
...
-    });
+private String getUserId(String idToken) {
+    return new JWT(idToken).getSubject();
+}
 ```
 
-When you get the `sub` value, call the [Management API](https://auth0.com/docs/api/management/v2#!/Users).
-
-Use the `UsersAPIClient` client and the user's ID to get the full user profile.
+Use the user id to call `getProfile` on the Users API client and obtain the full user profile. Use the received data to update the layout.
```
cocojoe: ~Use the user id~
Use the `userId`

articles/quickstart/native/android/04-user-profile.md
```diff
@@ -37,67 +37,57 @@ WebAuthProvider.init(auth0)
...
 
-Create instances of the API clients. You will use them to request the users' profile data.
+
+To get the user's information:
+1. Use the user's access token to call the `userInfo` method in the `AuthenticationAPIClient` client instance.
+The profile obtained this way is OIDC-conformant. Depending on the [scopes](/scopes/current) you requested when logging in, the profile contains different information. The result will never contain fields outside the OIDC specification.
+2. Get the user's full profile using the [Management API](/api/management/v2#!/Users). This step is explained next:
+
+
+Create an instance of the Users API client using the id token that you saved in the log in step. In the snippet bellow we obtain it from the extras that the LoginActivity class has passed when starting this activity. The API client is used to request the users' profile data.
```
cocojoe: ~users' profile data.~
user's profile data.

cocojoe (changes requested):

articles/quickstart/native/android/03-session-handling.md
lbalmaceda: Is an interface that the manager requires to store/retrieve stuff. Users can implement their own logic or use the one we provide. I'll prepend some context and accept your suggestion for the last sentence.

articles/quickstart/native/android/04-user-profile.md
lbalmaceda: I rather not use that as the claim is called "sub" and we referred to it in the above paragraph as "obtaining the ... user id"

cocojoe:
@lbalmaceda where do these changes come from? e.g. what is the reference for changing the QS?
https://github.com/auth0/docs/pull/5281/commits/4b7eb6d3d912f0e6eea3509e48160abf2b7938f9#diff-8f93acdfacbf630ba49b65118cf5b42fR37

lbalmaceda:
@cocojoe I guess [this](https://github.com/auth0-samples/auth0-link-accounts-sample/blob/master/SPA/app.js#L18) can answer your question

cocojoe:
@lbalmaceda can you modify you PR info to state what it is going on here please. As it looks outdated, you mention `using /userinfo to verify` then in the commit history there is `remove /userinfo calls to "verify token"` Also should probably be clear it's the AT (access token) in that context.

lbalmaceda:
@cocojoe I've squelched some commits into a more descriptive one. The confusion here was because of the coming and going from one implementation to the previous one, along with some redaction improvements in the middle. Basically the PR adds the usage of the `CredentialsManager` to handle credentials, ensures a /userinfo call is made whenever profile data is required given an access token (since we don't verify id_tokens in our SDKs yet) and also changes the `UsersAPIClient` to be instantiated with an access token with the correct scopes instead of using the id token. I'll edit the description as well.

PR #5282, opened 2017-11-16
should be elsif not elseif

DavidBates:


mpaktiti:
Good catch, thank you @DavidBates !

PR #5290, opened 2017-11-20
Update index.md

aaguiarz:
Removing oauth/token endpoint documentation, as we want to discourage users from using it.


jeffreylees (approved):
Go ahead and push this to remove the method documentation... do we want to add the `login` method that's co/authenticate as well as removing this one? Or leave it for now?

PR #5301, opened 2017-11-21
fix url typo

newswim:
Noticed a broken link in the quick-start doc for `Single Page App > React`. The link labeled "Embedded Login sample" is broken, I believe this should fix it.


mpaktiti:
Yeap, that fixes it. Thank you @newswim ! 👍

PR #5314, opened 2017-11-23
Added description for config value GITLAB_URL

dctoon:
As of Gitlab version `2.3` the config value `GITLAB_URL` is added to support private instances.

This version is not yet available, will remove `do-not-merge` label when available.


dctoon:
Read to merge!

PR #5340, opened 2017-11-28
Fix token renewal by explicitly setting postMessageOrigin in renewAuth

jmangelo:
The Angular quickstart token renewal step is not working out of the box due to the `postMessageOrigin` option not being provided and the fact that the silent handler is served from a different origin than the application itself.

This PR adds the missing option; a similar PR was added to the associated quickstart sample, see:

https://github.com/auth0-samples/auth0-angular-samples/pull/70


luisrudge:
lgtm

PR #5342, opened 2017-11-28
update broken link to "custom-db" section

newswim:
One of the links in this articles appears to have broken. I tracked down the article that I believe it was intending to link to, and adjusted the path.


mpaktiti:
Thank you for fixing that @newswim ! 👍

PR #5372, opened 2017-11-30
Java code for Creating code challenge is not correct

dpancs:
The JAVA code that is used in the document is not actually JAVA but android code.  Base64.URL_SAFE | Base64.NO_WRAP | Base64.NO_PADDING is not standard Java code.


lbalmaceda (changes requested):
Good catch! Although I rather use the [apache commons](https://commons.apache.org/proper/commons-codec/apidocs/org/apache/commons/codec/binary/Base64.html#encodeBase64URLSafeString-byte:A-) method instead, as using the Java 8 one won't work for people still using JDK 7. I'd also add a comment line above the method call to let the audience know what dependency to include. What do you think?

deep-lyra:
@lbalmaceda I don't mind if you use the apache commons library... It was just kind of misleading to use the android reference for JAVA.  I spent quite some time stuck with the code as couldn't figure the dependencies.

lbalmaceda:
Closed due to inactivity in favor of https://github.com/auth0/docs/pull/5506

PR #5379, opened 2017-12-04
Update auth0.js github repo URL

mpaktiti:
v9 beta has been released so `master` now points there. I updated the v8 note to point to [auth0.js v8 branch](https://github.com/auth0/auth0.js/tree/v8)


mpaktiti:
cc @jeffreylees

PR #5413, opened 2017-12-08
Improved checkSession documentation

aaguiarz:
checkSession() returns tokens if the user has a valid session in the server, regardless if it was started from the hosted login page or not.


mpaktiti (approved):
LGTM

PR #5435, opened 2017-12-12
From Grean to Criipto

flensted:
Grean has changed it's name to Criipto. The docs are updated to reflect this change. Text and images changed.


mpaktiti:
Thank you @flensted ! This looks good. I will push a couple of commits to your branch in order to update the folders structure as well (URLs as auto-generated based on folder structure and we don't want to have `grean` as part of the URL). I'll also add some redirects so we avoid errors.

PR #5437, opened 2017-12-13
Prepare custom authorization > wrong audience info

PudiPudi:
Says to use Client ID, should be api identifier instead
This is correctly explained on the readme.md from the github linked previously on the same page


khorne3:
@PudiPudi Here's an [explanation of why it's client ID](https://github.com/auth0/auth0.js/issues/473) and not API identifier. We'll take a look into the sample. Thank you!

PR #5448, opened 2017-12-17
NodeJS WebApp Pug Template Fix

ajschmidt8:
As per the following closed pug issue, the `src` attribute should not contain variable substitution syntax.
https://github.com/pugjs/pug/issues/100


aaguiarz:
Thanks for the PR!

PR #5470, opened 2017-12-21
Move self-managed steps to separate doc and cleaned up screenshots

jhiner:


jhiner:
I added a note about token issuance to the `Features Supporting Use of Custom Domains` section. Katie I'm curious to get your feedback or if you think there's a better way. Thanks!

PR #5471, opened 2017-12-21
Fix typo in Access Token docs

martinbjeldbak:
It's (it is) to its.


mpaktiti:
Good catch, thank you @martinbjeldbak ! 👍

PR #5477, opened 2017-12-21
Fix typo in Getting Started > The Basics > Account and Tenants

jiesiren:


mpaktiti:
Good catch, thank you @jiesiren ! 👍

PR #5490, opened 2017-12-26
Fix broken link

nicogarcia:
### Briefing

Link with text "Non interactive clients" was pointing to the non-existent path `/client`, it was changed to `/clients#client-types`, client types was added to be more specific to the section where the "Non interactive clients" definition is placed 🙂 

Blame on me, I didn't read the whole contributing guide, hope it's not necessary for this minor fix, sorry 🙏


articles/integrations/azure-api-management/configure-auth0.md
```diff
@@ -81,4 +81,4 @@ At this point, you've set up Auth0 for use as an OAuth 2.0 authorization server.
 
 <%= include('./_stepnav', {
  prev: ["Integrate Azure API Management Service with Auth0", "/integrations/azure-api-management"], next: ["2. Configure Azure API Management", "/integrations/azure-api-management/configure-azure"]
-}) %>
```
nicogarcia, line 84: I don't know why this line appears changed, seems to be the same 🤔 , I edited the file in Github editor.

aaguiarz (approved):

PR #5506, opened 2017-12-27
Fix java code-challenge generation snippet

lbalmaceda:
Use Apache Commons Codec library instead of Android Base64 class.


cocojoe (approved):
LGTM

PR #5511, opened 2017-12-28
Organizing auth0-swift reference docs

jeffreylees:
Separating single doc into several subject-specific pages.

https://trello.com/c/km4DRYis

https://auth0-docs-content-pr-5511.herokuapp.com/docs/libraries/auth0-swift


mpaktiti (commented):
That's ok but you are depending solely on the sidebar. Maybe use the `next-steps` element at the end of the doc(s) so you may suggest what to read next?

PR #5516, opened 2017-12-29
## 2. Secure the API endpoints - Code error

twophoney:
Line 149 has a call to checkJwt which doesn't exist in this example.
The middleware for checking the JWT has been added twice; in the call to 'app.use' (line 131) and in app.post to /timesheets (line 149). 
Personally I prefer the call to the middleware on each endpoint rather than the whole api. However I've just removed the call to the middleware on line 149 as I assume this example is supposed to protect the whole api regardless of the endpoint. 
Cheers, Dave


mpaktiti:
That is correct, thank you @twophoney ! I merged another PR that fixes that, so I'm gonna close this one, but thanks again for pinging us and pushing a fix! 👍

PR #5517, opened 2017-12-29
Remove stray comma in json code sample

iantabolt:


mpaktiti:
Good catch @iantabolt , thank you for fixing that! 👍

PR #5521, opened 2017-12-30
Update rails quickstart login

miketheman:
Introduced in #5102 - update callback URL for consistency


miketheman:
Turns out, this wasn't correct - I had a different bug preventing the Hosted Login Page from loading.

PR #5523, opened 2017-12-31
Fix a small typo

swarupdonepudi:
Replace "if you secret is base64" with "if your secret is base64"


mpaktiti:
Good catch, thanks for fixing that @swarupdonepudi ! 👍

PR #5546, opened 2018-01-04
Renamed master/child to Relaying Party/OIDC Provider [DO NOT MERGE - requires updated screenshots]

nicosabena:
Renamed "master" to "Relying Party" and "child" to "OIDC Provider" to use the correct OIDC terminology.

Before merging, it needs updated screenshots with the same terminology change.


articles/connections/social/auth0-oidc.md
```diff
@@ -101,26 +107,17 @@ Once the user is authenticated, the resulting profile will contain the [Auth0 No
 
 Note that the generated `user_id` has the following format:
 
-`auth0-oidc|YOUR_AUTH0_CONNECTION_NAME|THE_CHILD_AUTH0_CONNECTION|THE_CHILD_USER_ID`
+`auth0-oidc|YOUR_AUTH0_CONNECTION_NAME|THE_OIDC_PROVIDER_AUTH0_CONNECTION|THE_OIDC_PROVIDER_USER_ID`
 
-The `access_token` is the JWT of the user in the child Auth0 connection. If you decode it, you will see all the properties that were requested in the `scope` of the auth0-oidc connection. For example, for `scope=openid identities` will return:
+The `access_token` is the JWT of the user in the OIDC Provider Auth0 connection. If you decode it, you will see all the properties that were requested in the `scope` of the auth0-oidc connection. For example, for `scope=openid email` will return:
```
saltukalakus, line 113: Access token (if it is JWT when audience is used) and the id token should have the `iss` of `Relaying Party` not the `OIDC Provider`. This is a bug inherited from the old version of the doc. May be we can also correct them.

articles/connections/social/auth0-oidc.md
```diff
@@ -2,23 +2,29 @@
...
-2. Take note of its **clientID** and **clientSecret**. You will need these to create the connection in the master tenant.
-3. Add the master tenant's login callback to the list of **Allowed Callback URLs**: `https://${account.namespace}/login/callback`
+1. Create a Client or edit an existing one. Set the client type to regular web app.
+2. Take note of its **clientID** and **clientSecret**. You will need these to create the connection in the Relying Party tenant.
+3. Add the Relying Party tenant's login callback to the list of **Allowed Callback URLs**: `https://${account.namespace}/login/callback`
 
 ![](/media/articles/connections/social/auth0-oidc/child-app.png)
 
-## Configure the Master Auth0 Tenant
+4. Ensure that the **OIDC-Conformant** toggle in the **OAuth** tab under the client's **Advance Settings** is turned **off**.
```
saltukalakus: I have tested while **OIDC-Conformant** toggle turned on for the client in OIDC Provider in a test tenant. I can still authenticate the user and get the tokens. May be this is already fixed now?

nicosabena: What id did the user got? I don't think it's fixed yet.

saltukalakus: Ah, yes it is undefined, just noticed it. `auth0-oidc|SALTUK-OIDC-CHILD-CON|undefined`. Please disregard this.

articles/connections/social/auth0-oidc.md
jeffreylees: Is there even a reason for this section to discuss decoding an access token? Why can't it just say that the requested claims will be included?

nicosabena: I'm not even sure why the server is putting an `id_token` in a field name `access_token`. Might be a better idea to remove the mention completely.

PR #5550, opened 2018-01-05
Update Flask QS fon new structure

alexisluque:


aaguiarz (changes requested):
It's not using /api/private, /api/public,etc

aaguiarz:
you need to update /docs/snippets/server-apis/python/use.md for the /private and /public ones

PR #5558, opened 2018-01-08
Update Ruby on Rails QS for new structure

alexisluque:


aaguiarz (changes requested):
Can you add the /public endpoint too?

PR #5569, opened 2018-01-10
Note the signupLink -> signUpLink casing change

dentarg:
The change in casing was significant, with the wrong casing, it didn't work.

(At least for Lock 11, didn't try with Lock 10 as I upgrade directly to 11.)


jeffreylees (approved):
This is correct, casing change for that option name occurred in Lock 10, and is the same in Lock 11.

Thanks @dentarg !

PR #5590, opened 2018-01-16
Update authorization-code-grant-pkce.md

kota:
"_" has to be escaped.


mpaktiti:
Good catch @kota, thank you for fixing that! 👍

PR #5598, opened 2018-01-17
update README.md

mikefitzbaxter-auth0:
Added text to indicate that only Auth0 team members can follow the link with install instructions for the docs local.


mikefitzbaxter-auth0:
Happy days. I assume this is how this works.

PR #5600, opened 2018-01-17
[DO NOT MERGE] Add 3.0 docs for Auth0 Delegated Admin Extension

mostekcm:
Need to get the 3.0 functionality updates done in preparation for the release of 3.0 of the Delegated Administration Extension.

### DIFF
--- a/articles/extensions/delegated-admin/hooks.md
+++ b/articles/extensions/delegated-admin/hooks.md
@@ -191,24 +191,37 @@ Supported action names:
  - `unblock:user`
  - `send:verification-email`
 
-#### Create Hook
+#### Write Hook
 
-Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Create Hook** allows you to configure.
+Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Write Hook** allows you to configure.
 
 Hook contract:
 
  - `ctx`: The context object.
+   - `request.originalUser`: The current user values, payload is the new set of fields.
    - `payload`: The payload object.
      - `memberships`: An array of memberships that were selected in the UI when creating the user.
      - `email`: The email address of the user.
      - `password`: The password of the user.
      - `connection`: The name of the user.
+   - `userFields`: The user fields array, if one is specified in the settings query.
+   - `method`: Either `create` or `update` depending on whether this is being called as a result of a create or an update call
  - `callback(error, user)`: The callback to which you can return an error and the user object that should be sent to the Management API.
 
 Example: **Kelly** manages the Finance department. When she creates users, these users should be assigned to her department.
 
 ```js
 function(ctx, callback) {
+  var newProfile = {
+    email: ctx.payload.email,
+    password: ctx.payload.password,
+    connection: ctx.payload.connection,
+    app_metadata: {
+      department: ctx.payload.memberships && ctx.payload.memberships[0],
+      ...ctx.payload.app_metadata
+    }
+  };
+
   if (!ctx.payload.memberships || ctx.payload.memberships.length === 0) {
     return callback(new Error('The user must be created within a department.'));
   }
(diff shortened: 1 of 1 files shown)


articles/extensions/delegated-admin/hooks.md
```diff
@@ -191,24 +191,37 @@ Supported action names:
  - `unblock:user`
  - `send:verification-email`
 
-#### Create Hook
+#### Write Hook
 
-Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Create Hook** allows you to configure.
+Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Write Hook** allows you to configure.
```
CriGoT, line 196: It's not clear that there hook is run also during update. Even if the contract includes `method` it would be better to clearly mention when this hook is executed

articles/extensions/delegated-admin/hooks.md
```diff
@@ -299,6 +314,145 @@ The **Settings Query** allows you to customize the look and feel of the extensio
...
+    },
+    ...
+]
+```
+
+- **property** (**required**): This is the property name of the ctx.payload object for the write hook.
+    - Example: `"property": "app_metadata.dbId"` will get set in `ctx.payload.app_metadata.dbId` in the write hook
+- **label**: This is the label that will be used when adding a label for the field on the user info page, create page, edit profile page, or search page
+- **sortProperty**: If sorting by a different field than this for the search table, use this field.  Dot notation is allowed.
+- **display**: true || false || stringified => This is the default display value.  If not overridden in search, edit, or create, it will use this value.
```
CriGoT, line 385: It would be good to define the default values. I'm assuming is `true` but I'm not a 100% sure from reading this

articles/extensions/delegated-admin/hooks.md
mostekcm: Fixed

articles/extensions/delegated-admin/hooks.md
mostekcm: Fixed

articles/extensions/delegated-admin/hooks.md
khorne3: I'm a bit confused by 
> assigned to the group/department/vendor/... of the current user

Does this mean that if I'm creating a user and I'm part of the IT group, then the new user is part of the IT group, too?

If so, then maybe something like this?

> The **Write Hook** allows you to automatically assign newly-created users to the appropriate group, department, or vendor. Generally, this is the same category to which you belong.

articles/extensions/delegated-admin/hooks.md
```diff
@@ -191,24 +191,37 @@ Supported action names:
...
-#### Create Hook
+#### Write Hook
 
-Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Create Hook** allows you to configure.
+Whenever new users are created you'll want these users to be assigned to the group/department/vendor/... of the current user. This is what the **Write Hook** allows you to configure.
 
 Hook contract:
 
  - `ctx`: The context object.
+   - `request.originalUser`: The current user values, payload is the new set of fields.
```
khorne3: > The current user values; the new set of fields is contained within the **payload** object

articles/extensions/delegated-admin/hooks.md
```diff
@@ -191,24 +191,37 @@ Supported action names:
...
    - `payload`: The payload object.
      - `memberships`: An array of memberships that were selected in the UI when creating the user.
      - `email`: The email address of the user.
      - `password`: The password of the user.
      - `connection`: The name of the user.
+   - `userFields`: The user fields array, if one is specified in the settings query.
```
khorne3, line 209: Provide link to settings query (#the-settings-query-hook) since it isn't defined or anything 'til later on in the page?

articles/extensions/delegated-admin/hooks.md
```diff
@@ -225,15 +240,17 @@ function(ctx, callback) {
...
   }
 
+  if (ctx.method === 'update') {
+    /* If updating, only set the fields we need to send */
+    Object.keys(newProfile).forEach(function(key) {
+      if (newProfile[key] === ctx.request.originalUser[key]) delete newProfile[key];
+    });
+  }
+
+  // NOTE: if you are setting userFields with restricted values using the settings query, you should enforce their limits here using the ctx.userFields array
```
khorne3, line 250: > NOTE: If you are using the settings query to popular userFields with restricted values, enforce their limits here using the ctx.userFields array

Also, I'm not sure what restricted values are, so maybe a short 3-5 word definition would be helpful?

articles/extensions/delegated-admin/hooks.md
```diff
@@ -299,6 +316,147 @@ The **Settings Query** allows you to customize the look and feel of the extensio
...
+  });
+}
+```
+
+## Custom Fields
+
+Starting with version 3.0 of the Delegated Admin Extension you can now specify custom fields.  These custom fields can be stored anywhere in user_metadata or app_metadata when creating,
+updating, or displaying users.  You can use these fields to search for users and display them in the search results.  Also, it allows you to customize existing fields, such as email, username,
+name, and connection.
+
```
khorne3, line 344: Beginning with version 3.0 of the Delegated Admin Extension, you can use custom fields when creating, updating, or displaying information about users.

Auth0 stores the information in custom fields in **user_metadata** or **app_metadata**. These fields are searchable.

Custom fields also allow you to modify existing fields, such as email, username, name, or connection.

PR #5609, opened 2018-01-18
Remove SSO ZenDesk doc

khorne3:
Removing this doc per this discussion: https://auth0.slack.com/archives/C1CCJPKAM/p1516141582000367


mpaktiti (changes requested):
I searched the repo (looking for other docs linking to this one) and found 2 references in the redirects file. It won't cause a problem, but let's do some cleanup wherever we can:
- https://github.com/auth0/docs/blob/remove-zendesk-sso/config/redirects.js#L536
- https://github.com/auth0/docs/blob/remove-zendesk-sso/config/redirects.js#L1450

PR #5626, opened 2018-01-22
Add state parameter desc for authorize method

denu5:
I think its necessary to mention the state parameter in this method, otherwise its hard to find


jeffreylees (approved):

jeffreylees:
Thanks @denu5 !

PR #5638, opened 2018-01-25
Updating migrations page

jeffreylees:
https://trello.com/c/nwLJgAXN

Adding recent migrations from December customer email to migrations page.

https://auth0-docs-content-pr-5638.herokuapp.com/docs/migrations

### DIFF
--- a/articles/migrations/index.md
+++ b/articles/migrations/index.md
@@ -25,6 +25,19 @@ If you need help with the migration, create a ticket in our [Support Center](${e
 
 Current migrations are listed below, newest first. For migrations that have already been enabled see [Past Migrations](#past-migrations).
 
+### Summary of Endpoint Migrations
+
+This table is a summary of the below endpoint migrations for the first part of 2018. See the entries below the table for more detailed explanations.
+
+| Endpoint | Replacement | Deadline |
+| --- | --- | --- | --- |
+| `usernamepassword/login` | `/co/authenticate` | April 01, 2018 |
+| `/oauth/ro` | `/oauth/token` | July 01, 2018 |
+| `/ssoData` | `/authorize?prompt=none` | April 01, 2018 |
+| `/tokeninfo` | `/userinfo` | June 01, 2018 |
+| `/delegation` | TBD | June 01, 2018 |
+| `/oauth/access_token` | TBD | TBD |
+
 ### Introducing Lock v11, Auth0.js v9
 
 | Severity | Grace Period Start | Mandatory Opt-In|
--- a/articles/migrations/index.md
+++ b/articles/migrations/index.md
@@ -27,24 +27,24 @@ Current migrations are listed below, newest first. For migrations that have alre
 
 ### Summary of Endpoint Migrations
 
-This table is a summary of the below endpoint migrations for the first part of 2018. See the entries below the table for more detailed explanations.
+This table is a summary of the endpoint migrations for the first part of 2018. See the entries below the table for more detailed explanations.
 
-| Endpoint | Replacement | Deadline |
-| --- | --- | --- | --- |
-| usernamepassword/login | /co/authenticate | April 01, 2018 |
-| [/oauth/ro](/api/authentication#resource-owner) | [/oauth/token](/api/authentication#authorization-code) | July 01, 2018 |
-| /ssoData | /authorize?prompt=none | April 01, 2018 |
-| [/tokeninfo](/api/authentication/reference#get-token-info) | [userinfo](/api/authentication#get-user-info) | June 01, 2018 |
-| [/delegation](/api/authentication#delegation) | TBD | June 01, 2018 |
-| [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) | TBD | TBD |
+| Endpoint | Replacement | Mandatory Opt-In | More Info |
+| --- | --- | --- | --- | --- |
+| /usernamepassword/login | /co/authenticate | 2018-04-01 | [More Info](#introducing-lock-v11-and-auth0-js-v9) |
+| [/oauth/ro](/api/authentication#resource-owner) | [/oauth/token](/api/authentication#authorization-code) | 2018-07-01 | [More Info](#introducing-resource-owner-support-for-oauth-token-endpoint) |
+| /ssoData | /authorize?prompt=none | 2018-04-01 | [More Info](#introducing-lock-v11-and-auth0-js-v9) |
+| [/tokeninfo](/api/authentication/reference#get-token-info) | [userinfo](/api/authentication#get-user-info) | 2018-06-01 | [More Info](#introducing-lock-v11-and-auth0-js-v9) |
+| [/delegation](/api/authentication#delegation) | TBD | 2018-06-01 | [More Info](#introducing-api-authorization-with-third-party-vendor-apis) |
+| [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) | TBD | TBD | [More Info](#introducing-resource-owner-support-for-oauth-token-endpoint) |
 
-### Introducing Lock v11, Auth0.js v9
+### Introducing Lock v11 and Auth0.js v9
 
 | Severity | Grace Period Start | Mandatory Opt-In|
 | --- | --- | --- |
 | Medium | 2017-12-21 |  2018-04-01 |
 
-We’re continually improving the security of our service. As part of this, we are deprecating a set of APIs (`usernamepassword/login`, `/ssodata`, [tokeninfo](/api/authentication/reference#get-token-info), [/delegation](/api/authentication#delegation)) used by Lock.js v8, v9, and v10 and and auth0.js, v6, v7, and v8. You should update your applications by **April 1, 2018**.
+We’re continually improving the security of our service. As part of this, we are deprecating a set of APIs (/usernamepassword/login, /ssodata, [tokeninfo](/api/authentication/reference#get-token-info), [/delegation](/api/authentication#delegation)) used by Lock.js v8, v9, and v10 and and auth0.js, v6, v7, and v8. You should update your applications by **April 1, 2018**.
 
 We are offering two options for this update:
 
@@ -53,7 +53,7 @@ We are offering two options for this update:
 
 #### Am I affected by the change?
 
-If you are currently using Lock v9 or v10, or Auth0.js v7 or v8, you will be affected.
+If you are currently implementing embedded login in your application with Lock v9 or v10, or Auth0.js v7 or v8, you will be affected by these changes.
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
@@ -99,7 +99,7 @@ We are deprecating the usage of [id tokens](/tokens/id-token) when calling [/use
 
 #### Am I affected by the change?
 
-If you are currently using [[id tokens](/tokens/id-token) to access any part of the Management API, your application will need to be updated.
+If you are currently using [id tokens](/tokens/id-token) to access any part of the Management API, your application will need to be updated.
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
@@ -113,7 +113,7 @@ The [userinfo](/api/authentication#get-user-info) endpoint is being updated to r
 
 #### Am I affected by the change?
 
-If you are currently using the [userinfo](/api/authentication#get-user-info) endpoint to retrieve user information, you are affected by this change and need to update your implementation.
+If you are currently using the [userinfo](/api/authentication#get-user-info) endpoint or receiving id tokens, you are affected by this change and need to update your implementation so that it expects normalized OIDC conformant user profile attributes.
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
--- a/articles/migrations/index.md
+++ b/articles/migrations/index.md
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
 
 Current migrations are listed below, newest first. For migrations that have already been enabled see [Past Migrations](#past-migrations).
 
+### Introducing Lock v11, Auth0.js v9
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-04-01 |
+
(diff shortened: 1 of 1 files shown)


articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
 
 Current migrations are listed below, newest first. For migrations that have already been enabled see [Past Migrations](#past-migrations).
 
+### Introducing Lock v11, Auth0.js v9
```
mpaktiti: I would replace the comma with `and`, it's only two things here and commas are not often used in headers

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-04-01 |
+
+We’re continually improving the security of our service. As part of this, we are deprecating a set of APIs (`usernamepassword/login`, `/ssodata`, `/tokeninfo`, `/delegation`) used by Lock.js v8, v9, and v10 and and auth0.js, v6, v7, and v8. You should update your applications by **April 1, 2018**.
+
+We are offering two options for this update:
+
+1. We are moving towards a centralized login experience and are recommending this option moving forward. There are many benefits of using a centralized login experience, read more about how Centralized Login works [here](/hosted-pages/login). Moving to the centralized login experience is straightforward and will simplify your code base. Read [this guide](/guides/login/centralized-vs-embedded) to learn more about the benefits of centralized login vs embedded login.
```
mpaktiti: We shouldn't use `here` or `click here` for links (read more [here](https://www.w3.org/QA/Tips/noClickHere) 😆  ). It's not informative, it's bad for SEO and bad for accessibility.

mpaktiti: Also, I'm not a big fan of this kind of self-promotion: `Moving to the centralized login experience is straightforward and will simplify your code base.` We seem to have it all over the place for HLP, and it's generally bad practice to tell a user how simple something is all the time.

mpaktiti: `This guide` is a bad name for a link as well. Why not sth like this? `Learn more about the benefits of [centralized login vs embedded login](/guides/login/centralized-vs-embedded).`

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+1. If you decide you need to continue using embedded login, here are the migration guides for [Auth0.js](/libraries/auth0js/v9/migration-guide) and [Lock](/libraries/lock/v11/migration-guide).
```
mpaktiti: Now that I read this bullet I'm kind of wondering where is the migration guide for the first option. If someone uses HLP they will keep using Lock 10, right? Maybe say so in the first bullet? Because otherwise, the user will learn about this only if they read the note in this page https://auth0.com/docs/hosted-pages/login#2-choose-a-technology

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
+
+### Introducing Resource Owner Support for oauth/token Endpoint
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-07-01 |
+
+We’ve introduced support for Resource Owner Password to the `oauth/token` endpoint earlier this year ([doc](/api/authentication#resource-owner-password)). We will deprecate the current `oauth/ro` and `oauth/access_token` endpoints. Applications must be updated before July 1, 2018, when `oauth/ro` will become unavailable. A migration guide will be available in Q1.
```
mpaktiti: - change the `doc` link
- all endpoints should have links to their sections at the Authentication API, and not be formatted as code
- I don't like this whole paragraph, the sentences do not make sense, can you reword this please? I don't understand what kind of migration guide I should expect. About /oauth/token? Or /oauth/ro? Or /oauth/access _token? All of the above?
- `Applications must be updated before July 1, 2018, when oauth/ro will become unavailable.`: what about /oauth/access_token?
- `A migration guide will be available in Q1.` -> `The migration guide...`

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+#### Am I affected by the change?
+
+If you are currently implementing the `oauth/ro` endpoint your application will need to be updated.
```
mpaktiti: The header says /oauth/token and here it says /oauth/ro. I know that what we want to say is "Hey, if you are using /oauth/ro, we are deprecating this, and you should use /oauth/token instead" but this is not clear from this doc.

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
+
+### API Authorization with 3rd Party Vendor APIs. Deprecating Delegation Endpoint
```
mpaktiti: We can't have two sentences and a full stop as header

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-06-01 |
+
+We’re re-architecting the mechanism by which you get tokens for 3rd Party/Vendor APIs (such as AWS, Firebase, and others). It now works the same as any custom API, providing better consistency. This new architecture will be available in Q1 2018 and its availability will officially deprecate the Delegation Endpoint. All clients must be updated prior to June 1, 2018. Formal migration guide will be available in Q1.
```
mpaktiti: - `We're` does not look good, let's not use it. We are.
- `We’re re-architecting the mechanism by which you get tokens for 3rd Party/Vendor APIs` -> `We are changing how you get tokens for 3rd Party/Vendor APIs`
- `such as` -> `for example,`
- `It now works`? It's not shipped so it doesn't work now like that
- `This new architecture will be available in Q1 2018 and its availability will officially deprecate` -> `This will be available in Q1 of 2018, when the Delegation Endpoint will be officially deprecated`
- add link to delegation endpoint
- `Formal` migration guide? `The migration guide...`

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+#### Am I affected by the change?
+
+If you are currently using delegation to provide third party authorization, your application will need to be updated.
```
mpaktiti: link to delegation

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
+
+### Deprecating the Usage of ID Tokens on the Auth0 Management API
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-06-01 |
+
+We are deprecating the usage of id tokens when calling `PATCH/GET` `/user` and `/device-credentials`. We have moved to regular access tokens, and this is available now. Applications must be updated by June 1, 2018, when the ability to use id tokens will become unavailable. Migration guides will be available by February 2018.
```
mpaktiti: link to `id tokens` and endpoints

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
+
+### Improved OpenID Connect Interoperability in Auth0
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-07-01 |
+
+We’re updating the `userinfo` endpoint to return OIDC conformant user profile attributes. The most notable change is that `user_id` becomes `sub`. This will deprecate the legacy Auth0 user profile (in `userinfo` and in ID Tokens). Applications must be updated to use the new user profile before July 1, 2018. Migration guide will be available in late Q1 2018.
```
mpaktiti: - We're -> we are
- link to endpoints, remove code formatting
- link to OIDC conformant user profile attributes
- link to the legacy Auth0 user profile

articles/migrations/index.md
```diff
@@ -25,6 +25,85 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+#### Am I affected by the change?
+
+If you are currently using the `userinfo` endpoint to return user information, you will need to ensure that your applications are migrated where applicable.
```
mpaktiti: - link to endpoint
- `to return` -> `to retrieve` or `to get`
- `you will need to ensure that your applications are migrated where applicable.` -> `you are affected by this change and need to update your implementation`.

mpaktiti (changes requested):

goduncan (changes requested):
Could we add a summary table of all the endpoint migrations and times to the top of this? See:
https://auth0.slack.com/files/U0NGV5Z70/F8Y84R00L/image.png

articles/migrations/index.md
jeffreylees: I agree that we should back off of the "it's simple" phrasing that we have in quite a few places, especially now that we have COA and custom domains as an alternative. It's enough to say that we recommend it because it's more secure.

In this specific case though, the statement is more of a statement of fact:

>will simplify your code base

By redirecting the user to HLP to do authentication, your code base is less complex as it doesn't include any auth stuff. It's not saying implementation is simpler, it's saying that less code is required in your application.

articles/migrations/index.md
```diff
@@ -25,6 +25,102 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-06-01 |
+
+We are deprecating the usage of [id tokens](/tokens/id-token) when calling [/users](/api/management/v2#!/Users/get_users_by_id) and [/device-credentials](/api/management/v2#!/Device_Credentials/get_device_credentials). We have moved to regular [id tokens](/tokens/access-token), and this is available now. Applications must be updated by June 1, 2018, when the ability to use id tokens will become unavailable. Migration guides will be available by February 2018.
+
+#### Am I affected by the change?
+
+If you are currently using [[id tokens](/tokens/id-token) to access any part of the Management API, your application will need to be updated.
```
mpaktiti: typo: `[[id tokens]`

articles/migrations/index.md
```diff
@@ -25,6 +25,102 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+| Severity | Grace Period Start | Mandatory Opt-In|
+| --- | --- | --- |
+| Medium | 2017-12-21 |  2018-07-01 |
+
+The [userinfo](/api/authentication#get-user-info) endpoint is being updated to return [OIDC conformant user profile attributes](/user-profile/normalized/oidc). The most notable change is that `user_id` becomes `sub`. This will deprecate the [legacy Auth0 user profile](/user-profile/normalized/auth0) (in [userinfo](/api/authentication#get-user-info) and in [id tokens](/tokens/id-token)). Applications must be updated to use the new user profile before July 1, 2018. Migration guide will be available in late Q1 2018.
+
+#### Am I affected by the change?
+
+If you are currently using the [userinfo](/api/authentication#get-user-info) endpoint to retrieve user information, you are affected by this change and need to update your implementation.
```
mpaktiti: What if I am not using userinfo but use id tokens instead and get the user profile from there? Don't I have to update my application? Shouldn't this be more generic and say something like your application should expect the normalized user profile structure instead?

articles/migrations/index.md
```diff
@@ -25,6 +25,102 @@ If you need help with the migration, create a ticket in our [Support Center](${e
 
 Current migrations are listed below, newest first. For migrations that have already been enabled see [Past Migrations](#past-migrations).
 
+### Summary of Endpoint Migrations
+
+This table is a summary of the below endpoint migrations for the first part of 2018. See the entries below the table for more detailed explanations.
+
```
mpaktiti: I found this matrix a bit confusing. To start with, I would use the same date format as the migrations do.

Either `Deadline` and `April 01, 2018`
or `Mandatory Opt-In` and `2018-04-01`

I also expected a `more info` link that would take me to the relevant migration notice in this doc.

articles/migrations/index.md
```diff
@@ -25,6 +25,102 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+| Endpoint | Replacement | Deadline |
+| --- | --- | --- | --- |
+| usernamepassword/login | /co/authenticate | April 01, 2018 |
```
mpaktiti: some endpoints are missing the `/` at the beginning

articles/migrations/index.md
```diff
@@ -25,6 +25,102 @@ If you need help with the migration, create a ticket in our [Support Center](${e
...
+
+We are offering two options for this update:
+
+1. We are moving towards a centralized login experience and are recommending this option moving forward. There are many benefits of using a centralized login experience, [read more about how Centralized Login works](/hosted-pages/login). Moving to the centralized login experience is straightforward and will simplify your code base. Read [this guide comparing centralized and embedded login](/guides/login/centralized-vs-embedded) to learn more about the differences between the two. If you choose to use centralized login, you will continue to use Lock 10 or Auth0.js v8 inside your Auth0 login page, as the new versions are designed for embedded scenarios.
+1. If you decide you need to continue using embedded login, here are the migration guides for [Auth0.js](/libraries/auth0js/v9/migration-guide) and [Lock](/libraries/lock/v11/migration-guide).
+
+#### Am I affected by the change?
+
+If you are currently using Lock v9 or v10, or Auth0.js v7 or v8, you will be affected.
```
mpaktiti: Shouldn't this be `If you are currently using Lock v9 or v10, or Auth0.js v7 or v8 with embedded login`? I'm not affected if I'm using them with HLP.

mpaktiti (changes requested):

goduncan (approved):

PR #5641, opened 2018-01-25
Add warning to SSO AD RMS doc

khorne3:
Add version warning to SSO AD RMS doc.

### DIFF
--- a/articles/sso/current/integrations/_template.md
+++ b/articles/sso/current/integrations/_template.md
@@ -1,5 +1,11 @@
 # ${service} Single Sign On Integration
 
+<% if (service === "Active Directory RMS") { %>
+::: warning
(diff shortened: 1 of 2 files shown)


mpaktiti (changes requested):
The warning appears before the H1, it should always be after https://auth0-docs-content-pr-5641.herokuapp.com/docs/sso/current/integrations/ad-rms

khorne3:
@mpaktiti This is one where the entire doc is a snippet:

> <%= include('./_template', { service: "Active Directory RMS", img: "ad-rms" }) %>

For these instances, should we omit the snippet and write everything out to ensure the header's in the right place?

erjohnson:
Could add this to the template partial:

```
<% if (service === "Active Directory RMS") { %>
::: warning
The steps in this tutorial are valid for Active Directory Rights Management Services 2008 and earlier.
:::
<% } %>
```

PR #5643, opened 2018-01-26
Adding Client > Advanced > Grant Types instructions

joshcanhelp:
This is a quick change to the WordPress Configuration doc based on some customer feedback on the latest release. Some older clients may not have the correct grant types, which will cause disruption to the login process.


cocojoe (approved):
LGTM

PR #5650, opened 2018-01-29
Update Falcor API for new structure

alexisluque:


rafanog (commented):
It's good, there are no typos and the language used is comprehensible

PR #5656, opened 2018-01-31
remove unsupported smtp port 465

sabues:


sabues:
ref: https://auth0.slack.com/archives/C1RC8M1MY/p1517368442000139

PR #5660, opened 2018-02-01
Removing mentions of Lock v11 not being supported in hosted pages

aaguiarz:


jeffreylees:
Also removes some neglected mentions of old versions of auth0.js and Lock

jeffreylees (approved):

PR #5687, opened 2018-02-06
Fix small typos on bulk import page

pmalouin:


mpaktiti:
Thank you for fixing that @pmalouin ! 👍

PR #5693, opened 2018-02-07
Logout default parameters value is added.

ehsankhfr:
**Logout** default parameter value for the **returnTo** is added.


jeffreylees:
Hi!

So, there is no default for the `returnTo` parameter - if you don't provide it, whatever you have in your Dashboard settings for your client's Allowed Callback URLs gets used, I believe. 

Thanks!

PR #5711, opened 2018-02-09
Update ID Tokens vs Mgmt APIv2 migration content

mpaktiti:


articles/migrations/index.md
```diff
@@ -88,17 +88,38 @@ If you are currently using [/delegation](/api/authentication#delegation) to prov
...
+- [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id)
+- [POST/api/v2/users/{id}/identities](/api/management/v2#!/Users/post_identities) (used for [Account Linking](/link-accounts), see warning panel below)
+- [DELETE /api/v2/users/{id}/identities/{provider}/{user_id}](/api/management/v2#!/Users/delete_provider_by_user_id)
+
+
+These endpoints will now accept regular [Access Tokens](/access-token). This functionality is available now. 
+
+To get a valid Access Token for these endpoints during authorization, you have to set the **audience** parameter to `https://${account.namespace}/api/v2/` and the **scope** parameter to the scopes required by each endpoint. 
+
+For example, the [GET /api/v2/users/{id} endpoint](/api/management/v2#!/Users/get_users_by_id) requires two scopes: `read:users` and `read:user_idp_tokens`. For detailed steps and code samples, see [How to get an Access Token](/tokens/access-token#how-to-get-an-access-token).
```
cabralmartin: We should list the new scopes that replaces the id_token, these are the `current_user:*` scopes.

articles/migrations/index.md
```diff
@@ -88,17 +88,38 @@ If you are currently using [/delegation](/api/authentication#delegation) to prov
...
+
+:::panel-warning Account Linking exception
+There are two ways of invoking the [POST/api/v2/users/{id}/identities](/api/management/v2#!/Users/post_identities) endpoint (follow the link for details). The use case where we send the **user_id** to link with, as part of the payload, **will not be available for Access Tokens issued to end users**. This use case requires a Dashboard Admin token. The other use case, where the second identity is established by attaching an ID Token, can be used with a valid Access Token, as described above.
```
cabralmartin: > The other use case, where the second identity is established by attaching an ID Token, can be used with a valid Access Token, as described above.

This one we will prolly not support. We should remove this for now.

cabralmartin (changes requested):

articles/migrations/index.md
mpaktiti: Do these scopes have be included in the request? For example, for this endpoint, I should also ask for the `read:current_user` scope? What changes from what we describe [here](https://auth0.com/docs/api/management/v2/changes#the-id_token-and-special-scopes)?

cabralmartin: Yes, you need to ask for the specific scope you need. We no longer will add them automatically as we did with the `id_token`.

articles/migrations/index.md
```diff
@@ -71,22 +73,55 @@ If you are currently using [/delegation](/api/authentication#delegation) to prov
...
+
+These endpoints will now accept regular [Access Tokens](/access-token). This functionality is available now.
+
+To get a valid Access Token for these endpoints during authorization, you have to set the **audience** parameter to `https://${account.namespace}/api/v2/` and the **scope** parameter to the scopes required by each endpoint.
+
+Note that the following scopes are also required by some endpoints:
+
+- `read:current_user` (required by: [GET /api/v2/users/{id}](/api/management/v2#!/Users/get_users_by_id), [GET /api/v2/users/{id}/enrollments](/api/management/v2#!/Users/get_enrollments))
+- `update:current_user_identities` (required by: [POST/api/v2/users/{id}/identities](/api/management/v2#!/Users/post_identities), [DELETE /api/v2/users/{id}/identities/{provider}/{user_id}](/api/management/v2#!/Users/delete_provider_by_user_id))
+- `update:current_user_metadata` (required by: [PATCH /api/v2/users/{id}](/api/management/v2#!/Users/patch_users_by_id))
```
mpaktiti: @cabralmartin Is this required to update *any* of the user info, or specifically the `user_metadata`?

articles/migrations/index.md
```diff
@@ -71,22 +73,55 @@ If you are currently using [/delegation](/api/authentication#delegation) to prov
...
+- `create:current_user_metadata` (required by: [PATCH /api/v2/users/{id}](/api/management/v2#!/Users/patch_users_by_id))
+- `delete:current_user_metadata` (required by: [DELETE /api/v2/users/{id}/multifactor/{provider}](/api/management/v2#!/Users/delete_multifactor_by_provider))
+- `create:current_user_device_credentials` (required by: [POST /api/v2/device-credentials](/api/management/v2#!/Device_Credentials/post_device_credentials))
+- `delete:current_user_device_credentials` (required by: [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id))
+
+For example, the [GET /api/v2/users/{id} endpoint](/api/management/v2#!/Users/get_users_by_id) requires the scopes: `read:users`, `read:user_idp_tokens`, and `read:current_user`. For detailed steps and code samples, see [How to get an Access Token](/tokens/access-token#how-to-get-an-access-token).
```
cabralmartin: Detail: `read:users` is an admin scope. `read:current_user` is an end user scope. For the intent of this migration (which is focused on end user flows) we should only list this one.

PR #5293, opened 2017-11-20
MFA API

khorne3:
Outstanding items:

- Replace legacy terminology in existing docs
- Documenting Common Scenarios
- Behavior resulting from deleting authenticators

Possible: FAQ to explain scenarios?

Questions:

- What should we say about associating an authenticator using self-association during the login process?
- Why two authenticators instead of one with push notifications?
- Why do we allow the association of multiple authenticators?

PR #5722, opened 2018-02-12
Add docs for MFA unified API

santiagoaguiar:
Adds documentation for the new MFA api on the Authorization API explorer.

## Depends on #5293


kusold (approved):
LGTM.

I do have a non-blocking question though. Are we allowed to use ES6 here? I didn't see anything specifying either way in the CONTRIBUTING.md.

If we are allowed, we may want to consider using [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) instead of `request` because it is likely to become a [web standard](https://fetch.spec.whatwg.org/) and is already shipped in most major web browsers.

mpaktiti:
@erjohnson @santiagoaguiar Is this ready for review or not? Also, should it be merged after https://github.com/auth0/docs/pull/5293 ? Thanks!

santiagoaguiar:
This is ready for review, and should be merged after #5293, yes.

santiagoaguiar:
Sorry for the title, was waiting for Eric's 👍 before changing it, but since he added the ready-for-review I think we are good to go :).

khorne3:
@santiagoaguiar Reverting due to build failure.

PR #5726, opened 2018-02-13
Update Enterprise Connection UI Screenshots and Text

khorne3:


cristiandouce:
Just one comment: The url's displayed on the screenshot browsers should either be `manage.auth0.com` or not be displayed at all. As they point to a heroku **internal only** deployment.

PR #5736, opened 2018-02-15
Include JsonWebToken class in Secured concern

ghost:
Tested on Rails 5.1.4, Ruby 2.4.2p198. 
When the `require_relative` is not included, you will get a `NameError (uninitialized constant Secured::JsonWebToken):` upon reaching any secured endpoint. Explicit `require_relative` fixes this.


ghost:
Turns out it isn't a problem with the AuthO implementation, just that `libs/` is not included by auto-loaded by default in some versions of Rails. 
To fix, add this in `config/application.rb`:
```
# config/application.rb

module MyModule
  class MyClass
    # ...
    config.autoload_paths << "#{Rails.root}/lib"
  end
end
```

alexisluque (approved):

PR #5738, opened 2018-02-15
Updating migrations page

jeffreylees:
https://auth0-docs-content-pr-5738.herokuapp.com/docs/migrations

### DIFF
--- a/articles/migrations/index.md
+++ b/articles/migrations/index.md
@@ -21,8 +21,6 @@ During the grace period, customers are informed via dashboard notifications and 
 
 If you need help with the migration, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
----
-
 ## Active Migrations
 
 Current migrations are listed below, newest first. 
@@ -45,13 +43,11 @@ If you are currently implementing login in your application with Lock v8, v9, or
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
----
+## Upcoming Endpoint Migrations
 
-## 2018 Upcoming Endpoint Migrations
+Based on customer feedback, we have adjusted our plans and will continue to maintain and support the below listed endpoints and features. 
 
-::: note
-Based on customer feedback, we have adjusted our plans and will continue to maintain and support the below listed proprietary endpoints. We will also publish guidance on the most effective ways to transition your applications to standards-based protocols. In the event that we need to make security enhancements to any of these legacy endpoints, we will promptly announce timeframes and guidelines for any required changes.
-:::
+We will publish guidance for each of the below scenarios on how to transition your applications to standards-based protocols. If we need to make security enhancements to any of these legacy endpoints which would require more urgency, we will promptly announce timeframes and guidelines for any required changes.
 
 ### Resource Owner Support for oauth/token Endpoint
 
@@ -61,7 +57,7 @@ Based on customer feedback, we have adjusted our plans and will continue to main
 
 Support was introduced for [Resource Owner Password](/api/authentication#resource-owner-password) to the [/oauth/token](/api/authentication#authorization-code) endpoint earlier this year. 
 
-The current [/oauth/ro](/api/authentication#resource-owner) and [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) endpoints will be deprecated at some point in the future.
+The current [/oauth/ro](/api/authentication#resource-owner) and [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) endpoints will be deprecated in 2018.
 
 #### Am I affected by the change?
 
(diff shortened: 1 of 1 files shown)


articles/migrations/index.md
```diff
@@ -56,62 +45,68 @@ If you are currently implementing login in your application with Lock v8, v9, or
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
-### Introducing Resource Owner Support for oauth/token Endpoint
+---
+
+## 2018 Upcoming Endpoint Migrations
+
+::: note
```
mpaktiti: I'd remove the panel here and leave the text as the introduction to this section (it's not really additional information)

articles/migrations/index.md
```diff
@@ -56,62 +45,68 @@ If you are currently implementing login in your application with Lock v8, v9, or
...
+Based on customer feedback, we have adjusted our plans and will continue to maintain and support the below listed proprietary endpoints. We will also publish guidance on the most effective ways to transition your applications to standards-based protocols. In the event that we need to make security enhancements to any of these legacy endpoints, we will promptly announce timeframes and guidelines for any required changes.
```
mpaktiti: Maybe simplify the language a bit? For example `In the event that we need` can be `If we need`.

articles/migrations/index.md
```diff
@@ -56,62 +45,68 @@ If you are currently implementing login in your application with Lock v8, v9, or
...
 | --- | --- | --- |
-| Medium | 2017-12-21 |  2018-07-01 |
+| Low | TBD | TBD |
 
 Support was introduced for [Resource Owner Password](/api/authentication#resource-owner-password) to the [/oauth/token](/api/authentication#authorization-code) endpoint earlier this year. 
 
-We will now deprecate the current [/oauth/ro](/api/authentication#resource-owner) and [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) endpoints.
-
-Applications must be updated before July 1, 2018, when [/oauth/ro](/api/authentication#resource-owner) will become unavailable. The migration guide will be available in Q1 2018.
+The current [/oauth/ro](/api/authentication#resource-owner) and [/oauth/access_token](/api/authentication#social-with-provider-s-access-token) endpoints will be deprecated at some point in the future.
```
mpaktiti: `at some point in the future` or within 2018?

articles/migrations/index.md
```diff
@@ -56,62 +45,68 @@ If you are currently implementing login in your application with Lock v8, v9, or
...
+::: note
+Based on customer feedback, we have adjusted our plans and will continue to maintain and support the below listed proprietary endpoints. We will also publish guidance on the most effective ways to transition your applications to standards-based protocols. In the event that we need to make security enhancements to any of these legacy endpoints, we will promptly announce timeframes and guidelines for any required changes.
+:::
+
+### Resource Owner Support for oauth/token Endpoint
 
 | Severity | Grace Period Start | Mandatory Opt-In|
 | --- | --- | --- |
-| Medium | 2017-12-21 |  2018-07-01 |
+| Low | TBD | TBD |
```
mpaktiti: Why did you change the severity to `Low`? Because it was pushed back or you were asked to do this specifically?

articles/migrations/index.md
```diff
@@ -56,62 +45,68 @@ If you are currently implementing login in your application with Lock v8, v9, or
 
 If you have any questions, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
-### Introducing Resource Owner Support for oauth/token Endpoint
+---
+
+## 2018 Upcoming Endpoint Migrations
```
mpaktiti: From the discussions I gathered that these migrations will start within 2018 (grace period start) and finish within 2018 (mandatory opt-in). Maybe the header should be `Upcoming Endpoint Migrations` and then clarify in the introductory text that the migrations will start within 2018 (as soon as we publish migration guides) and end (mandatory opt in) within 2019.

As is now I understand that these migrations will start and end within 2018.

articles/migrations/index.md
```diff
@@ -21,25 +21,14 @@ During the grace period, customers are informed via dashboard notifications and
 
 If you need help with the migration, create a ticket in our [Support Center](${env.DOMAIN_URL_SUPPORT}).
 
+---
```
mpaktiti: I'd remove the horizontal lines. This is not something we normally use, and if you think we should, we'd better follow up with design before start using them

mpaktiti (changes requested):

articles/migrations/index.md
jeffreylees: I can't imagine what the problem with using an `hr` is, or why it would require design's input... they're perfectly normal demarcation characters... but I will remove them.

PR #5743, opened 2018-02-15
Add acs migration

erjohnson:
Tutorial on moving from Azure Access Control to Auth0. Should this live under `/migrations` or stay where it is?

https://auth0-docs-content-pr-5743.herokuapp.com/docs/users/migrations/azure-access-control


mpaktiti:
@erjohnson Both 😄 You should add an entry for this at the migrations page (probably under Lock which we want to be more prominent), and then link to this new doc of yours for the migration guide.

mpaktiti (changes requested):
Looks good Eric 👍 Two things before I merge:
- add an update feed entry
- add this doc as `hidden: true` at the sidebar, under Migrations, so the breadcrumbs work for it

Thanks!

PR #5745, opened 2018-02-16
Fix invalid JSON

bhavikkumar:
Fix client credentials example to contain valid JSON.


mpaktiti:
Thanks for fixing that @bhavikkumar ! 👍

PR #5219, opened 2017-10-31
Update react-native docs

paulioceano:
Fix #4705

Issue #4705, opened 2017-07-20
No file path/name as a comment in most snippets for react ios tutorial


mikearias3:
**Environment Information**
Browser: Google Chrome v58.0.3029.96 (64-bit)
URL: https://auth0.com/docs/quickstart/native/react-native

**Description**
Most of the snippets presented in these tutorials don't have the name of the file and the path commented in the first line. The ones that do only have the name of the file in the comment

**Screenshot**
<img width="796" alt="screen shot 2017-07-19 at 10 10 30 pm" src="https://user-images.githubusercontent.com/5594487/28397398-49887ff2-6ccf-11e7-8843-b67863a28482.png">
![Uploading Screen Shot 2017-07-19 at 10.10.38 PM.png…]()

albertoperdomo:
Fixed via #5749. Closing.

PR #5749, opened 2018-02-17
Fix comments #5219

paulioceano:
Fix #5219


albertoperdomo:
Fixes #4705.
Addresses requested changes in #5219.

cocojoe (approved):
To be honest, I don't really like the inline comment with the file path. However, if this is the standard and it has been proven to help then I'm okay with it.

albertoperdomo:
@cocojoe Totally understand. Looking to move away from it soon. The file path will be present, but not as part of the code block.

PR #5753, opened 2018-02-20
an to a

mscharlock:
A very small change: "an URL" technically breaks grammar rules regarding when to use a vs an because it is a soft "yoo" sound. If it rhymed with curl, "an url" would work. Since this is 1-character change, I'm not going to add anything to update feed/heroku/etc.


mpaktiti:
You are right @mscharlock , thank you for fixing that! 👍

PR #5757, opened 2018-02-20
Add Client Logo text

khorne3:
https://auth0-docs-content-pr-5757.herokuapp.com/docs/clients/client-settings/native


mpaktiti:
@khorne3 I've already merged this but it would be good if we had a list of all these places that the logo will appear (especially those visible to our customers' customers :P ). Not sure if we have enough material, but we could maybe have a "branding" page. Wdyt? Maybe add a trello card if only for getting this list?

PR #5761, opened 2018-02-21
Change zoomable and remove backslashes (for html-to-react PR)

arielger:
- Change CONTRIBUTING.md to use zoomable attribute with html instead of markdown container
- Update images using old configuration
- Remove backslashes from imports

We need to merge the html-to-react PR in auth0-docs to merge this.


ramasilveyra:
the attr`zoomable` is invalid html, but `data-zoomable` is not! WDYT about changing to `data-zoomable`?

davidpatrick (approved):

PR #5773, opened 2018-02-23
fix AWS SAML mapping JSON

abbaspour:


mpaktiti:
Thanks for fixing that @abbaspour ! 👍

PR #5779, opened 2018-02-24
Fix verification/fallback page link

d60402:


mpaktiti:
Thanks for fixing that @d60402 👍 I will also update the link's text so it's the same with the paragraph's title.

PR #5782, opened 2018-02-26
Remove beta markers from custom domains

mpaktiti:


jhiner:
FYI There's also a beta label here: https://auth0.com/docs/custom-domains/self-managed-certificates

mpaktiti:
Thanks @jhiner I removed it

PR #5791, opened 2018-02-26
Rewrite 'The Basics' page for clarity and brevity.

kmshultz:
Hi there,

I'm new to Auth0 and just read through some of your introductory docs. This page (and a few others) were a little confusing to me, so I rewrote it with new users (like myself) in mind.

I didn't touch the ending sections ('Where to go from here' and 'Keep reading') even though these were probably the most confusing bits to me. I understand each individual point they make, but they're all over the place. As a new user, I have no idea which thing I should actually read next—there's just too much there. If you agree that they need to be cleaned up, I'll leave that to you in another PR. :)

If I've made any wrong assumptions in the rewrite, let me know!

Thanks,
Kent

### DIFF
--- a/articles/getting-started/the-basics.md
+++ b/articles/getting-started/the-basics.md
@@ -89,6 +89,8 @@ After ExampleCo signs up for an Auth0 account (and creates a tenant), it needs t
 
 ## Where to go from here
 
+If you wish to learn more about the next steps in setting up Auth0, read more:
+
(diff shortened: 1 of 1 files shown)


articles/getting-started/the-basics.md
```diff
@@ -5,93 +5,91 @@ toc: true
...
+## Connections
 
-::: note
-We won't get into details on how to create a client, since this article is meant to provide a high level overview. However, if you want to know more, refer to the [Clients](/clients) documentation.
-:::
+Just as clients help your applications connect to Auth0, **connections** help Auth0 connect to the identity providers that authenticate your applications' users. Auth0 supports a wide variety of identity providers and credential types:
 
-## Connection
+- [Database connections](/connections/database): Usernames and passwords stored in either your Auth0 account or your own (external) database.
+- [Social logins](/identityproviders#social): Google, Facebook, Twitter, and more.
```
jeffreylees, line 68: I haven't reviewed the rest of this, just spun up the Heroku build and saw that fail - my guess is that the issue is that these items need to be `/connections/social` and `/connections/enterprise` - traversing the document structure rather than the URLs they end up being.

khorne3:

PR #5792, opened 2018-02-26
Altering /clients/connections with a temporary multitenancy scenario

jeffreylees:
https://trello.com/c/UYWX9NJf

Replacing recommendation to create separate connections for separate clients in the same tenant with a very basic multi-tenancy scenario. This section will eventually be replaced with a link to a more in-depth multi-tenancy document with an example and diagram.

https://auth0-docs-content-pr-5792.herokuapp.com/docs/clients/connections

### DIFF
--- a/articles/clients/connections.md
+++ b/articles/clients/connections.md
@@ -6,17 +6,7 @@ crews: crew-2
 
 Connections are sources of users. They are categorized into Database, Social, Enterprise and Passwordless and can be shared among different applications.
 
-You can configure any number of connections, and then, at client level, choose which of them should be enabled for the given client.
-
-To view all available connections for a client navigate to [Dashboard > Clients > Connections](${manage_url}/#/clients/${account.clientId}/connections).
-
-![Client Connections List](/media/articles/applications/connections-dashboard-list.png)
-
-To enable or disable a connection toggle the switch. 
-
-In the screenshot above the user has many connections configured (a `Helpdesk` database connection, an Amazon social connection, a Bitbucket social connection, and more) but only two are enabled for this client: a database connection named `test-db` and the Facebook social connection. So when the users want to access this client, they have to either login using username and password or their Facebook connection.
-
-To view all the connections that you have configured or create new ones navigate to [Dashboard](${manage_url}/#/) and select the connection type you want:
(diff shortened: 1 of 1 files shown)


articles/clients/connections.md
```diff
@@ -6,37 +6,24 @@ crews: crew-2
...
-- For the `Fabrikam Employee Timesheets` client enable the Active Directory / LDAP and Google Apps connections.
-- For the `Fabrikam Customer Portal` client enable the ADFS, Microsoft Azure AD, Google Apps and LinkedIn connections.
-- That's it!
+
+- Create a tenant `Fabrikam-Internal` for your internal domain, and a client within it `Fabrikam Employee Timesheets` for timesheets.
+- Create a second tenant `Fabrikam-Public` for your public-facing domain, and a client within it `Fabrikam Customer Portal` for the customer portal.
+- Configure the following [Enterprise connections](${manage_url}/#/connections/enterprise) for the `Fabrikam-Internal` tenant: Active Directory / LDAP, and Google Apps; once each is set up, check the **Clients** tab to enable them.
+- Configure the following [Enterprise connection](${manage_url}/#/connections/enterprise) for the `Fabrikam-Public` tenant: Google Apps; once it is set up, check the **Clients** tab to enable it.
+- Configure the following [Social connections](${manage_url}/#/connections/social) for the `Fabrikam-Public` tenant: Facebook, LinkedIn; once each is set up, check the **Clients** tab to enable it.
```
mpaktiti, line 29: Here we say to use the Clients tab to enable/disable a connection but we removed the guidance that says how to do this earlier in the doc.

mpaktiti (changes requested):

articles/clients/connections.md
jeffreylees: Responded on Slack... TL;DR is that those were are separate places, and unfortunately, with the dashboard UI as it is currently, you have to toggle the clients on. The idea is that you should toggle them all on, I guess.

jeffreylees:
@mpaktiti let's go ahead and merge this unless you have anything else for me to change?

jeffreylees:
@mpaktiti ping

PR #5804, opened 2018-02-28
Rewriting and formatting

lakhanivai:

### DIFF
--- a/articles/cms/wordpress/how-does-it-work.md
+++ b/articles/cms/wordpress/how-does-it-work.md
@@ -32,7 +32,7 @@ Assumption : The user exists in your WordPress database and Auth0 **prior** to i
    * WordPress user **exists** whose credentials match an Auth0 `user_id`. (Does the plugin allow login ? Does it check email address too?)
    * WordPress user **does not exist** whose credentials match Auth0 `user_id` field. In this case it will check if there is a user with matching email address. If a email match is found, the Auth0-WordPress plugin identifies a user with the provided `email`. Next it checks to see if the email is verified by Auto0 user. 
      * If the user has **verified** their email, the Auth0-WordPress plugin will update the WordPress user's `user_id` and log the user in.
-     * If the user has **not** verified their email, the Auth0-WordPress plugin will end the authentication process, indicating that the user needs to verify their email prior to proceeding.
+     * If the user has **not verified** their email, the Auth0-WordPress plugin will end the authentication process, indicating that the user needs to verify their email prior to proceeding.
 
 ### Case 3: Login attempt by user who exists in Auth0 and a newly-created WordPress user.
 #### Solution: Login Successful
--- a/articles/cms/wordpress/how-does-it-work.md
+++ b/articles/cms/wordpress/how-does-it-work.md
@@ -11,12 +11,12 @@ description: This page explains the scenarios of how Auth0 integrates with WordP
 ###  Case 1: Login attempt by user who exists in Auth0 but not in WordPress 
 #### Solution : Auth0 creates a new user profile on WordPress and logs the user in.
 #### How it works - 
-1. The user tries to login to your WordPress site's using their login credentials.
+1. The user tries to login to your WordPress site using their login credentials.
 2. Auth0 attempts to authenticate the user.
 3. On successful login, the Auth0-WordPress plugin receives the user's Auth0 profile.
 3. Next the Auth0-WordPress plugin checks to see if there is a user in the WordPress database with credentials that match their Auth0 `user_id`. There are two possibilities here -
 
-   * WordPress user **exists** whose credentials match an Auth0 `user_id`. Next it checks if the `email` address of the user matches. (What happens if the email address does not match ? Does the user get notified that the user id is taken ?)
+   * WordPress user **exists** whose credentials match an Auth0 `user_id`. Next it checks if the `email` address of the user matches. >(What happens if the email address does not match ? Does the user get notified that the user id is taken ?)
    * WordPress user **does not exist** whose credentials match an Auth0 `user_id`, the Auth0-WordPress plugin creates a new user profile and logs the user in.
 
 
@@ -25,7 +25,7 @@ description: This page explains the scenarios of how Auth0 integrates with WordP
 #### How it works -
 Assumption : The user exists in your WordPress database and Auth0 **prior** to installing the Auth0-WordPress plugin.
 
-1. The user tries to login to your WordPress site's using their login credentials.
+1. The user tries to login to your WordPress site using their login credentials.
 2. Auth0 attempts to authenticate the user.
 3. On successful login, the Auth0-WordPress plugin receives the user's Auth0 profile.
 3. Next the Auth0-WordPress plugin checks to see if there is a user in the WordPress database with credentials that match their Auth0 `user_id`. There are two possiblities here -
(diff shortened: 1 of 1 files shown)


articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
-6. The Auth0-WordPress plugin finds the user in the WordPress database using their Auth0 user_id, so it logs the user in.
-
-## Scenario: Data Migration
-
-If you enable [data migration](/connections/database/migrating), the Auth0-WordPress plugin will expose two secure endpoints that allow Auth0 authenticate the users. These endpoints are secured with a secret token and only available to IP addresses associated with Auth0. You can change this in the Auth0 Dashboard's Client [Advanced Settings](${manage_url}/#/clients) page.
-
-The login flow is as follows:
-
-1. The user access your WordPress site's login page and provides their credentials.
+###  Case 1: Login attempt by user who exists in Auth0 but not in WordPress
```
joshcanhelp, line 11: "Login attempt by a user ..."

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
-
-## Scenario: Data Migration
-
-If you enable [data migration](/connections/database/migrating), the Auth0-WordPress plugin will expose two secure endpoints that allow Auth0 authenticate the users. These endpoints are secured with a secret token and only available to IP addresses associated with Auth0. You can change this in the Auth0 Dashboard's Client [Advanced Settings](${manage_url}/#/clients) page.
-
-The login flow is as follows:
-
-1. The user access your WordPress site's login page and provides their credentials.
+###  Case 1: Login attempt by user who exists in Auth0 but not in WordPress 
+#### Solution : Auth0 creates a new user profile on WordPress and logs the user in.
```
joshcanhelp, line 12: Auth0 doesn't create it, the plugin does based on a successful login/registration. I would also say "creates a new user in WordPress," profile might be confusing.

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
+#### How it works - 
+1. The user tries to login to your WordPress site using their login credentials.
```
joshcanhelp, line 14: This is a little confusing since they are actually logging into Auth0 once the plugin is active and configured.

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
+2. Auth0 attempts to authenticate the user.
+3. On successful login, the Auth0-WordPress plugin receives the user's Auth0 profile.
+3. Next the Auth0-WordPress plugin checks to see if there is a user in the WordPress database with credentials that match their Auth0 `user_id`. There are two possibilities here -
```
joshcanhelp, line 17: * Remove "Next"
* "Auth0-WordPress plugin" - formal name is "Login by Auth0," I think "the plugin" will suffice in this case
* "... to see if there is a user in the WordPress database with their Auth0 `user_id`"

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
 description: This page explains the scenarios of how Auth0 integrates with WordPress.
 ---
 
-# How Auth0 Integrates with WordPress
+# Auth0-WordPress Integration
```
joshcanhelp, line 5: "Auth0-WordPress" feels like a formal name, which, in this case, is incorrect. I would keep as-is.

articles/cms/wordpress/how-does-it-work.md
joshcanhelp: I am just now noticing that "Auth0-WordPress" is in the original text :)

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
+###  Case 1: Login attempt by user who exists in Auth0 but not in WordPress 
+#### Solution : Auth0 creates a new user profile on WordPress and logs the user in.
+#### How it works - 
+1. The user tries to login to your WordPress site using their login credentials.
+2. Auth0 attempts to authenticate the user.
+3. On successful login, the Auth0-WordPress plugin receives the user's Auth0 profile.
+3. Next the Auth0-WordPress plugin checks to see if there is a user in the WordPress database with credentials that match their Auth0 `user_id`. There are two possibilities here -
+
+   * WordPress user **exists** whose credentials match an Auth0 `user_id`. Next it checks if the `email` address of the user matches.  
+   >(What happens if the email address does not match ? Does the user get notified that the user id is taken ?)
```
joshcanhelp, line 20: _"What happens if the email address does not match? Does the user get notified that the user id is taken?"_ - If the email does not match, then WordPress is updated with the Auth0 email address (we assume that Auth0 is the source of truth for identity once the plugin is running). The second sentence could probably be removed since it sounds like a condition of logging in.

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
+   * WordPress user **does not exist** whose credentials match an Auth0 `user_id`, the Auth0-WordPress plugin creates a new user profile and logs the user in.
+
+
+###  Case 2: Login attempt by user who exists in Auth0 **and** WordPress.
+#### Solution : Login successful if either 'user_id' or 'email'(w verification) match.
```
joshcanhelp, line 25: Login is successful if either the 'user_id' or 'email' (with verification) matches what comes from Auth0.

articles/cms/wordpress/how-does-it-work.md
```diff
@@ -2,58 +2,56 @@
...
+2. Auth0 attempts to authenticate the user.
+3. On successful login, the Auth0-WordPress plugin receives the user's Auth0 profile.
+3. Next the Auth0-WordPress plugin checks to see if there is a user in the WordPress database with credentials that match their Auth0 `user_id`. There are two possiblities here -
+   * WordPress user **exists** whose credentials match an Auth0 `user_id`. 
+   * WordPress user **does not exist** whose credentials match Auth0 `user_id` field. In this case it will check if there is a user with matching email address. If a email match is found, the Auth0-WordPress plugin identifies a user with the provided `email`. Next it checks to see if the email is verified by Auto0 user. 
+     * If the user has **verified** their email, the Auth0-WordPress plugin will update the WordPress user's `user_id` and log the user in.
+     * If the user has **not verified** their email, the Auth0-WordPress plugin will end the authentication process, indicating that the user needs to verify their email prior to proceeding.
+
+### Case 3: Login attempt by user who exists in Auth0 and a newly-created WordPress user.
```
joshcanhelp, line 38: I don't know that this third case is really all that helpful, it sounds a bit like one of the conditions of the first scenario. I know that's on us :)

Seems like the 4 scenarios are:

1. Not in WP, not in Auth0
2. In WP, not in Auth0
3. Not in WP, in Auth0
4. In WP, in Auth0

... and then data migration is a bit of an outlier.

joshcanhelp:
@vatechnology - thank you for the feedback here. I left a few review comments and @mpaktiti will probably have another round once this is technically correct. 

I am, however, a little hesitant to update this at all. This is essentially a walk-through of all the logic that's already available in the plugin code:

https://github.com/auth0/wp-auth0

I'm wondering if there is much of an audience for someone who wants a technical walk-through but is not able to read the code. It seems like these scenarios could be found just by testing the current version. 

Interested to hear your perspective ... thanks again!

mpaktiti:
Hi @vatechnology ! Any news on this? Btw I'll add a `do-not-merge` label in this PR so it doesn't show up on my daily to-review list, when you address the comments Josh made please ping me so I can review. Thanks!

joshcanhelp (changes requested):

khorne3:

PR #5809, opened 2018-03-01
Adding section on error responses

jeffreylees:
https://trello.com/c/iDHjFcDx

Noting the differences between the old and new error responses in Lock/Auth0.js

https://auth0-docs-content-pr-5809.herokuapp.com/docs/libraries/lock/v11/migration-v10-v11#error-responses

https://auth0-docs-content-pr-5809.herokuapp.com/docs/libraries/auth0js/v9/migration-v8-v9#error-responses

@arcseldon I used the exact examples you gave me, hope those are all right?

### DIFF
--- a/articles/libraries/_includes/_error_responses.md
+++ b/articles/libraries/_includes/_error_responses.md
@@ -1,0 +1,35 @@
+### Error Responses
+
+The error response returned for the `/usernamepassword/login` endpoint is different from that retured from the `/co/authenticate` endpoint.
+
+The legacy error response is not OIDC Conformant. A standard DB connection had a JSON response payload similar to the following:
+
+```
+{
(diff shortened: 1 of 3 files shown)


articles/libraries/_includes/_error_responses.md
```diff
@@ -0,0 +1,35 @@
+### Error Responses
+
+The error response returned for the `/usernamepassword/login` endpoint is different from that retured from the `/co/authenticate` endpoint.
```
jeffreylees, line 3: I might alter this to just say "errors returned in previous versions were different than those which are returned in the newest version" rather than naming endpoints, since we've done an OK job of not needing to talk about low level items in the rest of the guides.

PR #5815, opened 2018-03-01
A couple of paragraphs were duplicated

davidelettieri:
I just removed a the duplicated paragraphs leaving what seems to be the best ones

https://auth0-docs-content-pr-5815.herokuapp.com/docs/quickstart/webapp/aspnet-core/v2#seed-project


jeffreylees (approved):
Thanks for noticing the duplication!

PR #5817, opened 2018-03-02
Add nonce to buildAuthorizeUrl() code example

nicogarcia:
### Description

In the `buildAuthorizeUrl(options)` documentation the code doesn't have the `nonce` option and is using `responseType: 'token id_token'`.
This option is required when [this condition](https://github.com/auth0/auth0.js/blob/master/src/authentication/index.js#L128) is satisfied.

I think a test can be added [here](https://github.com/auth0/auth0.js/blob/master/src/authentication/index.js) in auth0.js for that case and that would help document usages 🙂


mpaktiti:
@aaguiarz this looks good to go, I'm merging and you can follow up on the test recommendation by @nicogarcia

PR #5831, opened 2018-03-05
Add warning to user search doc

khorne3:
Add warning on what to do when running a search returning > 10k users.

https://auth0-docs-content-pr-5831.herokuapp.com/docs/api/management/v2/user-search


mpaktiti:
@khorne3 Looks to me like this https://github.com/auth0/docs/pull/5831/files#diff-5baea89ed2d91d5aef120803a49934bdR40 is supposed to do the same thing but the guidance is different. Maybe remove the contact support panel and leave the new guidance you added?

thameera:
In any case, it would be good to explicitly mention the number (10,000) I think.

PR #5834, opened 2018-03-05
Fix heading typo on extension

emsearcy:
Replaces "Exisiting" with "Existing" on the heading for the export/import extension.


mpaktiti:
@emsearcy Good catch, thank you for fixing that!

PR #5835, opened 2018-03-05
The Auth0 Management APIv2 Token

khorne3:
Edited doc for clarity and added instructions on how to set scopes for individual non interactive clients.

https://auth0-docs-content-pr-5835.herokuapp.com/docs/api/management/v2/tokens


articles/api/management/v2/tokens.md
```diff
@@ -1,78 +1,3 @@
...
-Once you have this information you are ready to call the API. Follow these steps:
-1. Go to the [Management API v2 explorer page](/api/management/v2)
-1. Click the __Set API Token__ button at the top left
-1. Set the __Domain__ and __API Token__ fields, and click __Set Token__
-1. Under the __Set API Token__ button at the top left, some new information is now displayed: the domain and token set, and the scopes that have been granted to this client
-1. Go to the endpoint you want to call, fill any parameters that might be required and click __Try__
-
-![Set the Token](/media/articles/api/tokens/set-token.png)
-
 ## Automate the Process
```
mpaktiti: I'm not sure how this happened, but I'm pretty sure you meant to delete this file altogether and not just some of its content 😄

articles/api/management/v2/tokens/index.md
```diff
@@ -0,0 +1,72 @@
+---
+description: Details on how to generate and use a token for the Auth0 Management APIv2
+section: apis
+crews: crew-2
+toc: true
+---
+# The Auth0 Management APIv2 Token
+
+To call the [Management API](/api/management/v2) endpoints, you'll need what we refer to as the **Management API Token**.
```
mpaktiti: Actually it's a standard Access Token. Maybe we need to start referring to it as such, otherwise people (like me ✋) might initially think it's some kind of different token. Or even though we might keep the `Management API Token` (since the `Access Token for the Management API` sounds kinda horrible) we should make it clear that this is just another Access Token in JWT format.

It's actually just a JWT with the `audience` set to `https://TENANT.auth0.com/api/v2/˜` and the scopes needed for each endpoint.

We could also change the header to sth like `Access Tokens for the Management API`. Wdyt?

articles/api/management/v2/tokens/index.md
```diff
@@ -0,0 +1,72 @@
...
+
+The Management API Token is a [JSON Web Token (JWT)](/jwt) that:
+
+1. Contains specific permissions (called **scopes**) on what the client is or isn't allowed to do with the API
+2. Is signed with the client's API key and secret
```
mpaktiti: I'm not sure if item (2) is correct. By "API key and secret" I understand Client ID/Client Secret. But the `Auth0 Management API` dashboard entry has the "Signing Algorithm" set to `RS256`. So it's signed with the Auth0's private key and validated by the consumer with the public key.

articles/api/management/v2/tokens/index.md
```diff
@@ -0,0 +1,72 @@
...
+
+## How to get an Management API Token
+
+There are two ways to get a Management API Token
+
+1. [Manually](/api/management/v2/tokens/manually) using the Dashboard
```
mpaktiti: I would like to see some more content here (otherwise I have to follow both links even though only one might apply in my case). Sth like:

1. You can get a token manually from the Dashboard. This is something you would do if you want to quickly test an endpoint. For more information on this see [Manually generate and use a token for the Management API](/api/management/v2/tokens/manually)

Likewise for the second. This one needs to change also a bit. Instead of saying `Programmatically, where you build a simple command line tool that generates tokens` we should say sth like `configure your application to get tokens`. The thing is you can get Management API tokens from **every client type**, not only CLIs

articles/api/management/v2/tokens/index.md
```diff
@@ -0,0 +1,72 @@
...
+1. Go to your [Client's Settings](${manage_url}/#/clients/${account.clientId}/settings) page
+2. Click the __Rotate__ icon <i class="notification-icon icon-budicon-171"></i> **or** use the [Rotate a Client Secret](/api/management/v2#!/Clients/post_rotate_secret) endpoint
+
+Please note that previously issued tokens will continue to be valid until their expiration time.
+
+**What are the Current User scopes in the ID Tokens?**
+
+Within the Users API, some endpoints have scopes related to the current user (such as read:current_user or update:current_user_identities). These are [special scopes](/api/v2/changes#the-id_token-and-special-scopes) in the ID Token that are granted automatically to the logged in user.
+
+## Keep reading
```
mpaktiti: This section should have the following links:
- https://auth0-docs-content-pr-5835.herokuapp.com/docs/api/management/v2/tokens/manually
- https://auth0-docs-content-pr-5835.herokuapp.com/docs/api/management/v2/tokens/programatically
- [Changes in Auth0 Management APIv2 Tokens](/api/management/v2/tokens-flows)

I don't see anything else that's relevant, this is now only a landing page.

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
+---
+description: How to manually generate and use a token for the Management API
+section: apis
+crews: crew-2
+toc: true
+---
+# Manually generate and use a token for the Management API
```
mpaktiti: - Use title case for first-level headings: "Log In with a Social Identity"
- Use sentence case for subheads: "About the login process"

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
+---
+description: How to manually generate and use a token for the Management API
```
mpaktiti, line 2: add `title` as well

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+section: apis
+crews: crew-2
+toc: true
+---
+# Manually generate and use a token for the Management API
+
+In this article, we will cover how you can generate the token required to call the Management API manually. We will then cover how you can use your newly-obtained token.
+
+## Token validity
```
mpaktiti: This cannot be the first paragraph, it looks out of place. It should be later on, where you talk about the configuration of things.

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+
+By default, the Management API Token is valid for 24 hours. Once the token expires, you will need to obtain a new token.
+
+If this doesn't work for you, you can either:
+
+1. [Change the validity period of the token](#2-get-the-token)
+2. [Automate the token request process](/api/management/v2/tokens/programatically).
```
mpaktiti: I don't follow how the second bullet is related to token validity

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+#### How to Set Scopes
+
+To set the scopes for the Non Interactive Clients that access the Management API:
+
+1. Launch the [Auth0 Management API](${manage_url}/#/apis)
+2. Open the **Non Interactive Clients** tab
+3. Find the client you're interested in configuring, and click the drop-down arrow located to the right of its name
+4. Select the scopes desired
+
+### 2. Get the token
```
mpaktiti: wrong header level (should be H2)

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+
+The Management API Token is automatically generated and displayed on [the API Explorer tab of your Auth0 Management API](${manage_url}/#/apis/management/explorer).
+
+Please note that this token has, by default, an expiration time of **24 hours** (86400 seconds). To change this, you can update the __Token Expiration (Seconds)__ field before clicking __Update & Regenerate Token__.
+
+![Test Client](/media/articles/api/tokens/copy-token.png)
+
+Finally, click __Copy Token__ to get the updated token so that you can make authorized calls to the [Management API](/api/management/v2).
+
+### 3. Use the token
```
mpaktiti: wrong header level (should be H2)

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+## Token validity
+
+By default, the Management API Token is valid for 24 hours. Once the token expires, you will need to obtain a new token.
+
+If this doesn't work for you, you can either:
+
+1. [Change the validity period of the token](#2-get-the-token)
+2. [Automate the token request process](/api/management/v2/tokens/programatically).
+
+## Generate a token for the Management API
```
mpaktiti: I don't think you need a paragraph for this, can be part of the article's intro

articles/api/management/v2/tokens/manually.md
```diff
@@ -0,0 +1,101 @@
...
+
+To generate a token for use with the Management API, you will need to:
+
+1. Create and authorize a client
+2. Get the token by copying it for later use
+
+## 1. Create and authorize a client
```
mpaktiti, line 16: For the manual process you don't have to do this I think. A client (named `API Explorer Client`) is automatically created once the `Auth0 Management API` is created at Dashboard > API (which is also automatic I think)

mpaktiti: You can just leave two sections in the doc: how to get a token, how to use a token

articles/api/management/v2/tokens/programatically.md
```diff
@@ -0,0 +1,153 @@
+---
+description: How to programatically generate and use a token for the Management API
+section: apis
+crews: crew-2
+toc: true
+---
+# Programatically generate and use a token for the Management API
```
mpaktiti: I'd remove the programatically. This is actually the "only" way to properly get a token for the API, the other option is just for test purposes.

articles/api/management/v2/tokens/programatically.md
```diff
@@ -0,0 +1,153 @@
...
+
+In this article, we will show you how you can programmatically generate tokens for use with the Management API.
+
+The [manual process](/api/management/v2/tokens/manually) for getting a token might work for you if you want to test an endpoint or invoke it sporadically. However, if you need to make scheduled/frequent calls, then you should build a simple CLI tool that provides you with a token automatically. Its behavior is functionally equivalent to having a non-expiring token.
```
mpaktiti: This should be adjusted. You can get tokens for the Management API from all kind of apps, not only CLIs.

mpaktiti (changes requested):
Plus add an update feed entry

articles/api/management/v2/tokens/index.md
khorne3: @mpaktiti I think what I'll do is change the title to `Access Token for the Management API` in titles (and maybe headers) and just call it the Access Token throughout the doc. Hopefully that won't be confusing but will also make it clear it's not a special Auth0-specific token or anything.

articles/api/management/v2/tokens/manually.md
khorne3: @mpaktiti That's what I thought too, but when I tested, that's not the case. With a brand new tenant, you get an Management API and a default app client, but not the NIC. Not sure when this changed (or if it was changed back?)

mpaktiti: @khorne3 Can you cross check with @cabralmartin ?

khorne3: Confirmed behavior w/ Crew 2

khorne3:
@mpaktiti Rebased

mpaktiti:
This still needs some changes, for example if you want to get programmatically a token the only way is not the /oauth/token endpoint, you can get one from /authorize. There are various use cases and I'm quite familiar with them since I worked a lot with Martin on this for the deprecation of ID Tokens on the Mgmt API.

I will take this over and work on it next week.

note-to-self: add info about custom domains 
> Auth0 issues tokens with the iss claim of whichever domain you used with the request. For example, if you used `https://northwind.auth0.com/authorize` to obtain an Access Token, the iss claim of the token you receive will be `https://northwind.auth0.com/`. 
> If you used your custom domain `https://login.northwind.com/authorize`, the iss claim value will be `https://login.northwind.com/`.
> If you get an Access Token for the Management API using an authorization flow with your custom domain, you must call the Management API using the custom domain (your token will be considered invalid otherwise and you'll get an `Unauthorized` error).

mpaktiti:
Too many conflicts, redoing changes here https://github.com/auth0/docs/pull/6485

PR #5836, opened 2018-03-05
Correct typos regarding RS256 benefits

bugged84:
- Split two benefits from one bullet to two bullets.
- Fixed HS256 typo referring to secret key as private key.


mpaktiti:
@bugged84 Thanks for fixing that! 👍

PR #5846, opened 2018-03-06
Updated notice for possible issues with third-party cookies

nicosabena:


jeffreylees:
I think this can probably be closed. This note in this partial has been removed in an open PR because we're going to basically be steering people towards either custom domains in the main migration guides or towards Universal Login if they can't do custom domains, without the in-between.

PR #5854, opened 2018-03-07
Update number of authorization levels from four to three

gtseres:
The described authorization levels are **three**, where we state that there are **four**. Change the number to **three**.


mpaktiti:
Thanks for fixing that @gtseres ! 👍

PR #5860, opened 2018-03-08
Fix for sign up slack message rule

neshanjo:
Also considering the following case:
User requests refresh token during sign up, hence user has registered and authenticated (context.stats.loginsCount is 1), message sent --> OK!
User uses refresh token to authenticate (context.stats.loginsCount still is 1 since refresh token usage does not count as login), message sent --> WRONG!

Checking for context.protocol === 'oauth2-refresh-token' fixes this issue.

### DIFF
--- a/articles/rules/current/index.md
+++ b/articles/rules/current/index.md
@@ -221,8 +221,11 @@ The following example is a Rule template for sending a Slack message when a new 
 
 ```js
 function(user, context, callback) {
-  // short-circuit if the user signed up already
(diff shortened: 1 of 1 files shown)


nicosabena:
LGTM

mpaktiti:
Thank you @neshanjo! I pushed also the same fix in our rules examples repo https://github.com/auth0/rules/pull/161

mpaktiti (approved):

PR #5863, opened 2018-03-08
GDPR content | user consent

mpaktiti:
- [x] remove custom navigation
- [x] add breadcrumbs
- [x] update `protect user data > step-up` content 
- [x] update `user consent` content
- [x] add new tutorial: Track Consent with Custom UI
- [x] add new tutorial: Track Consent with Lock


articles/compliance/gdpr/features-aiding-compliance/data-portability.md
```diff
@@ -26,9 +26,4 @@ You can export a user's full profile using our Management API. The response will
 - [Search for a user using their email address](/users/search#users-by-email)
```
cabralmartin: Thinking that for GDPR purposes, might be better to only list GET by ID and User Export. User Search shouldn't be used (and won't be available) on many scenarios.

PR #5879, opened 2018-03-12
Migration guide for the Mgmt API/ID Token deprecation

mpaktiti:
For some use cases you could use [ID Tokens](/tokens/id-token) as credentials in order to call the [Management API](/api/management/v2). This functionality is being deprecated.

This PR adds the guide for this migration.


articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+
+This article will help you migrate your solution from the old implementation to the new one. First, we will see what this migration is about and which use cases are affected. We will continue with reviewing how you can use [scopes](/scopes) to get tokens with different access rights, and then see all the ways you can use to get an Access Token. Finally, we will review the [Account Linking](/link-accounts) process and the changes introduced to it.
+
+## Overview
+
+Some of the [/users](/api/management/v2#!/Users/get_users_by_id) and [/device-credentials](/api/management/v2#!/Device_Credentials/get_device_credentials) endpoints, could be called using ID Tokens. You can now use Access Tokens to access them. The grace period for this migration has already started and will end on the `1st of June, 2018`, when you will no longer be able to use ID Tokens as credentials for the Management API.
```
cabralmartin: We need to generate the final timeline.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+| [POST /api/v2/device-credentials](/api/management/v2#!/Device_Credentials/post_device_credentials) | Create a public key for a device |
+| [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id) | Delete a device credential |
+| [POST/api/v2/users/{id}/identities](/api/management/v2#!/Users/post_identities) | [Link user accounts](/link-accounts) from various identity providers |
+| [DELETE /api/v2/users/{id}/identities/{provider}/{user_id}](/api/management/v2#!/Users/delete_provider_by_user_id) | [Unlink user accounts](/link-accounts#unlinking-accounts) |
+
+Note that the last two endpoints are used for [Account Linking](/link-accounts#the-management-api). The changes in this functionality are the following:
+- You can no longer use an ID Token at the `Authorization` header
+- If you already use Access Tokens at the `Authorization` header you don't need to change something in your implementation, but do note that now you can send at the request's body the secondary account's ID Token, instead of the `user_id` (or you can keep using the `user_id`)
```
cabralmartin: > (or you can keep using the `user_id`)

You can keep using `user_id` with scope `update:users`. Scope `update:current_user` (the equivalent of using `id_token` for auth) won't support using `user_id`.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+| [PATCH /api/v2/users/{id}](/api/management/v2#!/Users/patch_users_by_id) | `update:current_user_metadata` | `update:users` |
+| [PATCH /api/v2/users/{id}](/api/management/v2#!/Users/patch_users_by_id) | `create:current_user_metadata` | `update:users` |
+| [DELETE /api/v2/users/{id}/multifactor/{provider}](/api/management/v2#!/Users/delete_multifactor_by_provider) | `delete:current_user_metadata` | `update:users` |
+| [POST /api/v2/device-credentials](/api/management/v2#!/Device_Credentials/post_device_credentials) | `create:current_user_device_credentials` | `create:device_credentials` |
+| [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id) | `delete:current_user_device_credentials` | `delete:device_credentials` |
+
+For example, if I get an Access Token that contains the scope `read:users` I can retrieve the data of **any user** using the [GET /api/v2/users/{id} endpoint](/api/management/v2#!/Users/get_users_by_id). But if my token contains the scope `read:current_user` I can only retrieve the information of the **currently logged-in user** (the one that the token was issued for).
```
cabralmartin, line 50: This is the story, not a standalone example. Current customers are using id_tokens as auth to modify the details of a particular user. This entire migration is to move these customers to use access_tokens as auth with the `*:current_user` scopes.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+
+There are several variations on how you authenticate a user and get tokens, depending on the technology and the [OAuth 2.0 flow you use to authenticate](/api-auth/which-oauth-flow-to-use):
+- Using the [Authorization endpoint](/api/authentication#authorize-client). This is where you redirect your users to login or sign up. You get your tokens from this endpoint if you authenticate users from a [single-page app](/api/authentication#implicit-grant) (running on the browser).
+- Using the [Token endpoint](/api/authentication#get-token).You get your tokens from this endpoint if you authenticate users from a [web app](/api/authentication#authorization-code) (running on a server), a [mobile app](/api/authentication#authorization-code-pkce-), a [server process](/api/authentication#client-credentials), or a [highly trusted app](/api/authentication#resource-owner-password).
+- Using [cross-origin authentication](/cross-origin-authentication) (endpoint: `/co/authenticate`). This endpoint is used to authenticate users when the requests come from different domains.
```
cabralmartin: Already synced, to be replace with examples for Lock / auth0js latest.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+    </div>
+  </div>
+</div>
+
+The differences that we introduced in order to get an Access Token that can access the Management API are the following:
+- We set the `audience` to `https://${account.namespace}/api/v2/`
+- We asked for the additional scope `read:current_user`
+- We set the `response_type` to `id_token token` so Auth0 will sent us both an ID Token and an Access Token
+
+Once you have the Access Token you can use it to call the endpoint. This part remains the same, nothing else changes in the request except for the value you use as `Bearer` token. The response remains also the same.
```
cabralmartin: Would be nice to show a sample access_token, highlighting the audience value, the `sub` and the `scopes`.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+}
+```
+
+### Use cross-origin authentication
+
+(to be updated)
+
+## Changes in Account Linking
+
+We are introducing some changes in the Account Linking process, one of which is breaking.
```
cabralmartin: We are in the context of a migration guide. We can remove this line.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+
+
+You can no longer invoke the endpoint with the authenticated primary account's ID Token in the `Authorization` header. You have to use an Access Token which must include the `update:current_user_identities` scope. We have already seen how you can get an Access Token. In the following snippets you can see what changes in the invocation of the endpoint (that is, only the value you use in the `Authorization` header).
```
cabralmartin: Removing `id_tokens` as auth is what this migration is about. No need to highlight the case for Account Linking. 
The item here should be that we will only support using `user_id` in payload for AT with scope `update:users`. No other scopes will be allowed when that payload is present.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,254 @@
...
+      </pre>
+    </div>
+  </div>
+</div>
+
+In order to use this new implementation, and send the `SECONDARY_ACCOUNT_ID_TOKEN` as part of the payload, the following must apply:
+- The ID Token must be signed using `RS256` (you can set this value at *Dashboard > Clients > Client Settings > Advanced Settings > OAuth*)
+- The claim `aud` of the ID Token, must identify the client, and be the same value with the `azp` claim of the Access Token
+
+If you already use Access Tokens at the `Authorization` header (use case described at the second bullet of [Linking User Accounts](/link-accounts#the-management-api)) you don't need to change something in your implementation, but do note that now you can send at the request's body the secondary account's ID Token, instead of the `user_id` (or you can keep using the `user_id`).
```
cabralmartin: This gives the impression that I can choose between `user_id` and token. In reality, only token is supported with `current_user` scope.

articles/migrations/index.md
```diff
@@ -94,25 +94,11 @@ This was used by the [/users](/api/management/v2#!/Users/get_users_by_id) and [/
 | [POST/api/v2/users/{id}/identities](/api/management/v2#!/Users/post_identities) | [Link user accounts](/link-accounts) from various identity providers |
 | [DELETE /api/v2/users/{id}/identities/{provider}/{user_id}](/api/management/v2#!/Users/delete_provider_by_user_id) | [Unlink user accounts](/link-accounts#unlinking-accounts) |
 
-These endpoints will now accept regular [Access Tokens](/access-token). This functionality is available now.
+These endpoints will now accept regular [Access Tokens](/access-token). The functionality is available now.
 
-To get a valid Access Token for these endpoints during authorization, you have to set the **audience** parameter to `https://${account.namespace}/api/v2/` and the **scope** parameter to the scopes required by each endpoint, according to the following table.
+Applications must be updated by June 1, 2018, when the ability to use ID Tokens will be disabled.
```
cabralmartin: Need final timelines.

cabralmartin (changes requested):

articles/migrations/guides/calling-api-with-idtokens.md
mpaktiti: Do you want me to update to something generic (like Q3 2018) or will we have a specific date these coming days?

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,298 @@
...
+  </div>
+</div>
+
+## Changes in Account Linking
+
+The changes in this functionality are the following:
+
+- You can no longer use an ID Token at the `Authorization` header
+- If you use an Access Token at the `Authorization` header, with `update:current_user_metadata` as the granted permission, then you must send the ID Token of the secondary account in the request's body
+- If you use an Access Token at the `Authorization` header, with `update:users` as the granted permission, then you must send the `user_id` of the secondary account in the request's body. The following must apply:
```
cabralmartin: If you use `update:users` then you get to chose between `user_id` or `id_token` in the payload. `update:users` is an admin scope, and admins can use their prefer method.

articles/migrations/guides/calling-api-with-idtokens.md
```diff
@@ -0,0 +1,298 @@
...
+  "method": "GET",
+  "url": "https://${account.namespace}/api/v2/users/USER_ID",
+  "headers": [{
+    "name": "Authorization",
+    "value": "Bearer YOUR_MGMT_API_ACCESS_TOKEN"
+  }]
+}
+```
+
+### Lock or auth0.js (embedded)
```
cabralmartin, line 214: Is the example listed here valid also for Lock? If so let's put a p making that explicit. If not let's add one example for Lock.

cabralmartin: Goal is to have link that TSE (or anyone) can share with a customer that addresses the customer scenario.

cabralmartin (changes requested):

articles/migrations/guides/calling-api-with-idtokens.md
mpaktiti: This is valid also for Lock in a sense, but it is done through Auth0.js. It's one of the cases where we recommend using them together:

> For simple uses, Lock is all that is necessary. However, while using Lock, if more customization is required in an application than Lock allows, functionality from the Auth0.js SDK can be used alongside Lock to meet those needs. An example might be using Lock to handle signups and logins, while using auth0.js to manage users (read and update user metadata, link user accounts together, and similar tasks).
https://auth0.com/docs/libraries/lock/v11/auth0js

I'll make it more clear in the text.

mpaktiti:
account linking updates https://github.com/auth0/docs/pull/5930

PR #5881, opened 2018-03-13
/oauth/ro migration guide

jeffreylees:
https://auth0-docs-content-pr-5881.herokuapp.com/docs/migrations/guides/migration-oauthro-oauthtoken


jeffreylees:
@jhiner this is very similar in content to that adoption guide page that we discussed which is in the docs already (almost verbatim, in fact). It's thinned down, step by step rather than the code pickers, and has less information, just comparing what the requests and responses would be like.

Anything you see that's inaccurate or out of date, or that should be added?

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+  "scope": "openid email favorite_color offline_access",
+  "device": "my-device-name"
+}
+```
+
+### Changes to Requests
+
+* The endpoint to execute token exchanges is now /oauth/token
+* [Auth0's own grant type](/api-auth/tutorials/password-grant#realm-support) is used to authenticate users from a specific connection `realm`. The [standard OIDC password grant](/api-auth/tutorials/password-grant) is also supported, but it does not accept Auth0-specific parameters such as `realm`.
+* `favorite_color` is no longer a valid scope.
```
jhiner: I think what we need to say here instead is something like (please wordsmith): "we support the standard OIDC scopes, along with scopes you have defined in your custom API".

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+  "client_id": "123",
+  "username": "alice",
+  "password": "A3ddj3w",
+  "realm": "my-database-connection",
+  "scope": "openid email offline_access",
+  "audience": "https://api.example.com"
+}
+```
+
+Note that the grant type is specified here as `http://auth0.com/oauth/grant-type/password-realm`, rather than `http://auth0.com/oauth/grant-type/password-grant`. The parameters `client_id`, `username`, and `password` are unchanged. The `realm` is included because we are using Password Realm grant type, and replaces the `connection` parameter from previous calls. The `scope` parameter is mostly the same, but does not accept non-OIDC values. Finally, the `audience` parameter can be added, indicating the API audience the token will be intended for.
```
jhiner: rather than `password` (password is the standard spec grant_type).

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+  "token_type": "Bearer",
+  "refresh_token": "8xLOxBtZp8",
+  "expires_in": 3600,
+  "id_token": "eyJ..."
+}
+```
+
+### Changes to Responses
+
+* The returned Access Token is valid for calling the [/userinfo endpoint](/api/authentication#get-user-info) (provided that the API specified by the `audience` param uses RS256 as signing algorithm) and optionally the resource server if one was specified.
```
jhiner: It might be useful here instead of saying "resource server" which is technically correct, to somehow map this to the "API" concept the customer creates in Auth0 Dashboard (https://auth0.com/docs/apis#how-to-configure-an-api-in-auth0).

jhiner (changes requested):

jeffreylees:
Anything else here @jhiner or are we good to release this?

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Migrating From /oauth/ro to /oauth/token
+description: Learn how to migrate your API calls and responses from /oauth/ro to /oauth/token
+toc: true
+---
+# Migration Guide for Resource Owner Password Credentials Exchange
```
mpaktiti: Move the migration guide here https://github.com/auth0/docs/tree/master/articles/migrations/guides

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
+---
+title: Migrating From /oauth/ro to /oauth/token
```
mpaktiti: set the `title` to the same value as the H1: Migration Guide for Resource Owner Password Credentials Exchange

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+description: Learn how to migrate your API calls and responses from /oauth/ro to /oauth/token
+toc: true
+---
+# Migration Guide for Resource Owner Password Credentials Exchange
+
+The Resource Owner Password Credentials exchange is used by highly-trusted clients to provide active authentication. Unlike the authorization code and implicit grants, this authentication mechanism does not redirect users to Auth0. It authenticates users with a single request, exchanging their password credentials for a token. This document describes the differences of this flow between the legacy and OIDC-conformant authentication pipelines.
```
mpaktiti: You need to talk about the migration at the intro, this is general information on the grant

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+Note that this guide is for users who use the resource owner password credentials exchange, and call /oauth/ro directly, without the use of any Auth0 libraries or SDKs. The major Auth0 libraries such as [Lock](/libraries/lock) or [Auth0.js](/libraries/auth0js) have already been updated to stop using /oauth/ro. If you use the `lock-passwordless` library, you can now use [Passwordless Mode](/libraries/lock/v11#passwordless) in Lock v11 instead.
```
mpaktiti: This should be under an [Does this affect me?](https://github.com/auth0/docs/blob/master/articles/migrations/guides/calling-api-with-idtokens.md#does-this-affect-me) paragraph

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Alter Your Requests
+
```
mpaktiti: You should have the requests side by side and the notes underneath. Like here https://auth0.com/docs/migrations/guides/calling-api-with-idtokens#lock-or-auth0-js-embedded- (looks broken atm but I reported it to docs-dev)

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+description: Learn how to migrate your API calls and responses from /oauth/ro to /oauth/token
+toc: true
+---
+# Migration Guide for Resource Owner Password Credentials Exchange
+
+The Resource Owner Password Credentials exchange is used by highly-trusted clients to provide active authentication. Unlike the authorization code and implicit grants, this authentication mechanism does not redirect users to Auth0. It authenticates users with a single request, exchanging their password credentials for a token. This document describes the differences of this flow between the legacy and OIDC-conformant authentication pipelines.
+
+Note that this guide is for users who use the resource owner password credentials exchange, and call /oauth/ro directly, without the use of any Auth0 libraries or SDKs. The major Auth0 libraries such as [Lock](/libraries/lock) or [Auth0.js](/libraries/auth0js) have already been updated to stop using /oauth/ro. If you use the `lock-passwordless` library, you can now use [Passwordless Mode](/libraries/lock/v11#passwordless) in Lock v11 instead.
+
+## Alter Your Requests
```
mpaktiti: only the H1 uses title case https://github.com/auth0/docs/blob/master/STYLEGUIDE.md#formatting

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+Previously, requests to /oauth/ro looked similar to this:
```
mpaktiti: I don't think there is any link to either of the endpoints in this doc. Link to the Authentication API.

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+```json
```
mpaktiti: this is not a request, this is the payload

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+  "connection": "my-database-connection",
+  "scope": "openid email favorite_color offline_access",
+  "device": "my-device-name"
+}
+```
+
+### Changes to Requests
+
+* The endpoint to execute token exchanges is now /oauth/token
+* [Auth0's own grant type](/api-auth/tutorials/password-grant#realm-support) is used to authenticate users from a specific connection `realm`. Auth0 supports the [standard OIDC scopes](/scopes/current#openid-connect-scopes), along with the scopes which you have defined in your [custom API](/api-auth/apis).
```
mpaktiti: This sentence has too many unknown terms and cross-linking. Try and explain it with simple words? It's not logical to expect users to read 3 articles just to understand 1 sentence.

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+* `favorite_color` is no longer a valid scope.
```
mpaktiti: You never explained the request (what is this supposed to do? explain your sample scenario) so the `favorite_color` makes no sense.

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+* The `device` parameter is removed.
+* The `audience` parameter is optional.
+
+Here is an example of a request to /oauth/token:
+
+```json
```
mpaktiti: not a request, just the payload of the request. Add the `curl` if you call it a request.

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+  "client_id": "123",
+  "username": "alice",
+  "password": "A3ddj3w",
+  "realm": "my-database-connection",
+  "scope": "openid email offline_access",
+  "audience": "https://api.example.com"
+}
+```
+
+Note that the grant type is specified here as `password-realm`, rather than the standard `password`. The parameters `client_id`, `username`, and `password` are unchanged. The `realm` is included because we are using Password Realm grant type, and replaces the `connection` parameter from previous calls. The `scope` parameter is mostly the same, but does not accept non-OIDC values. Finally, the `audience` parameter can be added, indicating the API audience the token will be intended for.
```
mpaktiti: Why use bullets in the previous one while using text here? Align these two (bullets are optimal for listing differences). Use the code picker bullets option.

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+## Alter Your Response Handling
```
mpaktiti: use sentence case, not title case

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+{
+  "access_token": "SlAV32hkKG",
+  "token_type": "Bearer",
+  "refresh_token": "8xLOxBtZp8",
+  "expires_in": 3600,
+  "id_token": "eyJ..."
+}
+```
+
+### Changes to Responses
```
mpaktiti: I don't see the point of having two headers for this: `Alter your response handling" with "Changes to responses" are both about the same thing. I'd keep the second one. The same goes for "Changes in requests"

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+* The returned Access Token is valid for calling the [/userinfo endpoint](/api/authentication#get-user-info) (provided that the API specified by the `audience` param uses RS256 as signing algorithm) and optionally the [custom API](/api-auth/apis) if one was specified.
+* The ID Token will be forcibly signed using RS256 if requested by a [public client](/clients/client-types#public-clients).
+* A Refresh Token will be returned only if the `offline_access` scope was granted.
```
mpaktiti: AND the API has the `Allow offline access` set

articles/guides/migration-oauthro-oauthtoken.md
```diff
@@ -0,0 +1,82 @@
...
+
+Here is an example of the OIDC conformant response from `oauth/token`:
+
+```json
```
mpaktiti: shouldn't you have both responses side by side so the user can cross check? And right underneath the scripts list the differences? Use the code picker

mpaktiti (changes requested):
+ update feed entry
+ update migrations page to list the guide here https://auth0.com/docs/migrations#resource-owner-support-for-oauth-token-endpoint (even if nothing else changes in terms of dates, just link to the migration guide)

articles/guides/migration-oauthro-oauthtoken.md
jeffreylees: How come? We have quite a few guides relating to migrations under /guides, are we moving those too?

articles/guides/migration-oauthro-oauthtoken.md
jeffreylees: Yeah, the initial code snippets were in a code picker, but I kind of dislike that for a migration guide. It seems like something a person might overlook, the tabs, whereas if you lay them out vertically you sort of have to see it.

Do we have a standard where we use code pickers for this?

PR #5890, opened 2018-03-14
Ionic 3 Quick Start

kmaida:
I have added an Ionic 3 Quick Start.

This one does not replace Ionic 2 (I have retitled Ionic 2+ to just Ionic 2) because Ionic 3 handles storage differently than Ionic 2 (returns promises).

This Quick Start should work with `auth0-cordova` with the server updates by @darkyen. The sample repo is here: https://github.com/auth0-samples/auth0-ionic3-samples


aaguiarz (approved):

articles/quickstart/native/ionic3/01-login.md
```diff
@@ -0,0 +1,262 @@
...
+// src/services/auth.config.ts
+
+export const AUTH_CONFIG = {
+  // Needed for Auth0 (capitalization: ID):
+  clientID: '${account.clientId}',
+  // Needed for Auth0Cordova (capitalization: Id):
+  clientId: '${account.clientId}',
+  domain: '${account.namespace}',
+  callbackURL: location.href,
+  packageIdentifier: 'YOUR_PACKAGE_ID' // config.xml widget ID, e.g., com.auth0.ionic
```
darkyen: Can we add, this value should be the same as the uri you specified for callback url?

articles/quickstart/native/ionic3/01-login.md
```diff
@@ -0,0 +1,262 @@
...
+3. In your [Auth0 Dashboard](https://manage.auth0.com), go to your Ionic app client's **Settings**.
+4. Add the full allowed navigation addresses (including `http://` and ports) to the **Allowed Origins (CORS)** settings for your Auth0 client and click the **Save Changes** button.
+
+You should now be able to log into your app!
+
+## Troubleshooting
+
+### Cannot read property 'isAvailable' of undefined
+
+This means that you're attempting to test this in a browser. At this time, you'll need to run this either in an emulator or on a device.
```
darkyen: This is a won't fix, I'd reword this as, "This library was designed to be ran on a device and expects a cordova environment with the SafariViewController plugin available" . Furthermore, this error can also be faced if you use `ionic serve` or `cordova serve`.

articles/quickstart/native/ionic3/01-login.md
```diff
@@ -0,0 +1,262 @@
...
+
+### Completely blank page when launching the app
+
+This could either mean that you've built the seed project using a different version of Ionic, or that the device where you are testing it isn't entirely supported by Ionic 3 yet. Be sure to check the console for errors.
```
darkyen: This could also be faced on iOS if you are running the app in a simulator and something else is listening on port `:8080` on the same machine. A good way to check for this is to run the application from XCode instead and check for the logs.

darkyen (approved):
I left some comments for possible improvements in future to make the documentation more helpful for the user. 

The sample looks good to me.

articles/quickstart/native/ionic3/01-login.md
kmaida: 👍

articles/quickstart/native/ionic3/01-login.md
kmaida: Is this comment referring to the packageIdentifier / line 77? If so, which callback URL are you referring to? Since there are actually a couple... (e.g., the `location.href` above, the Client settings URL which is `[PACKAGE_ID]://[DOMAIN]/cordova/[PACKAGE_ID]/callback`)

I think just saying this introduces more confusion on what we're referring to. Can you clarify? Or did you mean for this to refer to line 76?

articles/quickstart/native/ionic3/01-login.md
kmaida: 👍

articles/quickstart/native/ionic3/01-login.md
darkyen: I meant it for line 77, that value should always be the one you used when installing custom-uri-scheme, you shouldn't need to specify line 76.

kmaida: 👍

kmaida:
Addressed all feedback / comments.

rafanog (approved):

enli:
@kmaida , have a suggestion on this. Can you please point how to authenticate the requests on server side (say Node.js) based on Ionic quickstart tutorial above. It is overwhelming to figure out considering the vast documentation that auth0 has on the site.

kmaida:
@enli That information would be in a backend Quick Start for Node. You can check one out here: https://auth0.com/docs/quickstart/backend/nodejs/01-authorization

PR #5896, opened 2018-03-15
Rule example shows incorrect parameter

d60402:
[Customer asked about this](https://auth0.zendesk.com/agent/tickets/37299) doc that seems to imply the first parameter of a rule can arbitrarily change between `user` and `client`. This is not the case. First parameter is always `user`. Client information is found on the second (client) parameter.


mpaktiti:
Thanks for updating that @d60402 ! 👍

PR #5915, opened 2018-03-22
Updating leftover Lock v10 links

jeffreylees:
A few leftover links back to v10 docs

https://auth0-docs-content-pr-5915.herokuapp.com/docs/libraries/lock/v11/api
https://auth0-docs-content-pr-5915.herokuapp.com/docs/libraries/lock/v11/i18n


mpaktiti:
@jeffreylees Only the old versions need the version in the URL. If you excluded this here then it would still go to v11

PR #5916, opened 2018-03-22
Fix typos

seouri:
Replace ``ths`` with ``the``.


mpaktiti:
Thanks for fixing that @seouri! 👍

PR #5919, opened 2018-03-23
Use snake case for all "use_mfa" examples

dabroek:
Removed the inconsistent use of `use_mfa` vs. `useMfa`.


mpaktiti:
Good catch, thanks @dabroek ! 👍

PR #5926, opened 2018-03-26
Add note for custom DB usage of metadata

khorne3:

### DIFF
--- a/articles/metadata/index.md
+++ b/articles/metadata/index.md
@@ -200,13 +200,13 @@ lock.getUserInfo(accessToken, function(error, profile) {
 });
 ```
 
-## Metadata and Custom Databases
-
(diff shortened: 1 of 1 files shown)


articles/metadata/index.md
```diff
@@ -204,6 +204,10 @@ lock.getUserInfo(accessToken, function(error, profile) {
 For details on how to use Lock to authenticate users and access their profile information, check out the [Lock documentation](/libraries/lock).
 :::
 
+## Metadata and Custom Databases
+
+If you are using a [custom database](/connections/database#using-your-own-user-store), use **metadata** instead of **app_metadata** when working with your user profile app_metadata.
+
```
mpaktiti: @khorne3 I don't understand this at all. What is `metadata` that I should use? The field that used to hold the `app_metadata` for Mgmt APIv1? Can't think of anything else.

PR #5940, opened 2018-03-29
User experience when running example

thetwosents:
https link generated with error message on first run


jeffreylees:
If you try to use an HTTPS URL without a certificate, you will be warned about that by the browser.

Also, unsure why this issue was opened as a pull request?

PR #5942, opened 2018-03-29
Specify correct sample django app module name

burcaw:
The handler specified in 'JWT_PAYLOAD_GET_USERNAME_HANDLER' points to the django app 'authorization' which is inconsistent with the rest of the documentation. The rest of this doc assumes this example project has a django app called 'auth0authorization'.


joshcanhelp:
Already complete

PR #5948, opened 2018-03-30
Add tokens error

khorne3:

### DIFF
--- a/articles/errors/libraries/auth0-js/invalid-token.md
+++ b/articles/errors/libraries/auth0-js/invalid-token.md
@@ -7,11 +7,12 @@ public: false
 
 ## Why this error occurred
 
-Tokens signed with the HS256 algorithm cannot be properly validated.
+Beginning with **auth0.js version 9** and **Lock version 11**:
 
-Beginning with **auth0.js version 9** and **Lock version 11**, API calls to **/userinfo** must be made with a valid Access Token.
+* HS256-signed ID Tokens are discarded
+* API calls to **/userinfo** are made with a valid Access Token
 
-If you do not have an valid Access Token, you will receive the following error:
+If you do not have a valid Access Token, you will receive the following error:
 
 ```
 accessToken parameter is not valid
@@ -19,7 +20,8 @@ accessToken parameter is not valid
 
 ## Ways to fix this error
 
-1. Change the value of your **responseType** parameter to **token id_token**, so that you receive an Access Token in the response
+1. Change the value of your **responseType** parameter to **token id_token**, so that you receive an Access Token in the response.
+
 1. Change your client's hash algorithm to RS256. You can do so using the Dashboard:
 
     1. Go to [Dashboard > Clients]({$manage_url}/#/clients)
@@ -27,3 +29,6 @@ accessToken parameter is not valid
     1. Scroll to the bottom of the **Settings** tab, and click **Show Advanced Settings**
     1. Open up the **OAuth** tab. Change the value of **JsonWebToken Signature Algorithm** to **RS256**
     1. Scroll to the bottom of the page and click **Save Changes**
+
+    If you proceed with this option and you are using Access Tokens to call your APIs, be sure to change your server code so that it validates tokens using the RS256 algorithm instead of HS256.
+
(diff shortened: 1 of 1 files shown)


articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,23 @@
+---
+public: false
+---
+# HS256-Signed Tokens
```
luisrudge: Parsing a HS256-Signed id_token without an access_token

articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,23 @@
...
+
+Error: calls made to the **/userInfo** return the following:
```
luisrudge: Tokens signed with the HS256 algorithm cannot be properly validated. Instead, starting from auth0.js@v9 and lock@v11, a call to `/userinfo` is made with a valid `access_token`. If you don't have an access_token, you'll see the error:

articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,23 @@
...
+
+```
+accessToken parameter is not valid
+```
+
+## Cause
```
luisrudge: Not sure this section is needed, tbh. We already described the issue in the description above.

articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,23 @@
...
+Calls made to the **/userInfo** endpoint using Access Tokens validated using the HS256 algorithm cannot be validated.
+
+## Affected Versions
+
+* auth0.js version 9
+* Lock version 11
+
+## Troubleshooting
+
+Change the value of your **responseType** parameter to **token id_token**. This will allow you to use an Access Token with the **/userInfo** endpoint.
```
luisrudge: - Change the value of your **responseType** parameter to **token id_token**, so you get an access_token in the response
- Change your client's hash algorithm to RS256. Go to https://manage.auth0.com/#/clients, select your client, click "Show advanced settings", click on OAuth and change the value of "JsonWebToken Signature Algorithm" to RS256

luisrudge (changes requested):

articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,29 @@
...
+---
+# Parsing an HS256-Signed ID Token Without an Access Token
+
+**Error Message**: accessToken parameter is not valid
+
+## Why this error occurred
+
+Tokens signed with the HS256 algorithm cannot be properly validated.
+
+Beginning with **auth0.js version 9** and **Lock version 11**, API calls to **/userinfo** must be made with a valid Access Token.
```
aaguiarz: This is not accurate, /userinfo always required an access token.

The issue here is that the user did not request an access_token. The fix implies requesting the token.

They don't need to change it RS256, but we can suggest them to do that. If they do, and they are using the access_token to call their own APIs, they will need to change the server code to validate the token with RS256 instead of HS256.

khorne3:
@aaguiarz Do you suggest leaving the RS256 changes as is or separate it out as a "You might also consider..." tip?

aaguiarz:
@khorne3 Made some small changes, please review

articles/errors/libraries/auth0-js/invalid-token.md
```diff
@@ -0,0 +1,33 @@
...
+Calling **/userinfo** requires an Access Token. If you don't ask for an Access Token when authenticating, you will receive the following error:
+
+```
+The ID Token cannot be validated because it was signed using the HS256 algorithm
+and public clients (such as a browser) can’t store secrets. Please read the associated
+doc for ways to fix this.
+```
+
+## Ways to fix this error
+
```
woloski: There are two ways to fix the error:

1) **(RECOMMENDED)** Change the client signature algorithm to RS256 instead of HS256.
2)  Change the value of your **responseType** parameter to **token id_token** (instead of the default), so that you receive an Access Token in the response.

To do (1) follow these instructions:
....
....

woloski:
what are we waiting for?

khorne3:
@woloski I was waiting for the go ahead. Will merge now.

PR #5975, opened 2018-04-05
Android update snippets

### LINKED ISSUE #5945: Difference between code and samples
In the [Android quickstart](https://auth0.com/docs/quickstart/native/android), There are some difference between the code and the samples:

- Session Handling:
In the "Log the user out" section:
image

- User Profile:
In the "Request user's info" section:
image

-Calling APIs:
At the "Get the user's access token" section:
image

At the "Attach token" section:
image

paulioceano:
Fix #5945


albertoperdomo:
After reviewing, the changes don't seem appropriate.
In the sample, variables are assigned because they are declared above. In the tutorial, if this omitted it can cause confusion so it's better to leave it as is, IMO.

Closing.

PR #5981, opened 2018-04-09
Update minor typo in "Using C#" docs

martinbjeldbak:
The `#` in `C#` was being parsed as a comment in the header.

screen shot 2018-04-09 at 10 48 47 am

### DIFF
--- a/articles/rules/current/csharp.md
+++ b/articles/rules/current/csharp.md
@@ -2,7 +2,7 @@
 description: Learn how to use C# in Rules with Edge.js
 toc: true
 ---
-# Using C# in Rules
+# Using C\# in Rules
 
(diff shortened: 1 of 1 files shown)


mpaktiti:
Thank you for fixing that @martinbjeldbak ! 👍

martinbjeldbak:
@mpaktiti thanks for taking a look at this! As pointed out in #5982, it seems like the markdown parser used on your docs site is different than GitHub, causing it to incorrectly render as "C\" after this PR. If you could point me in the direction of the markdown parser you are using, I could look at submitting another PR to once and for all fix this. Cheers.

[Link](https://auth0.com/docs/rules/current/csharp) for below screenshot
skaermbillede 2018-04-12 kl 22 41 50

PR #5982, opened 2018-04-09
Escape '#' via HTML char code

sopelt:
... this should avoid prematurely ending the headline as "Using C"

### DIFF
--- a/articles/rules/current/csharp.md
+++ b/articles/rules/current/csharp.md
@@ -2,7 +2,7 @@
 description: Learn how to use C# in Rules with Edge.js
 toc: true
 ---
-# Using C# in Rules
+# Using C&#35; in Rules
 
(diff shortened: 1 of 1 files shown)


mpaktiti:
Hello @sopelt and thanks for fixing that! Some minutes ago I merged another PR that did the same fix (https://github.com/auth0/docs/pull/5981) so I'm going to close this one. Thanks again!

sopelt:
Hey @mpaktiti .. it looks like the other PR did not work as expected as

https://auth0.com/docs/rules/current/csharp

currently shows 

> Using C\

I am not saying that I know that my version will work (I was unable to determine what markdown implementation is being used) but you could give it a try ;)

martinbjeldbak:
@sopelt I have a feeling your PR may solve it :) would be great to see another Heroku Review App for this PR so we can confirm

mpaktiti:
Hi @sopelt Thanks for checking that and following up! I tried several approaches but nothing works correctly (with `&#35;` I get `C&` in the header). The only thing that looks like it's working is using an inline span element, i.e. `# Write Rules in C<span>#</span>`.

I'll look into this some more and push a fix in a while. Thanks again!

PR #5900, opened 2018-03-19
Restructure user search section

erjohnson:

PR #6001, opened 2018-04-11
Restructure user search section

mpaktiti:
Follow up PR to https://github.com/auth0/docs/pull/5900 after feedback from @siacomuzzi


articles/users/index.md
```diff
@@ -106,7 +106,15 @@ description: Learn about working with users in Auth0
   <li>
     <i class="icon icon-budicon-715"></i><a href="/users/search">User Search</a>
     <p>
-        Learn the best practices to use when retrieving users with API endpoints.
+        Learn how to search for users using the Management API.
     </p>
+    <ul>
+      <li>
+        <i class="icon icon-budicon-695"></i><a href="/users/search/query-syntax">Learn how you can use the query string syntax to build custom queries</a>
```
siacomuzzi, line 113: is `/users/search/query-syntax` a valid url?

siacomuzzi:
make sure to update `config/redirects.js` if necessary

articles/users/index.md
mpaktiti: Yes it's valid, it will navigate to the latest version by default (`/users/search/v3/query-syntax`) and the user can use the version dropdown to go other versions (v2) if they want. 

Only the sidebar needs the specific version as part of the URL.

mpaktiti:
@erjohnson I made some changes after Iaco's feedback, reach out to Slack if you have questions.

PR #5543, opened 2018-01-04
Fix JSON snippets

mpaktiti:

PR #6019, opened 2018-04-13
Fix digital ocean snippet

mpaktiti:
related to https://github.com/auth0/docs/pull/5543 and https://github.com/auth0/auth0-docs/issues/1612


khorne3:

PR #6027, opened 2018-04-13
fix code example for user content on localhost 

### LINKED ISSUE #6024: Docs: code example for user content on localhost creates problems for most development servers
Hi,

In order to skip user consent, the [User Content](https://auth0.com/docs/api-auth/user-consent#skipping-consent-for-first-party-clients) guide recommend to add an entry in the host file such as the following:
```
127.0.0.1       myapp.dev
```

This is not a good example because for a few months, [Chrome and Firefox now force .dev domains to HTTPS via preloaded HSTS](https://ma.ttias.be/chrome-force-dev-domains-https-via-preloaded-hsts/) and most of development servers only handle HTTP traffic.

According to this [guide](https://iyware.com/dont-use-dev-for-development/), something like `myapp.example` should be fine.

mlcdf:
This PR fixes #6024.

I'll add details to the update feed when I get home.


mlcdf:
Done.

mpaktiti:
Thank you for fixing this @mlcdf ! 👍

PR #6031, opened 2018-04-13
Add link to metadata handling in custom DBs

khorne3:
https://auth0-docs-content-pr-6031.herokuapp.com/docs/connections/database/custom-db#metadata

### DIFF
--- a/articles/connections/database/custom-db.md
+++ b/articles/connections/database/custom-db.md
@@ -160,7 +160,7 @@ If you use [Lock](/libraries/lock), you can customize the error messages that wi
 
 ## Metadata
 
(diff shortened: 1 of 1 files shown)


mpaktiti (commented):
Two notes before I merge:

1. Is the first sentence syntactically correct or is there some weird copy-paste error? This doesn't sound right "you may return a user profile available to Auth0 apps"

2. The last sentence looks like a mystery :) but the answer is quite simple. I think it would be simpler if you said that " the app_metadata field should be referred to as metadata in the scripts" and link to that doc for more info. Wdyt?

PR #6039, opened 2018-04-16
add info about CORS call

luisrudge:


jeffreylees (approved):

danbovey:
It should be `Allowed Web Origins` not `Allowed Callback URLs`.

This confusion between the two lists was the main problem I was facing in auth0/auth0.js#742.

luisrudge:
@danbovey indeed. I misread your comment and added the wrong info. I'm fixing it in another PR

PR #6068, opened 2018-04-20
Fix broken SAML link

thameera:
The `Special Configuration Scenarios` in this page is not working: https://auth0.com/docs/protocols/saml


mpaktiti:
Thank you @thameera ! 👍

PR #6078, opened 2018-04-23
Added link to default page name per user feedback.

rkhoriander:
Per user feedback, added links to version control page to help users find default page names. Also, updated password reset page to specify that the user email address should already have been specified. (User feedback indicated confusion about where in the password reset process this page appeared.)

### DIFF
--- a/articles/custom-domains/index.md
+++ b/articles/custom-domains/index.md
@@ -9,6 +9,8 @@ Auth0 allows you to map the domain for your tenant to a custom domain of your ch
 
 For example, if your Auth0 domain is **northwind.auth0.com**, you can have your users to see, use, and remain on **login.northwind.com**.
 
+Currently, each tenant on the Auth0 public cloud supports **one** custom domain.
+
 Using custom domains with universal login is the most seamless and secure experience for developers and end users. For more information, please see our docs on [universal login](/hosted-pages/login).
 
 ## Prerequisites
--- a/articles/hosted-pages/custom-error-pages.md
+++ b/articles/hosted-pages/custom-error-pages.md
@@ -29,7 +29,7 @@ If you choose to display a custom error page, you have two options:
 -  Redirect the user to a custom error page.
 -  Configure Auth0 to render a custom error page on your behalf (please note that this feature is only available via the Management API).
 
-#### Redirecting Users to a Custom Error Page
+#### Redirect Users to a Custom Error Page
 
 To redirect users to a custom error page:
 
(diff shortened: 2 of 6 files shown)


rkhoriander:
https://auth0-docs-content-pr-6078.herokuapp.com/docs/hosted-pages

mpaktiti:
@rkhoriander I'm not sure I follow the logic behind the cross-linking. What the _How to Use Version Control to Manage Your Hosted Pages_ page says is the file names that I should use **if** I'm using version control. 
(btw, @jeffreylees is it required to have these specific names?) 

Maybe it's just me but I still don't understand what the URL is. Is it `https://www.my-tenant.auth0.com/password_reset`? Should the users redirect to this URL or use some API/library? (because I know that for login we don't want them to redirect to the login page URL, but use the /authorize endpoint instead).

cc @jeffreylees wdyt?

jeffreylees:
Side note @mpaktiti - Yes, I believe for VCS extensions the page names need to be specific.

I know that for the login page in specific, we are actively discouraging users from navigating to it via the URL - you should not put the URL here at all. Calling it that way results in them hitting the wrong endpoint. The password reset page also isn't one that should be called directly as far as I know, but you might need to ask someone else.

TL;DR  **Don't** add a direct URL to the hosted login page anywhere. I'd ask engineering before adding direct URLs to any of the other hosted pages (and if we did add them, I see no reason why we'd need to cross-reference to the VCS doc).

I'm unclear why they'd want the name of the page unless they're looking for a direct URL as @mpaktiti was seeming to say, which is why I was just talking about URLs ^

mpaktiti:
@rkhoriander The reason why the user asked for a URL is because they wanted to load the page and display it to their users. They thought the solution was to get the URL and just redirect, but it might not be it. 

Nevertheless, the feedback is valid (even if the solution is not a URL redirect) because we are missing is a "how to load the page" section.

For example, for password reset, we have sections on how to:
- [Enable the page](https://auth0.com/docs/hosted-pages/password-reset#enable-the-custom-password-reset-page)
- [Edit the page](https://auth0.com/docs/hosted-pages/password-reset#edit-the-custom-password-reset-page)
- (something missing here)
- [Revert our changes](https://auth0.com/docs/hosted-pages/password-reset#revert-your-changes)

What is missing is **how to actually call the page**. So what you have to do is follow up with engineering to find the correct way to load the page (maybe it is a URL redirect, maybe not). I would start with crew-brucke (post in their channel asking for this info, loop me in if you want).
I would also double check 

I'll close this PR and wait for the next one!

PR #6097, opened 2018-04-25
Fix custom error_page PATCH examples

stefan-endouble:
Missing "application/json" content-type header and example curl requests fail with `{"statusCode":400,"error":"Bad Request" ... Payload validation error: 'Additional properties not allowed: error_page`.

Also fixed body according to management api docs: https://auth0.com/docs/api/management/v2.


mpaktiti:
Thanks for fixing that @stefan-endouble ! I'll push a small syntax fix so both headers are properly rendered and merge.

PR #6103, opened 2018-04-26
Update index.md

jgateley:
Fixed very small typo: "later" replaced with "latter"


mpaktiti:
Thanks for fixing that @jgateley ! 👍

PR #6114, opened 2018-04-26
Update which-oauth-flow-to-use.md

jgateley:
Changed passive sentence to active, removing superfluous comma at the same time.


mpaktiti:
Looks better like that, thanks @jgateley ! 👍

PR #6121, opened 2018-04-30
Fixing typos at the rails authorization guide

artemgurzhii:


articles/quickstart/backend/rails/01-authorization.md
```diff
@@ -120,7 +120,7 @@ With the `Secured` Concern in place, you can now apply it to whichever endpoints
 class PublicController < ActionController::API
   # This route doesn't need authentication
   def public
-    render json: { message: 'Hello from a public endpoint! You don't need to be authenticated to see this.' }
+    render json: { message: "Hello from a public endpoint! You don't need to be authenticated to see this." }
```
aaguiarz, line 123: We are using single quotes everywhere else, why should we use double quotes here? Could you please change it back to single quotes?

Thanks for the other change!

artemgurzhii: It's highlighting incorrectly, because of single quote usage inside this string: `You don't need`

davidpatrick:
Stale PR

PR #6130, opened 2018-05-01
fix additionalSignUpFields casing

luisrudge:


mpaktiti:
I found one more doc using the wrong casing and fixed it

PR #6137, opened 2018-05-03
Updated default picture rule script to modern syntax

shawnmclean:
Updated documentation example script for setting a default picture if one does not exist.

To keep inline with the update to rules repo: https://github.com/auth0/rules/pull/150


mpaktiti:
Thank you @shawnmclean ! 👍

PR #6149, opened 2018-05-07
Removal and change of wording in DB Conn Steps

timosperisen:
In step 1 of Create DB Connection I replaced "Create Application" with "Create DB Connection".
In step 2 Create DB Connection I changed the wording to a similar style like in "Create Application step 2".


mpaktiti:
Thanks @timosperisen ! 👍

PR #6163, opened 2018-05-11
fix typo in filename to match curl command

mrcnc:


mpaktiti:
Good catch, thanks for fixing that @mrcnc ! 👍

PR #6167, opened 2018-05-15
Update store-tokens.md

Tarasovych:
Space added on row 63 between "options" and "for".


mpaktiti:
Thanks for fixing that @Tarasovych ! 👍

PR #6178, opened 2018-05-19
Use ActiveSupport to_query

philipbjorge:
This method already exists in Rails Apps -- https://apidock.com/rails/Hash/to_query


philipbjorge:
Opened for a year, no response.

PR #6182, opened 2018-05-21
Fix file name for non-interactive clients

khorne3:
Non-interactive clients are now machine-to-machine applications

https://auth0-docs-content-pr-6182.herokuapp.com/docs/applications/application-settings/machine-to-machine

### DIFF
--- a/articles/applications/application-settings/index.yml
+++ b/articles/applications/application-settings/index.yml
@@ -5,9 +5,9 @@ versioning:
     - native
     - single-page-app
     - regular-web-app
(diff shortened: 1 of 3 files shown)


mpaktiti (commented):
@khorne3 There are some more, shouldn't we change them as well? https://github.com/auth0/docs/search?q=%22non+interactive+client%22&unscoped_q=%22non+interactive+client%22

khorne3:
I forgot about this PR and created a new one.

PR #6193, opened 2018-05-23
Arch. Scenarios Folder Reorg

khorne3:


articles/architecture-scenarios/index.md
```diff
@@ -1,80 +1,65 @@
...
 title: Architecture Scenarios
+classes: topic-page
 description: Learn about the common architecture scenarios that you will use to solve the authorization and authentication needs of your application.
 ---
 # Architecture Scenarios
 
 This page describes the typical architecture scenarios we have identified when working with customers on implementing Auth0.
 
-Click on any scenario to get more information.
+![Sample Architecture Scenario](/media/articles/architecture-scenarios/mobile-api.png)
```
mpaktiti: I find this very confusing since it is out of context (when I first saw it I thought it was left like this by mistake). Consider removing it?

articles/architecture-scenarios/index.md
```diff
@@ -1,80 +1,65 @@
...
 
 ## Application Configurations
 
 These scenarios describe the different type of technology architectures your application may use, and how Auth0 can help for each of those.
```
mpaktiti: Can you add a couple more sentences here and explain the structure of these scenarios? (that we present an imaginary business case, set goals and requirements, explain the architectural solution, and offer a sample implementation)

articles/architecture-scenarios/under-construction/index.md
```diff
@@ -0,0 +1,31 @@
+---
```
mpaktiti: I believe this doc is not needed. I understand that you created it so the breadcrumb "Under Construction" leads here, but I don't think you even need this directory.

I'd take all these files https://github.com/auth0/docs/tree/breadcrumbs-fixes/articles/architecture-scenarios/under-construction
and move them under https://github.com/auth0/docs/tree/breadcrumbs-fixes/articles/architecture-scenarios

mpaktiti (changes requested):
One more note on the structure. All the separate "implementations" folders (for example, `mobile-api-implementations`) were built because the idea was that we would add more languages. This didn't happen and I don't see it happening any time soon.

So I'd suggest that you remove those folders and move the files under the corresponding scenario. For example, move these files https://github.com/auth0/docs/tree/breadcrumbs-fixes/articles/architecture-scenarios/mobile-api-implementations under https://github.com/auth0/docs/tree/breadcrumbs-fixes/articles/architecture-scenarios/mobile-api

PR #6214, opened 2018-05-29
A few more additions for WCAG updates

mostekcm:
Katie, this is a couple of additional changes.  I may have a few more tomorrow.


mpaktiti:
@khorne3 I noticed that this doc still has the "Go Back" link at the end. Just a hands up that we should remove this once the breadcrumbs are live.

PR #6227, opened 2018-06-01
Refer to JWT claims not parameters

mmaddex:
Small nitpick: a JWT has an `aud` claim, not an `audience` parameter.


mpaktiti:
Good catch, thanks @mmaddex 👍

PR #6231, opened 2018-06-04
Fixed url to Deprecation Error Reference page

robbpjr:
Updating the url to the Deprecation Error Reference doc.


mpaktiti:
Thanks for fixing that @robbpjr 👍

PR #6238, opened 2018-06-06
Add paid plan requirement for custom domains

neocolmartin:
The custom domains settings UI for free accounts informs you that you need to "upgrade your account to any paid plan", but no mention is made of this in the documentation.

I do not know if this change requires an update to the version number (it doesn't look like the page is currently versioned), and I presume it does not need an entry in the update feed?

Should my change use the word "subscription" rather than "plan"? I've copied the wording from the management console, but the contributing guidelines seem to suggest the word "subscription".

### DIFF
--- a/articles/custom-domains/index.md
+++ b/articles/custom-domains/index.md
@@ -13,7 +13,8 @@ Using custom domains with universal login is the most seamless and secure experi
 
 ## Prerequisites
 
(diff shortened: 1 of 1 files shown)


mpaktiti:
Thank you @neocolmartin, the doc definitely needed this information. I updated the text a bit so it matches the dashboard message and I'll merge in a bit. Thanks again! 🙌

PR #6258, opened 2018-06-11
Update capitalization for Universal Login

rkhoriander:


mpaktiti (changes requested):
@rkhoriander There are a lot of conflicts here 😬 This happened because all these files have been updated since the last time you retrieved locally the content of auth0/docs (that's why we have to always sync our local files with the auth0/docs repo before we make changes).

You have two options:
- use the Resolve Conflicts, review each one, and resolve it manually (while being very careful not to override any of the existing changes)
- close this PR, retrieve locally the latest changes of auth0/docs (`git pull`), and then redo the renaming

Slack me for more info.

PR #6293, opened 2018-06-15
Update link to to point to #authorize-application

mauricioschneider:
Update link to point to the right section of the auth explorer that changed after rename of Clients -> Applications


mpaktiti:
Thanks @mauricioschneider ! Pushing some more fixes of the same issue on other docs.

PR #6311, opened 2018-06-21
Fix a typo in the `/oauth/token` example response

oponder:
Just a small typo I noticed while reading through https://auth0.com/docs/tokens/refresh-token/current


mpaktiti:
Thank you @oponder ! 👍

PR #6318, opened 2018-06-21
Update to the default Angular quickstart

DanOnCall:
Based on feedback from a test user, updated quickstart to be in sync with the code of the downloadable sample app for Angular.


articles/quickstart/spa/angular2/_includes/_centralized_login.md
```diff
@@ -2,7 +2,34 @@
 
 ## Create an Authentication Service
 
-Create a service to manage and coordinate user authentication. You can give the service any name. In the examples below, the service is `AuthService` and the filename is `auth.service.ts`.
+As a first step, create an `auth` sub-directory within your `src` directory. If you created your project using the Angular CLI, a good location for the `auth` sub-directory is `src/app`. 
+
+Next, inside the `auth` sub-directory, create a file named `auth0-variables.ts` that will store the Auth0 application keys and callbackURL constants that are going to be used to communicate with Auth0 through an Angular service.
+
+```ts
```
aaguiarz: Do we need to tell them to create this file? We have it to make it simpler for us to override the configuration, but they could actually set the client_id/domain when initializing webAuth, right?

articles/quickstart/spa/angular2/_includes/_centralized_login.md
```diff
@@ -100,51 +173,50 @@ export class AuthService {
...
 ## Provide a Login Control
 
-Provide a template with controls for the user to log in and out.
+Provide a template with controls for the user to log in and out. For example: 
 
 ```html
 <!-- src/app/app.component.html -->
 
-<nav class="navbar navbar-default">
+<nav class="navbar navbar-dark bg-dark">
```
aaguiarz: Why all the styling here? Not sure if it adds a lot. Also, we don't have it most QS, I think only the Angular Login has, and because I accepted a PR that I should have not ;)

aaguiarz (changes requested):
Looks great! added a few comments

articles/quickstart/spa/angular2/_includes/_centralized_login.md
DanOnCall: Cool, I can remove the Bootstrap stuff :) It's not really needed for the quickstart as the user will have their own styling. Only the HTML markup is key for the example.

articles/quickstart/spa/angular2/_includes/_centralized_login.md
DanOnCall: I think it's a good idea to have it. It is a good practice to keep a separate file for these type of constants in Angular. If the services need to be refactored or changed, the constants remain in one easy to find place. This also makes testing easier since the variables can be mocked - for whatever reason needed.

articles/quickstart/spa/angular2/_includes/_centralized_login.md
DanOnCall: @aaguiarz Done :) Removed the styling with the last commit

albertoperdomo (dismissed):

articles/quickstart/spa/angular2/_includes/_centralized_login.md
```diff
@@ -27,11 +28,11 @@ import * as auth0 from 'auth0-js';
...
-    domain: '${account.namespace}',
+    clientID: <YOUR AUTH0 CLIENT ID>,
+    domain: <YOUR AUTH0 DOMAIN>,
     responseType: 'token id_token',
-    audience: 'https://${account.namespace}/userinfo',
-    redirectUri: 'http://localhost:3000/callback',
+    audience: `https://${<YOUR AUTH0 DOMAIN>}/userinfo`,
+    redirectUri: <YOUR AUTH0 ALLOWED CALLBACK URL>,
     scope: 'openid'
   });
```
aaguiarz: Please use '${account.clientId}' and '${account.namespace}' as it will be replaced with the proper values.

For redirectUri please use http://localhost:3000/callback' and add a comment saying 'If you specified another callback URL please use that one here'.

Remove the 'scope' parameter as it will default to 'openid profile email'.

Remove the 'audience' parameter as it's not needed. 

Some of these changes are not done in other QS but we'll start updating them in that way. Please also make the changes in the sample code.

aaguiarz (changes requested):
Added more comments

albertoperdomo:
I have removed file `articles/quickstart/spa/angular2/dashboard-default.md`, so there's now a conflict in your branch. You can go ahead and delete it to solve this.

articles/quickstart/spa/angular2/_includes/_centralized_login.md
DanOnCall: Done for the doc. I need to update the source code of the sample app.

albertoperdomo:
As discussed today, we are closing this PR and will be doing major changes in the near future via a new PR.

PR #6321, opened 2018-06-22
Add missing headers required for /co/authenticate

lbalmaceda:
Any of the two headers are required by the endpoint. If cloudfront doesn't whitelist them, the call fails.


gnarea (approved):
lgtm

PR #6331, opened 2018-06-26
Libraries index page changes

tamigoodall:
Clarifications and simplifications to the  Libraries index page based on customer feedback.

### DIFF
--- a/articles/libraries/index.md
+++ b/articles/libraries/index.md
@@ -2,44 +2,44 @@
 section: libraries
 classes: topic-page
 title: Auth0 Libraries
-description: Overview of the Auth0 Libraries and SDKs
+description: Auth0 Libraries and SDKs overview
 topics:
   - libraries
   - lock
   - auth0js
+contentType: 
+    - index
+    - concept
+    - reference
 ---
 
 <div class="topic-page-header">
 <div data-name="example" class="topic-page-badge"></div>
 <h1>Auth0 Libraries</h1>
 <p>
-  There are several widgets and SDKs available for developers to provide a frictionless, simple experience when using Auth0. Take a look below to find documentation on the tools that you need to get started!
+  There are several widgets and SDKs available to provide a frictionless simple experience when using Auth0. Take a look below to find documentation for the tools that you need to get started!
 </p>
 </div>
 
 <%= include('../_includes/_lock_auth0js_deprecations_notice') %>
 
-## How Should You Implement Auth0?
+## Options for adding Auth0 to your web applications
 
-When adding Auth0 to your web apps, the best solution is to use Auth0's [Universal Login](/hosted-pages/login). Using Universal Login is a simple process, and prevents the pitfalls of [cross-origin authentication](/cross-origin-authentication). The login page uses by default the Lock Widget to authenticate users, but there is also a template for Lock Passwordless and a template for a custom UI built with the Auth0.js SDK available. 
+You can use Auth0 [Universal Login](/hosted-pages/login) to add Auth0 to your web apps to prevent the pitfalls of [cross-origin authentication](/cross-origin-authentication). The login page uses the Lock widget by default to authenticate users. There are also templates for Lock Passwordless and for a custom UI built with the Auth0.js SDK available. You can customize the page in the [Hosted Pages Editor](${manage_url}/#/login_page).
 
-You can customize the page in the [Hosted Pages Editor](${manage_url}/#/login_page).
-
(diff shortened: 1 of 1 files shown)


articles/libraries/index.md
```diff
@@ -2,44 +2,44 @@
...
 <p>
-  There are several widgets and SDKs available for developers to provide a frictionless, simple experience when using Auth0. Take a look below to find documentation on the tools that you need to get started!
+  There are several widgets and SDKs available to provide a frictionless simple experience when using Auth0. Take a look below to find documentation for the tools that you need to get started!
 </p>
 </div>
 
 <%= include('../_includes/_lock_auth0js_deprecations_notice') %>
 
-## How Should You Implement Auth0?
+## Options for adding Auth0 to your web applications
```
mpaktiti, line 26: I think it's still a bit confusing. I believe we should just list the options: 
- Lock (mentioning the variations like mobile versions, passwordless, etc)
- some Auth0 SDK
- Authentication API

This is not a reference, not guidance on how to implement login (where we would discuss universal login vs embedded).

I also believe this H2 is redundant and should be removed.

cc @jeffreylees @tamigoodall

jeffreylees: Honestly, now that I'm revisiting this, in the light of the embedded-login-needs-cnames-which-means-paid-plan issue, I'd be more tempted to double down on the embedded / universal separation 😁 

Options for adding:
- Universal (blah blah blah can implement lock/auth0.js via default templates or you can customize blah blah blah)
- Embedded (via one of the following methods; note that embedded authentication will require custom domains to avoid the pitfalls of cross-origin authentication, and custom domains are a paid feature).
    - Lock
    - SDKs (of which there are many)
    - API

If they are on this page and actually don't know what libraries / implementation methods we offer, and we don't steer them correctly at this point, we're wasting their time IMO.

mpaktiti: The embedded vs universal login thing does not seem to me like it belongs here. This is a Reference doc, it should list all our libraries. How you choose a library is part of another discussion, a more architectural one. For me this is a common problem amongst our docs, we try to do everything within a single article and the content ends up confusing.

jeffreylees: We'll probably have to agree to disagree on that point and just make a call about what to do. I don't believe that all users do or will follow our formula or funnel for how to find information. 

---

Example use case: 

I'm a new Auth0 user. I'm an experienced web dev who just wants to drop Auth0 into my JS app and move on with life. I set up an account.

I then just Google "Auth0 SDKs" or "Auth0 Libraries", knowing I want to include some sort of library to get going and use Auth0. I haven't read the getting started guides or docs, because who needs that? I'm a developer after all, I'll just get my hands dirty. I see this Libraries page, I see a list of libraries, I grab the GitHub link for Lock.js and start working. 

Hours of work later, I have a "working" solution. I basically did this by looking at the code and a few glances at the README. Piece of cake. But then I realize hours, days, or weeks later that for X% of my customers it may fail silently because I used embedded login when I shouldn't have.

---

We can take the discussion offline, or you all can just make a decision, but I'm just saying that I don't think this is a confusing element to have here - I think it's necessary. Maybe we can re-word it, format it differently, use UI elements, or something else to make it less confusing, if it seems confusing, but I think not having it at all would be a disservice to the customer because of our product situation, where we're still offering the use of our libraries to achieve embedded login even though it won't work correctly for free customers.

mpaktiti: Let's discuss this in our call next week. My main concern is that we seem to be replicating the same guidance over and over again and this thing is not manageable (and it becomes tiresome at times). Maybe the solution is not to put this guidance in all the docs, but notify the user someway different if they try to do something wrong. 

It might also be interesting to look into the customer feedback stats for this page.

tamigoodall: Adding it to the list for discussion for our meeting this week. My inexperienced thoughts on the matter are that simple is better and if we define the purpose of this page correctly, we can write it to serve that purpose. If this truly is an "overview" of the libraries and how they should be used, then a detailed conceptual description should be there with as much guidance as possible. If, however, the purpose is to provide links to the libraries in one easy-to-use list, then the less introductory text the better.

tamigoodall: Going back to the original feedback message from the user, it says "Assumes A LOT." I can reach out to the user and see if we can get more information regarding what they think is missing.

mpaktiti: I agree with both points @tamigoodall. It's just that I think right now the doc is a mix that confuses me. And yes by all means do reach out to the customer to make sure we are not making wrong assumptions.

PR #6338, opened 2018-06-28
Update devkeys.md

mjax-code:
I believe there is a typo on line 28. Changed "They" --> "The"


mpaktiti:
Correct, thanks for fixing that!

PR #6341, opened 2018-06-29
Added languages to Lock Internationalization

rkhoriander:


rkhoriander:
https://auth0-docs-content-pr-6341.herokuapp.com/docs/libraries/lock/v11/i18n

PR #6344, opened 2018-06-29
Added text to point to correct page for users who are trying to reset their own password

rkhoriander:


rkhoriander:
https://auth0-docs-content-pr-6344.herokuapp.com/docs/hosted-pages/password-reset

PR #6350, opened 2018-07-02
Update iOS Quickstart User Sessions inline with current flow

cocojoe:
Using read/write scopes
Using Management API Audience
Performing Get/Patch using AT


cocojoe:
Note to merge https://github.com/auth0-samples/auth0-ios-swift-sample/pull/52

cocojoe:
@aaguiarz better?

PR #6352, opened 2018-07-02
Add/fix troubleshooting steps; add new filters; fix how it works page

joshcanhelp:
- Clarify the How It Works page
- Add new filters to the Extending page
- Review, update, and add to the Troubleshooting section

Need to add details to the [update feed](https://github.com/auth0/docs/tree/master/updates) when complete.


tamigoodall (approved):
Looks good. No changes.

rkhoriander:
https://auth0-docs-content-pr-6352.herokuapp.com/docs/cms/wordpress/extending

articles/cms/wordpress/extending.md
```diff
@@ -371,6 +371,33 @@ function auth0_theme_hook_auth0_auth_scope( $scopes ) {
...
 
+### auth0_nonce_cookie_name
+
+Use this filter to modify the cookie name used for nonce validation. See the `auth0_state_cookie_name` filter below for an example.
+
+### auth0_state_cookie_name
+
+Use this filter to modify the cookie name used for [state validation](https://auth0.com/docs/protocols/oauth2/oauth-state). This can add a prefix or suffix or replace the name entirely. Make sure to use valid characters in any modifications made:
+
+> A <cookie-name> can be any US-ASCII characters except control characters (CTLs), spaces, or tabs. It also must not contain a separator character like the following: ( ) < > @ , ; : \ " /  [ ] ? = { }.
```
mpaktiti: @joshcanhelp I need a new format for `<cookie-name>` since this right now does not render correctly. 

image

I'll use code formatting (`<cookie-name>`), if that's ok with you.

Second thing, we use quotes
> like this

only when something is a direct quote from another page. If that's a note of ours then we use the `note` UI element instead. Should I change it or is this indeed a quote?

mpaktiti (commented):
@joshcanhelp I need your input on the below. I pushed some more fixes related to formatting. Other than that this is ready to be merged as soon as we fix this last thing. Thanks!

articles/cms/wordpress/extending.md
joshcanhelp: NP on code formatting ... that is indeed a direct quote

mpaktiti: Thanks Josh! Fixing and merging

PR #6365, opened 2018-07-05
Added note about Refresh Tokens and Implicit Grant

tamigoodall:
Added note for using silent authentication instead of refresh tokens.


mpaktiti (commented):
@tamigoodall Can you give me some context for these changes? Most of the text in the note replicates what the intro says. Maybe you can update the intro text instead?

Also, I don't get the reference to Refresh Tokens (since they are not mentioned anywhere in this doc and I can't see why they should).

Thanks!

tamigoodall:
@mpaktiti Here's the link to the Trello card with the feedback from a customer to add information on using refresh tokens with this feature (and the fact that you can't): https://trello.com/c/K185Fovr . I thought it made more sense as a note. I'll go over the intro again and see if I can make it fit better.

PR #6381, opened 2018-07-11
Libraries index simplification and custom domains disclaimer

tamigoodall:


articles/libraries/index.md
```diff
@@ -53,20 +41,13 @@ The Lock widget for each platform has detailed reference documentation.
   </li>
 </ul>
 
-### Lock Support Table
-
-Below are the GitHub links and support status for the various Lock widgets.
+## Lock links and support status
```
jeffreylees: This should be an H3. For this, and the corresponding heading for SDKs, I think I'd say `Lock repositories and support status` or something rather than the ambiguous "links"?

articles/libraries/index.md
```diff
@@ -83,18 +64,14 @@ The most commonly used SDKs have reference documentation.
...
-### SDK Support Table
-
-Below are listed all of the SDKs that are available for Auth0, with GitHub links and their support status.
+### SDK links and support status
 
 <%= include('../_includes/_libraries_support_sdks') %>
 
-
-### Framework/Platform Integration SDK Support Table
+### Framework/Platform integration SDK support status
```
jeffreylees: This is confusing to read (not that you wrote it, but we might as well fix now).

I wonder if we could say `### Platform integration repositories and support status` or something. Or we can just throw them all in under SDKs, but they were separated for a reason.

jeffreylees (changes requested):

articles/libraries/index.md
tamigoodall: Oops. I think you're correct about the H3. And I'll change the title.

articles/libraries/index.md
tamigoodall: Ok - I like that change.

khorne3:
@tamigoodall Can you please check the snippets that've been included? The app's erring because the file path looks to be off. Thank you!

tamigoodall:
@khorne3 Does it say which file is causing the problem? I added the include snippet to three different files.

articles/libraries/index.md
```diff
@@ -89,18 +63,14 @@ The most commonly used SDKs have reference documentation.
   </li>
 </ul>
 
-### SDK Support Table
-
-Below are listed all of the SDKs that are available for Auth0, with GitHub links and their support status.
+### SDK links and support status
```
jeffreylees: Same wording change here?

jeffreylees:
When doing these includes you have to check the `../../` traversing back through the file hierarchy to where the include lives - `/libraries` should be `../includes` (just step up one level) while `/libraries/lock/v11` should be `../../../includes` (step up three levels) etc etc

khorne3:
@tamigoodall Two errors show up when I run locally:

* > Error compiling document articles/libraries/auth0js/v9/index.md: Couldn't resolve include ../_includes/_embedded_login_warning included from file /docs/articles/libraries/auth0js/v9/index.md

* > Error compiling document articles/libraries/lock/v11/index.md: Couldn't resolve include ../_includes/_embedded_login_warning included from file /docs/articles/libraries/lock/v11/index.md

tamigoodall:
OK - I think they're fixed now.

khorne3:
@tamigoodall There were still some issues w/ the _embedded_login_warning snippet, but I've fixed. Thanks!

PR #6384, opened 2018-07-12
Update part-2.md

gautamceg:
Text or typo correction 
from -  "When creating an API in the Auth0 Dashboard" 
to-  "When creating a new application in the Auth0 Dashboard"


khorne3:
@gautamceg The doc as it exists is correct. Creating an API in Auth0 comes w/ an accompanying application. Thanks!

PR #6415, opened 2018-07-20
Add explicit Arbiter connectivity req

sabues:
Aims to clarify connectivity requirements for the arbiter role for Geo-HA


khorne3 (approved):

ndelgado:
maybe say "You must enable the same outbound [internet connectivity](/appliance/infrastructure/network#internet-connectivity) from the Arbiter for system updates as all other nodes in the cluster."

We aren't asking for entirely open outbound but outbound to those specific IP addresses and ports. 

Or if you think it's clear enough with the link, that is fine :)

ndelgado:
LGTM as is, thanks!

PR #6417, opened 2018-07-23
ws-fed: remove client id from URL

mpaktiti:
If you don’t add the client id you’ll get the metadata from the global client (the cert and other info from the tenant). If you add it you get the ones from a particular client. We would like to have only the one from the tenant, so we are removing the guidance that says otherwise. Any requests to support the client id should be forwarded to crew-auth, alongside with info on the use case from the customer.


mpaktiti:
Not sure what happened here, closing and opening a new PR

`mpaktiti wants to merge 1 commit into mpaktiti-patch-1 from mpaktiti-patch-1-1`

PR #6426, opened 2018-07-25
Quickstarts / Webapp / Node-js: Merge profile and user login + general improvements

albertoperdomo:
* Merge user profile and login into one single step
* Explain which middleware modules need to be installed and what they are used for
* Remove unnecessary code
* Explain how `ensureLoggedIn` middleware is used
* Clean up includes


albertoperdomo:
I've tested the heroku deployment. I have doubts about [this section](https://auth0-docs-content-pr-6426.herokuapp.com/docs/quickstart/webapp/nodejs#specify-a-profile-scope). It is needed at all? Why express (AFAIK there is no other mention of express in the document)?

What do you think?

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +15,158 @@ github:
...
+npm install passport passport-auth0 connect-ensure-login --save
+
+# installation with yarn
+yarn add passport passport-auth0 connect-ensure-login
+```
+
+### Configure the Middleware
+
+Create a new instance of the `Auth0Strategy` strategy. 
+Enter your Auth0 application details as configuration values. Tell `passport` to use this strategy.
```
aaguiarz: of the 'Auth0Strategy' passport strategy.

albertoperdomo:
* Add section on `layout.pug`, copying the code from the sample.
* Remove logout link from `user.pug`

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
 ---
 <%= include('../_includes/_getting_started', { library: 'Node.js', callback: 'http://localhost:3000/callback' }) %>
 
-<%= include('_includes/_login') %>
+## Configure Node.js to Use Auth0 
+
+### Install and load the Dependencies
+
```
aaguiarz: I think "dependencies" should go in lowercase

Maybe 'install and instantiate' instead of Load? 'install and configure'?

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+To follow along this guide, install the following dependencies.
+
+* [passport](http://www.passportjs.org/) - an authentication middleware for Node.js
+* [passport-auth0](https://github.com/auth0/passport-auth0) - an Auth0 authentication strategy for Passport
+* [express-session](https://www.npmjs.com/package/express-session) - an Express middleware to manage sessions
+* [connect-ensure-login](https://github.com/jaredhanson/connect-ensure-login) - a middleware to ensure the user is logged in order to access certain routes
```
aaguiarz: we say 'an Express' middleware and then 'a middleware'. We should use the same term in both.

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+if (app.get('env') === 'production') {
+ sess.cookie.secure = true; // serve secure cookies, requires https
+}
+
+app.use(session(sess));
+```
+
+### Configure Passport with the Auth0 Strategy and Your Auth0 Application Settings
+
+In `app.js`, include the `passport` and `passport0-auth0`modules, configure Passport to use a new instance of `Auth0Strategy` with your Auth0 application settings. Load the `passport.initialize()` and `passport.session()` middlewares to initialize Passport with persistent login sessions.
```
aaguiarz: modules, *and* configure Passport

(Add the "and")

>Load the `passport.initialize()` and `passport.session()` middlewares to initialize Passport with persistent login sessions.

I don't think we are 'loading them to initialize them', maybe 'Use the ..'

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+}
+
+app.use(session(sess));
+```
+
+### Configure Passport with the Auth0 Strategy and Your Auth0 Application Settings
+
+In `app.js`, include the `passport` and `passport0-auth0`modules, configure Passport to use a new instance of `Auth0Strategy` with your Auth0 application settings. Load the `passport.initialize()` and `passport.session()` middlewares to initialize Passport with persistent login sessions. 
+
+Please use the client secret from the application settings in your Auth0 dashboard.
```
aaguiarz: Maybe this comment can be above the clientSecret line

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+
+In this example, we will implement the following routes:
+
+* a route `/login`, that will trigger authentication by calling Passport's `authenticate` method. The user will be redirected to the login page.
+* a route `/callback`, that will be finalize the authentication when the user returns from the Auth0 login page and redirects to the profile page on sucessful login.
+* a `/user` route, to display the user's profile.
+* a route `/logout`, that closes the local user session and redirects the user again to the root index `/`.
+
+In the authentication step, make sure to pass the scope parameter with values `openid email profile` in order to access email and the other attributes stored in the user profile within the tenant. This will be needed in order to display the user's information on the profile page.
+
```
aaguiarz: I'd remove 'within the tenant'

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+The user's profile route should only be accessible if the user is logged in. Use the `ensureLoggedIn` middleware for this purpose. To have full access to the user profile on `userProfile`, stringify the `user` object.
+
+:::note
+This tutorial implements logout by closing the local user session. After logging out, the user will continue to have an open session at the tenant level. For other implementations please refer to [our logout docs](https://auth0.com/docs/logout). 
+:::
```
aaguiarz: 'at the tenant level'

'in the Auth0 server"?

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+  req.logout();
+  res.redirect('/');
+});
+
+module.exports = router;
+```
+
+Include this new router in your app, by requiring the file in the top section of  `app.js`:
+
+```js
```
aaguiarz: 'this new router' should be 'this new route'

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+// Perform session logout and redirect to homepage
+router.get('/logout', (req, res) => {
+  req.logout();
+  res.redirect('/');
+});
+
+module.exports = router;
+```
+
+Include this new router in your app, by requiring the file in the top section of  `app.js`:
```
aaguiarz: extra whitespace before 'app.js'

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+
+//..
+//..
+
+app.use('/', indexRouter);
+//..
+app.use('/', authRouter);
+//..
+```
+
```
aaguiarz, line 186: Maybe we need to explain the image (e.g. run the app, click 'Login' and you'll see this)

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+extends layout
+
+block content
+  br
+  img(src=user.picture)
+  p Welcome #{user.nickname}!
+  br
+  h2 User Profile
+  pre #{userProfile}
+```
```
aaguiarz: Maybe we need to wrap up the document saying "if your un the app and navigate to .." or "you log in', you'll be able to see the user profile after logging in.

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,239 @@ github:
...
+
+## Implementing Login, User Profile and Logout
+
+### Adding the Routes
+
+In this example, we will implement the following routes:
+
+* a route `/login`, that will trigger authentication by calling Passport's `authenticate` method. The user will be redirected to the login page.
+* a route `/callback`, that will be finalize the authentication when the user returns from the Auth0 login page and redirects to the profile page on sucessful login.
+* a `/user` route, to display the user's profile.
```
aaguiarz: >"that be finalize",

"that will be called when authentication finishes". 

Is not that the user returns, he'l be returned from Auth0

aaguiarz (changes requested):
Added some comments

albertoperdomo:
>>### Install and load the Dependencies
>I think "dependencies" should go in lowercase

The capital letter came from the title in the Rails Quickstart, which I just copied.
There, titles are capitalized. For now it seems that it is inconsistent, so don't know how much effort to put into that.

>Maybe 'install and instantiate' instead of Load? 'install and configure'?

Will go for _Install the Dependencies_, which is accurate and shorter.

>we say 'an Express' middleware and then 'a middleware'. We should use the same term in both.

I am not experienced with Node.js. I know that `express-session` is a middleware specific to the `Express` framework. With `connect-ensure-login`, the description states is a middleware, so I am not sure if this would be usable with other Node.js frameworks?
I think I'll address it by removing the mention of Express. It will still work and it will be consistent.

>Maybe this comment can be above the clientSecret line

So, you prefer that comment in the code block, instead of in the text?
Example:

```
//passport-auth0
var strategy = new Auth0Strategy({
  domain: '${account.namespace}',
  clientID: '${account.clientId}',
  // Please use the client secret from the application settings in your Auth0 dashboard.
  clientSecret: 'YOUR_CLIENT_SECRET',
  callbackURL: 'http://localhost:3000/callback'
 },
```

>'this new router' should be 'this new route'

I think in this case it's a router. It is a file defining multiple routes using the router framework.
In fact, the variable for each imported file is also named `somethingRouter`, see:

```
var indexRouter = require('./routes/index');
var authRouter = require('./routes/auth');
```

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,247 @@ github:
...
+In `app.js`, include the `passport` and `passport0-auth0` modules, and configure Passport to use a new instance of `Auth0Strategy` with your Auth0 application settings. Use `passport.initialize()` and `passport.session()` to initialize Passport with persistent login sessions.
+
+```js
+// app.js
+var Auth0Strategy = require('passport-auth0'),
+    passport = require('passport');
+
+//passport-auth0
+var strategy = new Auth0Strategy({
+  domain: '${account.namespace}',
```
jeffreylees, line 70: One question I have in the Quickstarts (other docs too) - do we want some kind of panel to help the user find this information, in case they are not logged in to their Dashboard, when browsing the Docs (or not seeing the info for the correct Application)? We've had the feedback before that customers were unaware of where to find this information, if they had never done so or use it infrequently.

jeffreylees, line 70: Another note - `${account.namespace}` will grab the tenant URL such as `foo.auth0.com`. However, if the customer is using a Custom Domain (`login.foo.com`) this URL should probably be that one, right? If so, we should probably communicate that somehow here, but without getting bogged down in those details too much.

articles/quickstart/webapp/nodejs/01-login.md
```diff
@@ -14,4 +16,247 @@ github:
...
+
+app.use(passport.initialize());
+app.use(passport.session());
+```
+
+### Storing and retrieving user data from the session
+
+In a typical web application, the credentials used to authenticate a user are only transmitted during the login request. If authentication succeeds, a session is established and maintained via a cookie set in the user's browser. Each subsequent request does not contain credentials, but rather the unique cookie that identifies the session.
+
+To support login sessions, Passport serializes and deserializes user instances to and from the session. Optionally, you may want to serialize only a subset to reduce the footprint, i.e., `user.id`.
```
jeffreylees, line 93: I do not know Passport, but this section is still unclear to me - what is it doing by default, and why would I want to change it, and is the below an example of me changing it? It looks like the example is just serializing `user`  - is that already happening, or do I need to add the below snippet to my `app.js`?

jeffreylees:
@albertoperdomo added some copyedits (just took the liberty of committing them) and also added some comments for your consideration. Ping me if I can help further!

albertoperdomo:
Thanks for the changes and the comments @jeffreylees.
Regarding your comments:

>do we want some kind of panel to help the user find this information, in case they are not logged in to their Dashboard, 

Valid question. I will discuss with Andres if there's anything we can do in the short-term to address this and then keep it mind to see if we can tackle this in the future with a better approach.

>Another note - ${account.namespace} will grab the tenant URL such as foo.auth0.com. However, if the customer is using a Custom Domain (login.foo.com) this URL should probably be that one, right? If so, we should probably communicate that somehow here, but without getting bogged down in those details too much.

I don't know about this. @aaguiarz Do you know what behavior we can expect in such a case? Is it worth adding a callout or note?

>I do not know Passport, but this section is still unclear to me - what is it doing by default, and why would I want to change it, and is the below an example of me changing it?

It is needed indeed. I can try to emphasize that it is needed, with this being the most simple implementation of it. A more advanced approach would be to serialize only the user id, then on lookup, find the User in the local database by looking up the user id.

albertoperdomo:
>Another note - ${account.namespace} will grab the tenant URL such as foo.auth0.com. However, if the customer is using a Custom Domain (login.foo.com) this URL should probably be that one, right? If so, we should probably communicate that somehow here, but without getting bogged down in those details too much.

After discussing with @aaguiarz, it might make more sense to introduce a new placeholder for the domain, which will resolve to the custom domain if one is in use.

aaguiarz (approved):

Issue #6420, opened 2018-07-24
Missing Instructions to install dependency on download page for React Native


rafanog:
At the [React Native download page](https://github.com/auth0/docs/blob/master/articles/quickstart/native/react-native/download.md), he instructions to install the dependencies is missing on the download page for the React Native example, to execute the sample code, the user needs to make a **npm install** command to download all the necessary dependencies.

![image](https://user-images.githubusercontent.com/15782359/43135910-775d8550-8f1d-11e8-8b1e-65dec2d7f45e.png)

valerianpereira:
I believe this is done already. Could you close the issue?

albertoperdomo:
It turns out I fixed it myself in August but I failed to close this issue.
Thanks for the catch!

PR #6430, opened 2018-07-26
Quickstarts React-native: Improve download instructions

albertoperdomo:
This PR contains two changesets:

* [the first commit](https://github.com/auth0/docs/commit/92281676672a5e9dc33defa6e8a9f899d5ae348c) addresses #6420 and adds npm install as a step to run the sample
* [the second commit](https://github.com/auth0/docs/pull/6430/commits/2ca4af1698b9c585f34354574348903b874520f1) is an experiment to add the comments in the bash section in-line to prevent the modal window from getting to large and keep the download button visible. This can be reverted if undesired.


aaguiarz (commented):
LGTM
@rafanog can you verify it works?

rafanog (approved):
Solves the problem of the missing instruction and also improves how the commands are displayed.

PR #6437, opened 2018-07-27
Updates Sumo Logic log extension documentation.

dmark:
1. Moves the Sumo Logic http endpoint stuff up top (you need the endpoint before you can configure the extension)
2. Adds details about additional configuration parameters.
3. A few other minor edits.


mpaktiti:
Good thinking @dmark , thanks for changing that. I'm pushing some copyedit changes and will merge in a bit. Thanks again!

mpaktiti (approved):

PR #6468, opened 2018-08-01
Auth0.js + hosted login page: add section on additional parameters

mpaktiti:


mpaktiti:
Not sure if we want to document this https://trello.com/c/omILLWfS/2523-38946-can-not-send-additional-parameter-with-authorize-call

PR #6472, opened 2018-08-02
Clarify 01-Login steps

joshcanhelp:
A few changes to make this quickstart a little more clear:

- `/dashboard` -> `/dashboard/show` to match the route
- Remove Auth0 styleguide HTML that will not be styled
- Add a note about using that login URL on an existing view
- Remove `--skip-template-engine` from the dashboard controller create command
- Add `session` data output to the dashboard/show view

### DIFF
--- a/articles/api/authentication/_login.md
+++ b/articles/api/authentication/_login.md
@@ -69,13 +69,13 @@ Social connections only support browser-based (passive) authentication because m
 
 1. At the *Configuration* tab, set the fields **Application** (select the application you want to use for the test) and **Connection** (the name of the social connection to use).
 
-1. Copy the **Callback URL** and set it as part of the **Allowed Callback URLs** of your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings).
+1. Copy the **Callback URL** and set it as part of the **Allowed Callback URLs** of your [Application Settings](${manage_url}/#/applications).
 
 1. At the *OAuth2 / OIDC* tab, click **OAuth2 / OIDC Login**.
 
 ### Remarks
 
-- The `redirect_uri` value must be specified as a valid callback URL under your [Application's Settings](${manage_url}/#/applications/${account.clientId}/settings).
+- The `redirect_uri` value must be specified as a valid callback URL under your [Application's Settings](${manage_url}/#/applications).
 
 - If `response_type=token`, after the user authenticates on the provider, it will redirect to your application `callback URL` passing the Access Token and ID Token in the address `location.hash`. This is used for Single Page Apps and also on Native Mobile SDKs.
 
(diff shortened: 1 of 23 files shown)


albertoperdomo:
Hello @joshcanhelp.
Thanks for taking the time to put this PR together.

I also followed the Quickstart a few days ago and my impression was that there was info missing regarding the routes:

* A root route needs to be added so that the user does not see the standard Rails welcome page. We do have code to implement `home#show`, but there is no guidance on how to actually use that. I added ` root to: "home#show"`
* The redirects where pointing to `dashboard` so rather than changing those like you did, I added a route `get 'dashboard' => "dashboard#show"`.

I had an item on my todo list to make this changes, so your timing for the PR is very good :)

What do you think?

joshcanhelp:
> A root route needs to be added so that the user does not see the standard Rails welcome page.

Good point. In fact a lot of that is off as well. I already had a home controller so I skipped some of it but you can do:

```bash
rails generate controller home show 
```

.. and you'll get the template. 

But I think an important question here is ... do we need to be instructing someone on how to make a home page? In other words, are we starting from nothing or assuming someone already has an app? If the former, then we probably also want to include `rails new` and also make sure we're adding notes for folks that already have an app (no need to generate a root/home path if you already have one). 

> The redirects where pointing to dashboard so rather than changing those like you did, I added a route get 'dashboard' => "dashboard#show".

I like that better!

I also found a few double quotes that could be single. 

Let me know on the first question and I can add that to here.

albertoperdomo:
> `rails generate controller home show`

The code for the controller and the view is already in the Quickstart. It's just not hooked up onto any route. We do ask the developer to change the routes for auth and callback, but the route for home is not mentioned anywhere.

For an experienced Rails developer, I think they will be able to figure out.

>But I think an important question here is ... do we need to be instructing someone on how to make a home page?

To your point, we are currently re-evaluating our criteria for what should be covered and what not. Whatever direction we agree on, I think the current implementation in this Quickstart is inconsistent.
I could see this going in following directions:

* leaving it up to the user to have somewhere a view with a link pointing to `/login`
* be explicit, include a dumb homepage and a link. The seasoned developer will know how to merge that into their existing codebase.

I think including the code for the controller and the view, but no mention on how to use them, is inconsistent. For the sake of consistency, I'd include the reference to the route, which is just asking the user to map `:root` to `home#show`.

If we later decide to cut back on the non-essential steps, we might consider removing any reference to the `home` controller, the view for `home#show` and the route mapping for it.

What do you think?

joshcanhelp:
> The code for the controller and the view is already in the Quickstart. It's just not hooked up onto any route. We do ask the developer to change the routes for auth and callback, but the route for home is not mentioned anywhere.

You're right ... I mean that you can add `show` to that command and not have to include the view code sample. 

> For an experienced Rails developer, I think they will be able to figure out.

IMHO, we don't want to assume too much. Might be less of a case of experience and more of just following each step and then suddenly it's wrong. 

> leaving it up to the user to have somewhere a view with a link pointing to `/login`

I vote for this one. Generating a new URL would be the most helpful for both types of devs: ones with an existing app and ones starting from scratch. 

So, in total, my recommendation for this PR based on what we talked about is:

- Change `dashboard/show` to `auth0/profile`
- Remove the `home`/`root` stuff and instead generate, say, `auth0/login` so we show how the link works but don't step on any toes
- Change the redirect on fail to `auth0/login`

Let me know what you think here/Slack and I'll make the changes!

joshcanhelp:
Thinking we might also want to add logout to this ...

albertoperdomo:
Yes, we are currently going through the existing Quickstart and putting together guidelines on what they should include. The current iteration states that we should include logout too.

That said I am aware that that goes beyond the scope of the fixes you are doing in this PR.

joshcanhelp:
Re-creating from a new branch, got mixed up in other commits.

PR #6479, opened 2018-08-02
angular2: don't clear window.location.hash in callback

bkylerussell:
Clearing window.location.hash on Firefox 61 (presumably to remove the user
hash from the URL) causes a page reload, so the user may observe a spurious
redirection to /home before landing back at /login/callback.  Setting
window.location.hash doesn't seem to trigger a reload on other browsers.

If the intent is to remove the hash from the URL, the following router
navigation to /home should be sufficient without clearing window.location.hash
separately.

https://community.auth0.com/t/auth0-angular-seed-application-does-not-properly-login-on-firefox/8698

This solution was proposed on the community form, and seems to address the
observed behavior.


joshcanhelp:
Out of date, statement no longer exists on this page. 

Appreciate the contribution, despite the very long delay in response!

PR #6480, opened 2018-08-02
Fix errors and add more information on routes, views for Rails quickstart

joshcanhelp:
Recreated from #6472

### DIFF
--- a/articles/quickstart/webapp/rails/01-login.md
+++ b/articles/quickstart/webapp/rails/01-login.md
@@ -42,7 +42,7 @@ This tutorial uses omniauth-auth0, a custom [OmniAuth strategy](https://github.c
 Use the following command to create the controller that will handle the Auth0 callback:
 
 ```bash
-rails generate controller auth0 --skip-template-engine --skip-assets
+rails generate controller auth0 --skip-template-engine --skip-assets --no-helper
 ```
 
 In the newly created controller, add success and failure callback handlers.
--- a/articles/quickstart/webapp/rails/01-login.md
+++ b/articles/quickstart/webapp/rails/01-login.md
@@ -49,6 +49,7 @@ In the newly created controller, add success and failure callback handlers.
 
 ```ruby
 # app/controllers/auth0_controller.rb
+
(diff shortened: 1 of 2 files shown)


albertoperdomo:
I have noticed that when creating the Auth0 controller, a helper `auth0_helper.rb` is created, which is not used. We ask the user to create another helper `session_helper.rb`.

Would it make sense to move the code from `session_helper.rb` to `auth0_helper.rb` or perhaps add a param to `rails generate` so that the helper is not created?

Thoughts, @joshcanhelp ?

albertoperdomo:
I walked through the quickstart again and it seems much better now.
I'm missing a logout implementation though. I know it was not present in the original user profile Quickstart, so we could address in a separate PR perhaps to get this out of the way.

What do you think?

joshcanhelp:
@albertoperdomo 

> Would it make sense to move the code from session_helper.rb to auth0_helper.rb or perhaps add a param to rails generate so that the helper is not created?

I would recommend the latter since it's possible to exclude helpers from loading based on the app config. I imagine that OmniAuth is auto-including the sessions file so I wouldn't want to change that. Good catch 👍 

> I'm missing a logout implementation ... we could address in a separate PR perhaps to get this out of the way.

Yeah, I think that's the right thing to do here. Let's get these changes out to get it in good shape, then add the logout. Happy to review or help write that when you're ready. 

Thanks!

joshcanhelp:
Fixed up a few syntax indicators and spacing. 

Ready for you @mpaktiti!

albertoperdomo:
>I would recommend the latter since it's possible to exclude helpers from loading based on the app config. I imagine that OmniAuth is auto-including the sessions file so I wouldn't want to change that. Good catch 👍

I have just added the param to skip the helper when generating the controller.

joshcanhelp:
@albertoperdomo - LGTM. Can you approve if this is ready to merge?

albertoperdomo (approved):

albertoperdomo:
Approved and merged!

PR #6492, opened 2018-08-06
Update _login.md

alexisluque:
Docs update for auth0-samples/auth0-servlet-sample#28.


jimmyjames:
Closing this PR as it's nearly a year old now, and if we do wish to update the UX of this sample we should use the new samples theme.

PR #6505, opened 2018-08-07
Be more explicit about opening xcworkspace in XCode.

albertoperdomo:
Addresses https://github.com/auth0-samples/auth0-ios-objc-sample/issues/43


aaguiarz (approved):
LGTM

PR #6514, opened 2018-08-08
Rename client to application in dynamic-application-registration.md

esarafianou:
Update the document to reflect the renaming of Clients to Applications

### DIFF
--- a/articles/api-auth/dynamic-client-registration.md
+++ b/articles/api-auth/dynamic-client-registration.md
@@ -1,6 +1,6 @@
 ---
-title: Dynamic Client Registration
-description: How to dynamically register clients with Auth0 using the Management API
+title: Dynamic Application Registration
+description: How to dynamically register applications with Auth0 using the Management API
 crews: crew-2
 toc: true
 topics:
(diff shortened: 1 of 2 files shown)


articles/api-auth/dynamic-client-registration.md
```diff
@@ -11,11 +11,11 @@ useCase:
   - call-api
 ---
 
-# Dynamic Client Registration
+# Dynamic Application Registration
```
mpaktiti: Hey @esarafianou, most changes in this doc make sense, but are you sure that we want to change the `Dynamic Client Registration` to `Dynamic Application Registration`? This is how this functionality is called in the OAuth 2.0 specification (see https://tools.ietf.org/html/rfc7591) so even though we renamed Clients to Applications in our Dashboard and terminology, this specific change looks wrong to me. Wdyt?

esarafianou: The reason I renamed this page is because that's how it's called in the side navigation menu:

screen shot 2018-08-11 at 6 41 27 pm

and also because under the tenants advanced settings we call it `OIDC Dynamic Application Registration`. 
However, if you prefer keeping it `Dynamic Client Registration`, we can do so but in such a case I think we'll have to keep the phrase _Dynamic Client Registration_  also in the side navigation menu and in lines 2, 18, and (maybe) 62 of this doc.

jeffreylees: Is there a consensus on which to do? I agree that we should be consistent throughout, either way.

esarafianou: Both seem valid to me. `Client` refers to OIDC and `Application` to our terminology. My intention was being consistent. We can use `Client` if we prefer using the OIDC here.

esarafianou: Shall I rename it to `Client` or leave it as is?

jeffreylees: Let's say `Client` and we can change the sidebar as well.

esarafianou: I updated the PR.

khorne3 (approved):

PR #6545, opened 2018-08-15
Fixed link

tamigoodall:
Fixed link to correct URL.


abbaspour (changes requested):
not quite sure this is the best fix. imho let’s keeping the link pointing to readable github js file. then ideally add a new section explaining steps to convert the js file to be used in `languageBaseUrl` mode as I mentioned.

tamigoodall:
Ok. I'll change the link back and add in steps. Can you @abbaspour provide the steps?

PR #6547, opened 2018-08-15
added link

tamigoodall:
Added link to SSO integrations that have built-in Auth0 support of which Salesforce is the one requested by the user.


tamigoodall:
Based on discussion with Jeff, closed this PR and will not update.

PR #6558, opened 2018-08-21
the allowed limit for each field (i.e. allowed callback URLs, allowed…

kamepo:
… web origins, allowed logout URLs) is 100

Not documented, but should be. Engineering determined this limit

### DIFF
--- a/articles/applications/webapps.md
+++ b/articles/applications/webapps.md
@@ -14,6 +14,10 @@ You'll need to define a Regular Web Application if you want to integrate Auth0 w
 
 <%= include('./application-settings/_token-endpoint-auth-method') %>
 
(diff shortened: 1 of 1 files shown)


khorne3:
@kamepo I'm assuming that this holds true regardless of application type, correct? If so, I think this would be best included in the _settings-pt2.md snippet.

Secondly, perhaps we could rephrase as follows?

> You can provide up to 100 URLs in the **Allowed Callback URLs**, **Allowed Web Origins**, **Allowed Logout URLs**, **Allowed Origins (CORS)** fields.

khorne3:
See https://github.com/auth0/docs/pull/6564

PR #6565, opened 2018-08-23
mmaddex update idp initiated 

mmaddex:
This PR includes the following changes
- Removes any mention of impersonation from the IdP Initiated page as I believe it's unrelated, please confirm this in PR review.
- Includes instructions for using Lock/Auth0.js on the IdP Initiated page

### DIFF
--- a/articles/protocols/saml/idp-initiated-sso.md
+++ b/articles/protocols/saml/idp-initiated-sso.md
@@ -13,16 +13,7 @@ useCase:
 
 # IdP-Initiated SSO
 
-If you are using [Auth0.js](/libraries/auth0js), you have to update the **webAuth.parseHash** of the [library](/libraries/auth0js/v9#extract-the-authresult-and-get-user-info) and set the flag **__enableIdPInitiatedLogin** to `true`.
-
-```javascript
-var data = webAuth.parseHash(
-  {
-    ...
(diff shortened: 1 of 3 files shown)


articles/protocols/saml/idp-initiated-sso.md
```diff
@@ -13,9 +13,7 @@ useCase:
 
 # IdP-Initiated SSO
 
-**Beginning with auth0.js v9.3.4, you must [enable the impersonation flags](/user-profile/user-impersonation#enable-impersonation) to use IdP-initiated login.**
```
jeffreylees, line 16: Is this statement not true / no longer true?

jeffreylees: Oh, I follow, you've just added that entire section here, so no need to link anymore

mmaddex: Yeah... and it looks about as bad as I imagined... I'll add a heading and move this down below to make this page less ugly.

mmaddex: @jeffreylees Ok, I think this looks better now. Feel free to carry on with the review process.

jeffreylees (approved):

PR #6591, opened 2018-08-30
Add new connection properties to rule's context

ghost:
This PR updates the docs to add the new connection properties which was completed as part of the work for this [RFC](https://auth0team.atlassian.net/wiki/spaces/ENG/pages/127863322/RFC-0007+Schema+for+Exposing+Connection+Details+in+Rule+s+Context) and [BC](https://auth0team.atlassian.net/wiki/spaces/PROD/pages/122618482/Business+Case+Add+Connection+Details+to+Rule+s+Context).

### DIFF
--- a/articles/rules/current/context.md
+++ b/articles/rules/current/context.md
@@ -23,6 +23,7 @@ The following properties are available for the `context` object.
 
 | Name | Description |
 |-|-|
+| `tenant` | A string containing the name of the tenant |
 | `clientID` | The client id of the application the user is logging in to. |
 | `clientName` | The name of the application (as defined on the dashboard). |
 | `clientMetadata` | An object for holding other application properties. Its keys and values are strings. |
@@ -37,6 +38,9 @@ The following properties are available for the `context` object.
 | `sso` | <%= include('./_context-sso.md') %> |
 | `accessToken` | Used to add custom namespaced claims to the [Access Token](/tokens/access-token). |
 | `idToken` | Used to add custom namespaced claims to the [ID Token](/tokens/id-token). |
+| `original_protocol` | After a [redirect rule](/rules/current/redirect) has executed and the authentication transaction is resumed, this property will be populated with the original protocol used to initiate the transaction. |
+| `multifactor` | An object representing the multifactor settings used in [implementing contextual MFA](/multifactor-authentication/custom). |
+| `redirect` | The object used to [implement the redirection of a user from a rule](/rules/current/redirect#how-to-implement-a-redirect). |
 | `sessionID` | Unique id for the authentication session. Value is kept only if `prompt=none`. |
 | `request` | <%= include('./_context-request.md') %> |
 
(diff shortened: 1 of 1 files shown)


articles/rules/current/context.md
```diff
@@ -23,35 +23,49 @@ The following properties are available for the `context` object.
...
 ```js
 {
+  tenant: 'mydomain',
   clientID: 'q2hn...pXmTUA',
   clientName: 'Default App',
   clientMetadata: {},
+  connectionID: 'con_V7s88...lgW97',
   connection: 'Username-Password-Authentication',
   connectionStrategy: 'auth0',
+  connectionOptions: {},
```
solepano, line 58: should we add examples for `tenant_domain` and `domain_aliases` inside options?

ghost: @solepano my concern is that we would have to change the current example since it's not an enterprise connection and it might confuse customers. I've documented them at the top of the document as an alternative: https://github.com/auth0/docs/pull/6591/files#diff-8800034d7401d684ab84602473faf32aR33

solepano: ok, can we do the same for all other objects?

solepano: also, do we know which fields are readable vs writable? something that jared pointed out here and might be interesting to know https://auth0.slack.com/archives/CBTDY4DJN/p1533321091000141?thread_ts=1533304955.000071&cid=CBTDY4DJN

solepano: and finally, why there's no review triggered from this PR? it would be easier to see the changes live. Maybe because you used a branch?

ghost: > why there's no review triggered from this PR?

I looked through the last 5-10 PRs and it seems that reviews are no longer generated :(

> ok, can we do the same for all other objects?

Sorry, could you clarify what you mean? Which objects are you referring to and what sort of changes?

> also, do we know which fields are readable vs writable?

I'd have to dig into this and get an understanding of what Jared means by readable vs writeable since they all vary depending on the property. i.e.: `context.redirect` is writeable and so is `context.connectionMetadata` but it won't persist beyond the Rules lifetime. It might make more sense to break that out into it's own story and investigate to avoid holding up the release of this specific feature.

@solepano

albertoperdomo:
>and finally, why there's no review triggered from this PR? it would be easier to see the changes live. Maybe because you used a branch?

FYI: Review apps are no longer generated automatically. You can still create a review app manually from the Heroku dashboard and it will be auto-destroyed automaticall after a certain number of days.

articles/rules/current/context.md
fyockm: > It might make more sense to break that out into it's own story and investigate to avoid holding up the release of this specific feature.

Completely agree. @FadyMak went beyond the call of duty for this particular PR. we could add another story to backlog if we feel this "readable vs writable" distinction is necessary.

khorne3 (approved):

PR #6607, opened 2018-09-05
Update to reflect terminology change-Client to Application

rkhoriander:


rkhoriander:
https://auth0-docs-content-pr-6607.herokuapp.com/docs/api-auth/which-oauth-flow-to-use

Flowchart updated

jeffreylees:
So, this is that weird terminology thing again, but I think that the box that now reads `Application Credentials Grant` needs returned to `Client Credentials Grant` - because that's the name of an Auth0-agnostic protocol (https://tools.ietf.org/html/rfc6749#section-4.4). I think the rest mostly make sense though.

jeffreylees (changes requested):
So, this is that weird terminology thing again, but I think that the box that now reads `Application Credentials Grant` needs returned to `Client Credentials Grant` - because that's the name of an Auth0-agnostic protocol (https://tools.ietf.org/html/rfc6749#section-4.4). I think the rest mostly make sense though.

rkhoriander:
Put in request and waiting on design to fix the wording.

rkhoriander:
Wording fixed.

jeffreylees (approved):

PR #6608, opened 2018-09-05
Note that user ID should be URL-encoded.

rkhoriander:


rkhoriander:
https://auth0-docs-content-pr-6608.herokuapp.com/docs/users/search/best-practices#users-by-id

jeffreylees (approved):

PR #6611, opened 2018-09-05
Update community-supported to correct div class

jerdog:
Update community-supported repos to have the correct div class to show differentiation


erjohnson (approved):
Thanks! 👌

PR #6612, opened 2018-09-05
Update saml hrd

mmaddex:
This PR removes the `selecting-between-multiple-idp.md` page and replaces references to it with links for Home Realm Discovery.

From the deleted page:

> * Specifying the Connection in the call invoking authentication;	
> * Specifying the email domain(s) for each Connection in the Connection Settings page of the Auth0 Dashboard;	
> * Adding custom buttons to the Lock widget so that users choose the Connection they want to use.

This third bullet does not apply to enterprise connections, the first two are better explained in the HRD doc.


jeffreylees (approved):
Added a redirect and I think this is 👍

PR #6615, opened 2018-09-06
Update modify.md

Spencer-Arbour:
The userPrincipalName misspelled as userPrincipleName can cause copy paste errors when modifying config.json


jeffreylees (approved):
Thanks!

PR #6651, opened 2018-09-20
Multiple Password Reset Warning

kamepo:
When a user requests multiple password resets, only the most recent is valid. A warning noting this.


nicosabena:
Already included the changes in https://github.com/auth0/docs/pull/6670, so no need to merge this.

PR #6658, opened 2018-09-20
additional error and solution for custom domains

kamepo:


rkhoriander (approved):

rkhoriander:
https://auth0-docs-content-pr-6658.herokuapp.com/docs/custom-domains#-you-should-not-be-hitting-this-endpoint-

PR #6681, opened 2018-09-24
Notes on Katana and web farms

nicosabena:
Katana uses ASP.Net's machine key settings to encrypt the session cookie and the state, so when using web farms it's important to configure the same machine key on all servers/applications.


rkhoriander:
https://auth0-docs-content-pr-6681.herokuapp.com/docs/quickstart/webapp/aspnet-owin#web-farm-setup

PR #6707, opened 2018-10-02
Removed Mongo

tamigoodall:
Removed Mongo

### DIFF
--- a/articles/connections/database/custom-db/templates/index.md
+++ b/articles/connections/database/custom-db/templates/index.md
@@ -25,15 +25,13 @@ While Auth0 has populated default templates in the Dashboard script editor, you 
 Use the following checklist to make sure your scripts achieve the results you intend:
 
 1. **Set a `user_id` on the returned user profile that is consistent for the same user every time.**
-   In the migration scenario, this is important because if you try to set a random `user_id` for each login and you call `forgot password` twice, and then try to follow the second link, the user is unable to migrate. One of the user stores in Mongo gets confused if you have the same email address but more than one ID if you do the forgot password twice. The result is that every time a user logs in, they will get a new user created in the user list. 
-
-   In the non-migration scenario, this is important because you can end up with duplicate users in the Auth0 database.
+   In the migration scenario, this is important because if you try to set a random `user_id` for each login and you call `forgot password` twice, and then try to follow the second link, the user is unable to migrate. In the non-migration scenario, this is important because you can end up with duplicate users in the Auth0 database.
 
 2. **If using a `username`, ensure that you aren't returning the same email address for two different users in the `get_user` or `login` script.**
    Auth0 will produce an error if you do this, but it is better to catch it in the script itself. 
 
 3. **If setting `app_metadata`, call it `metadata` in the script.**
-   In Mongo, `app_metadata` is called `metadata`. Everywhere that `metadata` is used, it is translated to `app_metadata` except in custom database scripts. If you don't use `metadata` in the script, you will get an error where `app_metadata` will work but if you use the API to merge `app_metadata` with a user, it will appear as if all of your `app_metadata` was lost. This happens because the script creates the new `metadata` field and won't return the original `app_metadata`.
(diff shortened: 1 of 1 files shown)


articles/connections/database/custom-db/templates/index.md
```diff
@@ -25,15 +25,13 @@ While Auth0 has populated default templates in the Dashboard script editor, you
 Use the following checklist to make sure your scripts achieve the results you intend:
 
 1. **Set a `user_id` on the returned user profile that is consistent for the same user every time.**
-   In the migration scenario, this is important because if you try to set a random `user_id` for each login and you call `forgot password` twice, and then try to follow the second link, the user is unable to migrate. One of the user stores in Mongo gets confused if you have the same email address but more than one ID if you do the forgot password twice. The result is that every time a user logs in, they will get a new user created in the user list. 
-
-   In the non-migration scenario, this is important because you can end up with duplicate users in the Auth0 database.
+   In the migration scenario, this is important because if you try to set a random `user_id` for each login and you call `forgot password` twice, and then try to follow the second link, the user is unable to migrate. In the non-migration scenario, this is important because you can end up with duplicate users in the Auth0 database.
```
mostekcm: Sorry, with more testing I've found you don't have to call forgot password twice, just once, so should just be:
```
In the migration scenario, this is important because if you try to set a random `user_id` in the get_user script and you call `forgot password`, and then change the password, the user will get duplicated every time they log in.  In the non-migration scenario if you set a random user_id you can end up with duplicate users for every login.
```

articles/connections/database/custom-db/templates/index.md
```diff
@@ -25,15 +25,13 @@ While Auth0 has populated default templates in the Dashboard script editor, you
...
    Auth0 will produce an error if you do this, but it is better to catch it in the script itself. 
 
 3. **If setting `app_metadata`, call it `metadata` in the script.**
-   In Mongo, `app_metadata` is called `metadata`. Everywhere that `metadata` is used, it is translated to `app_metadata` except in custom database scripts. If you don't use `metadata` in the script, you will get an error where `app_metadata` will work but if you use the API to merge `app_metadata` with a user, it will appear as if all of your `app_metadata` was lost. This happens because the script creates the new `metadata` field and won't return the original `app_metadata`.
+   Everywhere that `metadata` is used, it is translated to `app_metadata` except in custom database scripts. If you don't use `metadata` in the script, you will get an error where `app_metadata` will work but if you use the API to merge `app_metadata` with a user, it will appear as if all of your `app_metadata` was lost. This happens because the script creates the new `metadata` field and won't return the original `app_metadata`.
```
mostekcm: Due to a backwards compatibility effort, in custom DB scripts `app_metadata` is called `metadata`. If you don't use `metadata` in the script, you will get an error where `app_metadata` will work but if you use the API to merge `app_metadata` with a user, it will appear as if all of your `app_metadata` was lost.  NOTE: `user_metadata` is not affected by this and can simply be called `user_metadata`.

mostekcm (changes requested):

articles/connections/database/custom-db/templates/index.md
tamigoodall: Made this change with a slight modification: `In the migration scenario, this is important because if you set a random `user_id` in the `get_user` script, then call `forgot password` and change the password, the user will get duplicated every time they log in.  In the non-migration scenario, if you set a random `user_id` you can end up with duplicate users for every login.`

articles/connections/database/custom-db/templates/index.md
tamigoodall: Made change with the following slight modifications:
To support backwards compatibility, `app_metadata` is called `metadata` in custom DB scripts. If you don't use `metadata` in the script, you will get an error where `app_metadata` will work but if you use the API to merge `app_metadata` with a user, it will appear as if all of your metadata was lost. 

   ::: note
   `user_metadata` is not affected by this and can simply be called `user_metadata`.
   :::

mostekcm (approved):
LGTM

jeffreylees (approved):

PR #6714, opened 2018-10-03
Fix usernameStyle description

thameera:
This is correctly described later in the article, but the info in this particular table is incorrect


jeffreylees (approved):
Thanks for the catch!

PR #6716, opened 2018-10-03
Removed dead link

gitdev481:
/appliance/dashboard#appliance-controls no longer exists, so I deleted the broken link directing to it.


jeffreylees (approved):
Thanks for spotting the bad link - it just needed fixed though, not removed!

PR #6721, opened 2018-10-03
updated links to status page

StrikerRUS:
Replaced `http` with `https` and fixed links to region-specific pages. All previous region-specific urls redirected to the **US** region.


jeffreylees (approved):
Thanks!

PR #6734, opened 2018-10-05
Added deprecation warning

tamigoodall:
Added deprecation warning

### DIFF
--- a/articles/libraries/auth0-swift/passwordless.md
+++ b/articles/libraries/auth0-swift/passwordless.md
@@ -11,7 +11,7 @@ useCase: enable-mobile-auth
 ---
 # Passwordless Authentication with Auth0.Swift
 
--<%= include('../../_includes/_native_passwordless_warning') %>
+<%= include('../../_includes/_native_passwordless_warning') %>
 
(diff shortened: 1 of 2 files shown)


jeffreylees (changes requested):
So, it looks like you've just added another warning above the existing one, when probably you're wanting to re-word or add to the existing warning, which says some of the same things?

tamigoodall:
I see that now. I spaced on the line above. I updated the include file.

articles/libraries/auth0-swift/passwordless.md
```diff
@@ -11,7 +11,7 @@ useCase: enable-mobile-auth
 ---
 # Passwordless Authentication with Auth0.Swift
 
-<%= include('../../_includes/_native_passwordless_warning') %>
+-<%= include('../../_includes/_native_passwordless_warning') %>
```
jeffreylees:
```suggestion
<%= include('../../_includes/_native_passwordless_warning') %>
```

jeffreylees: Just trying out this suggested change feature, which is new? Anyway, not sure if that hypen is a stray mark?

tamigoodall:
I accepted your change to remove the leading dash in that line. I think it worked.

jeffreylees (approved):

Issue #5615, opened 2018-01-19
Native Windows Universal: Inconsistency between Sample Project and Doc snippets


paulioceano:
In the file App.xml.cs of the sample project, there is an extra code line in the OnLaunched method that doesn't appear in the Doc snippets for this method (https://auth0.com/docs/quickstart/native/windows-uwp-csharp#set-up-the-auth0-callback-url).

The line is ´this.DebugSettings.EnableFrameRateCounter = true;´

![capture-for-jira-screenshot-20180117-124745-160](https://user-images.githubusercontent.com/15818783/35155290-8f3cc430-fd03-11e7-869a-bb56bf72249d.png)
![image](https://user-images.githubusercontent.com/15818783/35155298-93ff2ada-fd03-11e7-8aa5-cfe3d7044f83.png)

albertoperdomo:
Don't think the FrameReateCounter is essential to integrate Auth0. Closing.

PR #6740, opened 2018-10-06
UWP: Inconsistency between Sample Project and Doc

danimart1991:
Fix Issue #5615


albertoperdomo:
Don't think this is issue is relevant. Sorry @danimart1991. See my comment on the issue.

PR #6741, opened 2018-10-06
Update Auth0 version for iOS

### LINKED ISSUE #3738: Outdated Versions - Lock and Auth0 pod packages

zkan:
Changes proposed in this pull request:
* Update Auth0 version for iOS in the Install Dependencies section.

Fixes #3738.

### DIFF
--- a/articles/quickstart/native/_includes/_ios_dependency_centralized.md
+++ b/articles/quickstart/native/_includes/_ios_dependency_centralized.md
@@ -5,7 +5,7 @@
 If you are using Carthage, add the following to your `Cartfile`:
 
 ```ruby
-github "auth0/Auth0.swift" ~> 1.0
+github "auth0/Auth0.swift" ~> 1.13
 ```
 
 Then, run `carthage bootstrap`.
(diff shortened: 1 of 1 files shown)


khorne3 (approved):

albertoperdomo:
Hello @zkan,

Looking at the Swift and Objective-C sample repos, it seems that they use newer versions.
Whenever possible, it would be great if we could keep the versions of dependencies in sync between the tutorials in docs and the corresponding sample repos.

I am not familiar with the way dependencies are managed on iOS.
Would you mind having a look at the sample repos?

https://github.com/auth0-samples/auth0-ios-objc-sample/
https://github.com/auth0-samples/auth0-ios-swift-sample

zkan:
@albertoperdomo Thanks for pointing me to them. Sure, I'll check them out.

PR #6748, opened 2018-10-08
Update dedicated-domains.md

matthew-nethercote:
Removed reference to specific appliance version as no longer relevant. Modified MAY to MUST for webtasks on dedicated domain, as this is mandatory and no longer optional.


matthew-nethercote:
@ndelgado can you please check wording is OK?

khorne3 (approved):

PR #6754, opened 2018-10-08
Updating Laravel version from 5.5 to 5.7 in urls

james-flynn-ie:
Updated Laravel version to current URLs.

### DIFF
--- a/articles/quickstart/webapp/laravel/01-login.md
+++ b/articles/quickstart/webapp/laravel/01-login.md
@@ -50,7 +50,7 @@ First, we need to add the Auth0 Services to the list of Providers in `config/app
 
 ${snippet(meta.snippets.setup)}
 
-If you want to use an `Auth0` facade, add an alias in the same file (not required, [more information on facades here](http://laravel.com/docs/5.5/facades)):
+If you want to use an `Auth0` facade, add an alias in the same file (not required, [more information on facades here](http://laravel.com/docs/5.7/facades)):
 
 ```php
 // config/app.php
--- a/articles/quickstart/webapp/laravel/01-login.md
+++ b/articles/quickstart/webapp/laravel/01-login.md
@@ -15,15 +15,15 @@ github:
 ---
 <%= include('../_includes/_getting_started', { library: 'Laravel', callback: 'http://localhost:3000/callback' }) %>
 
-## Install and Configure Laravel 5.5
+## Install and Configure Laravel 5.7
 
 If you are installing Auth0 to an existing app, you can skip this section. Otherwise, walk through the Laravel guides below to get started with a sample project.
 
-1. **[Installation](https://laravel.com/docs/5.5/installation)**
+1. **[Installation](https://laravel.com/docs/5.7/installation)**
     * Use any of the install methods listed to start a new project
     * PHP can be served any way that works for your development process (we use Homebrew-installed Apache and PHP)
     * Walk through the "Configuration" section completely
-2. **[Configuration](https://laravel.com/docs/5.5/configuration)**
+2. **[Configuration](https://laravel.com/docs/5.7/configuration)**
     * Create a .env file, used later for critical and sensitive Auth0 connection values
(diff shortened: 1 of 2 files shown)


albertoperdomo:
Thanks for looking into this @james-flynn-ie.

Looking at the [corresponding sample app](https://github.com/auth0-samples/auth0-laravel-php-web-app/blob/master/01-Login), it seems that it is on Laravel 5.6.12

In [composer.lock](https://github.com/auth0-samples/auth0-laravel-php-web-app/blob/master/01-Login/composer.lock):

```
        {
            "name": "laravel/framework",
            "version": "v5.6.12",
            "source": {
                "type": "git",
                "url": "https://github.com/laravel/framework.git",
                "reference": "82d8165d1ea86bdd81ddfa1db9343fa19e7d1450"
            },
```

Ideally, we'd keep both versions on the tutorial and the sample in sync.
What do you think @joshcanhelp?

* Does it makes sense to set the version in the tutorial to 5.6.12?
* How easy would be to bump the version of the sample to 5.7?

joshcanhelp:
@albertoperdomo - I would set both to be the same, so if we bump the sample to 5.7, I'd bump the QS as well. 

Laravel 5.5 requires PHP 7.0 or higher:

https://laravel.com/docs/5.5/upgrade#upgrade-5.5.0

5.6 (and probably 5.7 as well since it does not mention a PHP req change) requires 7.1.3 or higher:

https://laravel.com/docs/5.6/upgrade#upgrade-5.6.0

The sample is already set at `7.1.3` so moving that to 5.7 would not make a difference. As for what to put on the Quickstart ... moving from 5.5 to 5.7 is a bump in PHP version. Our code should be fine, I don't see anything in there that's deprecated between the two:

https://secure.php.net/manual/en/migration71.deprecated.php

Also, I _just_ ran through the QS in 5.7 and did run run into any problems or differences. 

So ... after typing all that out ... LGTM 😄

joshcanhelp (changes requested):
There is one more instance of `5.5`, search for that here:

https://raw.githubusercontent.com/auth0/docs/7bc958819f3c16a079d9daf16e4afa58c8f61aea/articles/quickstart/webapp/laravel/01-login.md

There are also 2 instances in the header, not entirely sure how those are changed:

screenshot 2018-10-11 12 02 16

albertoperdomo:
Regarding the remaining references:

* The link below "I want to integrate with my app" is automatically added from the titles in the document, so that should be ok.
* The system requirements below that are set in the `index.yml` in the same directory.

Regarding bumping the sample:

* I'd bump the sample to 5.7 too. Would you volunteer do do that @james-flynn-ie?

albertoperdomo:
Added `do-not-merge` label until sample is bumped to make sure they are in sync.

james-flynn-ie:
Thanks for the review, Alberto and Josh.

I've updated the URL on login.md as requested.

The System Requirements were stated in index.yml, and the Laravel version has been updated there.

With regards to the other instance of '5.5' in the header, this seems to be taken from the markdown tags in 01-login.md. Looking at the source for the web page, React seems to fetch the doc's layout using a GET query. The previously updated headers in 01-login.md should populate this automatically.

james-flynn-ie:
Updated sample, https://github.com/auth0-samples/auth0-laravel-php-web-app/pull/30

joshcanhelp (approved):
LGTM! Thank you!

james-flynn-ie:
Hi, @albertoperdomo , do we need to remove the do-not-merge tag since it has been reviewed by Josh?

albertoperdomo:
Forgot to merge this last week, sorry.
I will do on Wednesday along with https://github.com/auth0-samples/auth0-laravel-php-web-app/pull/30

albertoperdomo:
I want to make sure both app and tutorial are updated more or less at the same time.
In order to do that I have to wait.

albertoperdomo:
Merged! Thanks, @james-flynn-ie!

Issue #4746, opened 2017-07-31
System requirement section with wrong version in Symfony Doc


paulioceano:
In the System Requirement section on https://auth0.com/docs/quickstart/webapp/symfony, show a requirement for Symfony 3.*, but the latest Symfony version is 3.3 and the Sample project uses version 2.8

![3359-1](https://user-images.githubusercontent.com/15818783/28762011-1ad17726-7581-11e7-97d0-e66c6bdab0c9.png)

![3359-2](https://user-images.githubusercontent.com/15818783/28762029-352fe40e-7581-11e7-80f0-3a6b49777e54.png)

AUTH-3359

arnas:
I will take it

albertoperdomo:
Closed via #6780.

PR #6780, opened 2018-10-11
Update symfony version

arnas:
Referecing https://github.com/auth0/docs/issues/4746 issue.
If found that the version of the example app is in fact 3.3* https://github.com/auth0-community/auth0-symfony-php-web-app/blob/master/00-Starter-Seed/composer.json#L31  . 
But as example uses greater version than the requirements . Maybe it would be good to update the minimal version in the documentation? In case if there is some breaking changes between 3.* and 3.3

### DIFF
--- a/articles/quickstart/webapp/symfony/index.yml
+++ b/articles/quickstart/webapp/symfony/index.yml
@@ -21,7 +21,7 @@ github:
   branch: master
 requirements:
   - PHP 5.6, 7.0
-  - Symfony 3.*
+  - Symfony 3.3.*
 next_steps:
   - path: 01-login
     list:


albertoperdomo:
Does it make sense to bump the Symfony version in the requirements section @joshcanhelp ?
Seem like sample is on 3.3 anyway, so it looks good to me.

albertoperdomo:
Closes #4746.

joshcanhelp:
Should not be breaking changes within a major but I don't see any reason not to update here.

joshcanhelp (approved):

albertoperdomo:
Thanks @joshcanhelp and @arnas !

PR #6794, opened 2018-10-15
Update User search v2 note with service end date

csvn:
https://auth0.com/docs/users/search/v3#migrate-from-search-engine-v2-to-v3

The migration guide says this API will be removed on November 13 2018. Is the migration guide out of date or this page?

### DIFF
--- a/articles/users/search/v2/index.md
+++ b/articles/users/search/v2/index.md
@@ -16,7 +16,7 @@ useCase:
 # User Search
 
 ::: version-warning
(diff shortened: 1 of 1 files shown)


khorne3:
Hey @csvn! The note refers to newly-created tenants, which do not have v2 functionality (existing tenants currently do, but the features will be removed come November). Does that make sense?

csvn:
Alright, thanks for the info @khorne3.

I landed on https://auth0.com/docs/users/search/v2 from Google, and could not find any info there on when v2 would be removed from cloud tentants. Wanted to double check that I had the correct date before I started trying to migrate.

PR #6805, opened 2018-10-17
Swap catches in Laravel middlewares

philbates35:
As `Auth0\SDK\Exception\InvalidTokenException` extends `Auth0\SDK\Exception\CoreException`, currently the `catch (InvalidTokenException $e)` block will never be reached because
all `InvalidTokenException`s will be caught in the `catch (CoreException $e)` block instead.

Swapping them around allows handling `InvalidTokenException` separately from all other `CoreException`s.

Edit: Accompanying PR for the sample Laravel API project is auth0-samples/auth0-laravel-api-samples#18.

### DIFF
--- a/articles/quickstart/backend/laravel/01-authorization.md
+++ b/articles/quickstart/backend/laravel/01-authorization.md
@@ -169,9 +169,9 @@ class CheckJWT
 
             \Auth::login($user);
 
-        } catch (CoreException $e) {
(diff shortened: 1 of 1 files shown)


albertoperdomo (approved):
Looks good according to https://github.com/auth0-samples/auth0-laravel-api-samples/pull/18

PR #6806, opened 2018-10-17
Update application-settings.md with correct format

sjwarner-bp:
Update note formatting as indicated in CONTRIBUTING.md, in order to display correctly.


rkhoriander (approved):

jeffreylees:
I'll verify this, but I believe the thing that needs fixed is the contributing guide, to remove that spacing which is unnecessary. Thanks though!

PR #6807, opened 2018-10-17
Fix question for SPAs

albertoperdomo:
SPA navigation is presented the question for regular web apps.

screen shot 2018-10-17 at 1 24 42 pm


albertoperdomo:
After a discussion with @aaguiarz, decided to not make this update. Closing.

PR #6814, opened 2018-10-18
Update

zaidishahbaz:
A working redirecting link to the user mangement dashboard at Auth0


rkhoriander:
The link is a variable that is replaced on our docs site. Leaving as is.

PR #6822, opened 2018-10-20
Added Auth0 Ruby to SDKs list

szemek:
There is a missing SDK name (https://auth0.com/docs/libraries)
image

I replaced it with `Auth0 Ruby` and `v4`.


jeffreylees:
The missing ones got lost during the last change to this doc, but it wasn't Ruby. We will add those back in, but adding Ruby to the support list would require some conversation with others in our org.

The team that oversees SDKs plans to overhaul this supported list sometime soon, so it'll be more reflective of the entire offering we have out there.

Thanks though!

szemek:
Thanks for the explanation.

PR #6826, opened 2018-10-20
Rules.md

kaiiyer:
Added Extra Rules for better clarity


jeffreylees:
Thanks for the contribution. Sorry, but these, while good points, don't go with this content. This content is about best practices for using the Rules feature of Auth0, which these pointers don't pertain to.

Thanks again though!

PR #6827, opened 2018-10-21
Update operations.md

cozyplanes:
Replace -> with →


jeffreylees:
Hi, sorry, this doesn't seem to be a necessary change. I think the syntax we use most often is actually `>` rather than either of those, but it will be more efficient to sweep the docs and change them all at once. Thanks though!

PR #6836, opened 2018-10-22
Fix quotes

kimdhendrick:

### DIFF
--- a/articles/quickstart/backend/rails/01-authorization.md
+++ b/articles/quickstart/backend/rails/01-authorization.md
@@ -157,7 +157,7 @@ The `/public` endpoint does not require to use the `Secured` concern.
 class PublicController < ActionController::API
   # This route doesn't need authentication
   def public
-    render json: { message: 'Hello from a public endpoint! You don't need to be authenticated to see this.' }
+    render json: { message: "Hello from a public endpoint! You don't need to be authenticated to see this." }
   end
 end
 ```


albertoperdomo:
No need to change the quotes. Closing.

kimdhendrick:
Why? This is invalid Ruby code as-is.

kimdhendrick:
@albertoperdomo Maybe I should have provided a description. This change is necessary because the message `Hello from a public endpoint! You don't need to be authenticated to see this.` has a single quote inside the `don't` and thus needs to be wrapped in double quotes `"` instead of single quotes `'`. 
As is, the Ruby code is invalid and breaks when users copy/paste this code.
Please re-open.

albertoperdomo:
Ok. Got it now.
Sorry I failed to see the issue with the single quote. Your explanation helps and you are right.

In fact, I just checked the sample at https://github.com/auth0-samples/auth0-rubyonrails-api-samples/blob/master/01-Authentication-RS256/app/controllers/public_controller.rb and I see an escape character is used, which is way uglier.

albertoperdomo:
In case you'd be interested to do similar fixes, you could submit a similar change in the sample repo (there may be multiple occurrences of this there).

albertoperdomo (approved):

albertoperdomo:
Thanks for the catch @kimdhendrick and for following up!

PR #6844, opened 2018-10-23
Update SLA/SLD percentage to match homepage

arusahni:
On your on main landing page, you advertise 99.95% uptime, while the docs list it as 99.90%.

screen shot 2018-10-23 at 1 59 14 pm


jeffreylees:
Hi! Thanks for the contribution, but the regular SLA is in fact 99.90%. 99.95% is the enterprise level SLA, which is indicated on the auth0.com site in the Enterprise addons area. Thanks though!

PR #6859, opened 2018-10-24
Translation of Words file

xvercruysse:
Localization in French of the [Words](WORDS.md) file


khorne3:
Thanks so much for your contribution! Unfortunately, we are accepting only those submissions for which we have resources to maintain. If we have plans for French translations, we will be in touch!

PR #6860, opened 2018-10-24
Update information on what authentication flow we use.

konradsopala:
**Short description**

The change is all about implicitly saying what authentication flow we use. 

**Reason**

https://community.auth0.com/t/which-oauth-2-0-flow-for-a-chrome-extension/15899/2


joshcanhelp:
Not necessary to add, IMHO.

PR #6861, opened 2018-10-24
Create README-TR.md

ghost:
Turkish version of README.md


khorne3:
Hi @aytekk. Thanks for your contribution! Unfortunately, we're only accepting submissions for which we have the resources to maintain in the future. If we need Turkish translations, we will be in touch!

PR #6862, opened 2018-10-24
Ia rework

rkhoriander:


khorne3:
@rkhoriander Can this be closed?

rkhoriander:
@khorne3, I think we should hang onto this for just a little bit longer. I've used it as a reference when redoing some of the chunking/rewriting of various sections because we all were changing and dumping things into it, and it's easier for me to see what changes we all made so I don't lose anything/reinvent the wheel. I should be finished with it after the next sprint and will make sure I close it then.

jeffreylees:
Any update on this?

rkhoriander:
@jeffreylees, if we could keep it open for one more sprint, we should be good. The focus of the last sprint I was talking about changed, but as long as I can get the final item in this next sprint, I should be able to close this out with no issue.

PR #6871, opened 2018-10-25
fix dead links in articles/api/management/v2/tokens.md

helmus:
Fixing dead links. This documentation should probably be generated or something because manually updating this links doesn't really seems sustainable. 

Happy hacktober !


jeffreylees:
Hi! They actually don't need the `/docs` at the beginning, but someone just fixed this batch of links in the tokens page internally a bit ago. Thanks though!

PR #6874, opened 2018-10-26
update index.md

diptinayan:


jeffreylees:
Sorry, but this is not a needed change. We use relative links, and the link doesn't need an explanation to click. Thanks though!

PR #6878, opened 2018-10-26
Added lines 30-33 (scalability)

ishan-sriv:
second last paragraph


khorne3:
Thanks for your contribution! Unfortunately, while informative, I don't think the content is best served on this page.

PR #6885, opened 2018-10-29
fixed some grammatical errors

swetajena98:


iamnishantmishra (changes requested):
No grammatical problems are there in the original text. Only the word 'reference' in line 24 can be replaced with 'refer'

khorne3 (approved):

PR #6894, opened 2018-10-30
Created a modern README.md page for auth0 repo!

ssiddhantsharma:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,18 @@
 # Auth0 Documentation
-This is the repository for the Auth0 documentation.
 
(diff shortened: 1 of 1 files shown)


vyaspranjal33 (changes requested):
Keep the word Project as it is , this is Auth0 docs repository not a Library , i suggest not to add badges here as this is docs repository @khorne3 can suggest more changes if needed

khorne3:
Thanks for your contribution! Unfortunately, we are not currently updating our README -- sorry!

PR #6902, opened 2018-11-01
Removed portion about enforcing namespacing when using OIDC flows

rkhoriander:
You always have to namespace custom claims as a URL now, not just when OIDC conformant flow is enabled.


ShayMe21 (changes requested):
Could you please also mention that this only applies to legacy tenants. Any new tenants created, will have to use URL namespaces as custom claims identifier. We no longer enable this feature.
Sort of like what is mentioned here in about legacy tenants being able to change Seamless SSO feature from tenant settings: https://auth0.com/docs/sso/current/setup#addendum-sso-configuration-for-legacy-tenants

Thanks!

rkhoriander:
@ShayMe21, Can you please provide info on where in the Dashboard legacy tenants will see different settings. I don't see this on my side, and the Slack conversation suggests that all flags set, etc. have to be done on the back-end now. What relevant flags no longer exist for a new customer that did for a legacy customer?

ShayMe21:
@rkhoriander Sorry for the late reply, I was OOF. Legacy tenants will have a feature flag enabled in which if they use non-OIDC flows, they can use non-URL custom claims. I am not sure what the name of this feature flag is. @nicosabena do you mind reviewing this PR, where we need to include the above detail about legacy tenants able to use non-URL custom claims or not? Thanks!

nicosabena:
Auth0 will allow non-OIDC claims without a namespace (a.k.a. the "legacy" user profile) if:
  - You are using the non-OIDC conformant pipeline (i.e. you are not using the `audience` parameter in the `/authorize` or token request and the application does not have the **OIDC-Conformant** toggle enabled).
  - You have the **Legacy User Profile** toggle turned on in the [tenant Advanced Settings](https://manage.auth0.com/#/tenant/advanced), under the **Migrations** section. This setting is only enabled for old tenants, but newly created tenants can't see or enable the **Legacy User Profile**.

We strongly recommend moving away from the Legacy User Profile.

We should include the above information for completeness.

ShayMe21 (dismissed):
LGTM, added a link to OIDC application guide.

rkhoriander:
https://docs-content-staging-pr-6902.herokuapp.com/docs//api-auth/tutorials/adoption/scope-custom-claims

khorne3 (approved):

PR #6924, opened 2018-11-07
Extra info about mapping a claim to `User.Identity.Name`

leemcmullen:
Add some information about mapping a claim to `User.Identity.Name` if it's null. Similar to what appears here: https://auth0.com/docs/quickstart/webapp/aspnet-core/02-user-profile


cocojoe:
@damieng may be outdated now. However, a quick check and then we can close or merge.

damieng (approved):

PR #6944, opened 2018-11-14
How to clear Auth0 session in node web js app

kamepo:
This is often asked in the community and through support

### DIFF
--- a/articles/quickstart/webapp/nodejs/01-login.md
+++ b/articles/quickstart/webapp/nodejs/01-login.md
@@ -170,6 +170,20 @@ module.exports = router;
 This tutorial implements logout by closing the local user session. After logging out, the user's session in the Auth0 authentication server is still open. For other implementations, please refer to the [logout documentation](/logout). 
 :::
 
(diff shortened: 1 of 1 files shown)


albertoperdomo:
Hello @kamepo,

we plan to improve the Quickstart to remove Auth0 session, so thank you for submitting this.
As to this implementation, I think that session should also be cleared.

I am not 100% sure about what the best method to do this would be, but redirecting to `/logout`would probably do the trick?

Let me ask about this and get back to you.

albertoperdomo:
After discussing this with others it seems that a better way to implement this would be:

- Keep the logout link pointing to the local `/logout` route
- In the authentication router, after clearing the local session (`req.logout()`), redirect to the Auht0 logout endpoint instead of `/`.
- The auth0 logout endpoint will then redirect the user to the application's root path.

Would you be willing to change your PR to follow this implementation path?

Thanks.

kamepo (commented):
updated with suggestions

cocojoe:
I don't think this is relevant now, has been superseded with Express QS. Please re-open if this is not the case. Thanks

PR #6976, opened 2018-11-29
The link is showing as  https://localhost:8989/#/account/webtasks for…

kamepo:
… Auth0 Management Dashboard's Webtask page

https://manage.auth0.com/#/tenant/webtasks


jeffreylees:
Hey - thanks for the report - this is actually a symptom of another problem, lots of documents use that manage url variable. We are working on it! Thanks.

PR #6980, opened 2018-11-30
Rework Anomaly Detection content

tamigoodall:
Chunked content


tamigoodall:
Here's the link to the top level of the content in the review app: https://auth0-docs-content-pr-6980.herokuapp.com/docs/anomaly-detection/concepts/overview-anomaly-detection

tamigoodall:
I still think there might be a breadcrumb problem. Any help to figure it out would be appreciated.

erjohnson:
For the breadcrumbs, the pages in the section would have to be children of the main anomaly detection page I think

tamigoodall:
Fixed breadcrumb issue.

articles/anomaly-detection/concepts/breached-passwords.md
```diff
@@ -0,0 +1,58 @@
...
+    - security
+    - passwords
+contentType: concept
+useCase: customize-anomaly-detection
+v2: true
+---
+
+# Breached Password Security
+
+When a user receives an email requesting that they change their password immediately, it is because their account could be the victim of a security breach. This may be the result of a compromise by a third-party application that experienced a security breach. The breach may not have happened to this account, but based on avaiable data, the user's credentials may have been released. Since many people reuse passwords, the request to change passwords is a precaution to make sure the user stays protected.
```
jeffreylees: `avaiable` -> `available`

articles/anomaly-detection/concepts/breached-passwords.md
```diff
@@ -0,0 +1,58 @@
...
+
+# Breached Password Security
+
+When a user receives an email requesting that they change their password immediately, it is because their account could be the victim of a security breach. This may be the result of a compromise by a third-party application that experienced a security breach. The breach may not have happened to this account, but based on avaiable data, the user's credentials may have been released. Since many people reuse passwords, the request to change passwords is a precaution to make sure the user stays protected. 
+
+Users may also want to change their password at any other sites where they suspect they used a shared password. 
+
+## General security tips
+
+Users typically can't prevent certain sites from security breaches, but there are some things they can do to help keep their accounts safe.
```
jeffreylees: I'm unsure what this first portion means. Perhaps "Security breaches on web applications are usually out of the end-user's hands, but there are some things they can do to keep their accounts as safe as possible." ?

articles/anomaly-detection/concepts/breached-passwords.md
```diff
@@ -0,0 +1,58 @@
...
+Always do a password reset through the actual site itself not via potentially false links in emails. Also note that secure website URL always starts with `https`.
+
+Here are some links for password resets on commonly used sites:
+* [Google](https://www.google.com/accounts/recovery/)
+* [Facebook](https://www.facebook.com/settings)
+* [Twitter](https://twitter.com/settings/password)
+
+### Never enter personal or financial information in email
+
+Emails in general are not very secure and re not a good way to communicate sensitive information. A trusted company/application would not ask for  information in this way. Make sure not to enter confidential information through false links in emails.
```
jeffreylees: `re` -> `are`

articles/anomaly-detection/concepts/breached-passwords.md
```diff
@@ -0,0 +1,58 @@
...
+
+Emails in general are not very secure and re not a good way to communicate sensitive information. A trusted company/application would not ask for  information in this way. Make sure not to enter confidential information through false links in emails. 
+
+### Never download files from unreliable sources
+
+Most web browsers detect suspicious sites. An alert should appear when you try to access a malicious site. Never download files from suspicious emails or websites.
+
+### Do not reuse passwords
+
+When one site has a breach of user data, if a user uses the same credentials elsewhere, information in other sites can also be accessed. There are many password tools available to help keep track of passwords.
```
jeffreylees: Maybe a line or two of clarification here -- something like this: 

"...accessed. The only way to prevent this is by not reusing passwords for multiple sites. The problem is that remembering countless numbers of different passwords is frustrating and often impossible. One solution to this problem is the use of a password manager. There are many password managers available which can help users to use separate and secure passwords for each account, but at the same time not be responsible for remembering all of them."

articles/anomaly-detection/concepts/breached-passwords.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Use strong passwords
+
+The longer a password is, the harder it becomes to guess. Try to make passwords long and use a mix of special characters, numbers, and upper- and lowercase letters.
```
jeffreylees: The longer a password is, the harder it becomes to be guessed via brute force methods. Many sites allow the use of pass-phrases (a phrase or sentence instead of just a complicated word). Try to make passwords long and use a mix of special characters, numbers, and upper- and lowercase letters.

articles/anomaly-detection/guides/set-anomaly-detection-preferences.md
```diff
@@ -0,0 +1,61 @@
...
+---
+# Set Anomaly Detection Preferences
+
+Customize the actions that occur after the triggers in the **Anomaly Detection** section on the [Dashboard](${manage_url}/#/anomaly).
+
+::: warning
+Auth0 recommends that you **do not** make changes to your anomaly detection features with the Management API.
+:::
+
+![](/media/articles/anomaly-detection/anomaly-detection-overview.png)
```
jeffreylees: Should put alt tags between the `[ ]` on this and the other images

articles/anomaly-detection/guides/customize-blocked-account-emails.md
```diff
@@ -0,0 +1,29 @@
...
+Auth0 never blocks the user itself, just the attempts from the suspicious origin.
+:::
+
+The email sent to the user looks like this:
+
+![Email Example](/media/articles/brute-force-protection/bfp-2015-12-29_1832.png)
+
+You can customize the template used for this message on the [Dashboard](${manage_url}/#/emails) under __Emails > Templates > Blocked Account Email__.
+
+## Keep reading
```
jeffreylees, line 27: Most of the Keep Reading headings in these docs are missing the line return after before starting the reading list.

articles/anomaly-detection/references/brute-force-protection-triggers-actions.md
```diff
@@ -0,0 +1,61 @@
...
+### Actions
+
+* Notify dashboard administrator(s).
+* Block suspicious addresses for 15 minutes.
+
+If this block is triggered, additional access attempts are released one-at-a-time over the course of 24 hours until 100 attempts are allocated. This results in approximately 1 additional attempt every 15 minutes.
+
+### Remove block
+
+Auth0 emails the dashboard administrator(s) when this block is triggered. Within this email there's a link the owner can click on to clear the block.
```
jeffreylees, line 54: Is this the only way to remove this particular block? I would've assumed the dashboard could be used too, rather than only an email link. Maybe this is a question for someone else / later, just curious.

articles/compliance/gdpr/features-aiding-compliance/protect-user-data.md
```diff
@@ -26,16 +26,28 @@ There are several Auth0 features than can help you achieve that, like user profi
...
+
+Every time Auth0 detects 100 failed login attempts in 24 hours or 50 sign up attempts from the same IP address, we will:
+
+- Notify dashboard administrator(s).
+- Block suspicious addresses for 15 minutes.
 
 You can enable brute-force protection, configure which actions you want to take, and customize the blocked account email using the Dashboard. 
 
-For more information and configuration steps, see [Brute Force Protection](/anomaly-detection#brute-force-protection).
+For more information, see [Brute Force Protection](/anomaly-detection/index#brute-force-protection).
```
jeffreylees: Just a note, if the URL here (and the same comment for others in the PR) are to the index page, probably `/anomaly-detection#brute-force-protection` would work?

jeffreylees (changes requested):
I didn't check every link - I might do a pass later - but here are a few comments!

articles/anomaly-detection/concepts/breached-passwords.md
tamigoodall: I think it is meant to convey that end-users don't have much control over security breaches that occur at companies they have accounts with. For example, a breach at Uber wouldn't necessarily be preventable by an Uber user.

tamigoodall: I'll reword it to make it more clear.

articles/anomaly-detection/references/brute-force-protection-triggers-actions.md
tamigoodall: Good question. I'll ask around.

jeffreylees (approved):

PR #6988, opened 2018-12-03
Update login for React quickstart to React Router 4

ButterBridge:
There was some incorrect code towards the end of this file describing how to use a callback function, which used Route's render props facility to call a '`handleAuthentication`' method which tests the hash and in turn call the `auth.handleAuthentication` method. This appeared to be an artefact from React Router 3's implementation of `onEnter`, though I'm not sure. In any case, the function would not have been called with the has argument as described, although the `auth.handleAuthentication` function would have been unaffected. I have rewritten it to pass the testing function - which I have renamed `testAuthHash` to distinguish it from the auth version - to the `Callback` component, where it is called on `componentDidMount`, which is the recommended way to handle this sort of thing in React Router 4.

### DIFF
--- a/articles/quickstart/spa/react/_includes/_centralized_login.md
+++ b/articles/quickstart/spa/react/_includes/_centralized_login.md
@@ -161,6 +161,10 @@ import loading from './loading.svg';
 class Callback extends Component {
   render() {
     const style = //...
+    
+    componentDidMount () {
+      this.props.testAuthHash(this.props.location.hash);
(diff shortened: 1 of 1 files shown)


albertoperdomo:
Thanks for your submission @ButterBridge.

After careful consideration, we have decided to not accept this PR for following reasons:

• Rendering a component to process the hash
• We are working on a new implementation which will supersede this one.

PR #6990, opened 2018-12-04
Decode json before load jwks

YannBeauxis:
I've had the TypeError: the JSON object must be str, not 'bytes'
at jwks = json.loads(jsonurl.read())


lbalmaceda:
I know it's old. I just tried and couldn't reproduce it.

PR #7030, opened 2018-12-11
Adds example spf record to email docs

dmark:
Adds an example SPF record to the Customizing Your Emails docs.


dmark:
The CloudWatch doc was not meant to be part of this PR. I'll withdraw it.

PR #7071, opened 2018-12-21
Update scopes docs

rkhoriander:
Added more general examples.
Moved content into more logical groupings.
Edited for clarity and wordiness.
Added relevant cross-links.
Fixed referring links.


articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,37 @@
...
 
-## Background
+For example, let's say you are building an API that provides data to a calendar application. You want some users to be able to edit items on the calendar, others to only be able to read them, and others to be able to both read and write to calendar items. To do this, you create two scopes for your API: one that authorizes write access (`write:appointments`) and one that authorizes read-only access (`read:appointments`). 
 
-When you [create an API in Auth0](/apis), you can use scopes to control how and what someone can access. For example, if you want users to be able to **read** and **delete** contact information, you would create the follow two scopes to reflect these two actions:
+Now, when an app calls your API, it will specify the scope it needs in its request. The app may request read access by including `read:appointments` in its scope, write access by including `write:appointments` in its scope, or both read and write access by including both `read:appointments` and `write:appointments` in its scope.
```
jeffreylees: So, each of those two items are scopes; so it's not that the app is including those items in its scope, but rather its request includes the `write:appointments` and `read:appointments` scopes.

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,37 @@
...
-  response_type=id_token%20token&
-  client_id=${account.clientId}&
-  redirect_uri=${account.callback}&
-  nonce=YOUR_CRYPTOGRAPHIC_NONCE
-  state=YOUR_OPAQUE_VALUE
-```
+## Limit API scopes
 
-Note the differences between the example above and the example on asking for standard claims. In the example above, we want to get an Access Token, that will allow us to access the API, with the rights to do specific actions. To do so, we changed two parameters and added a new one:
+An application can request any defined scope and the user will be prompted to approve that scope during the authorization flow. Instead, you may want to limit available scopes based on, for example, the role of a user. To do this, you can use the [Authorization Extension](/extensions/authorization-extension) and a custom [Rule](/rules).
```
jeffreylees: This seems to imply that you would be limiting the scopes available to an application based on roles, which isn't quite accurate.

Maybe something like this: `An application can include any defined scope in its request. Instead of including _all_ available scopes in every request, however, the scopes can be limited. For example, a user of your application can be given a role so that requests on their behalf are limited to just the scopes assigned to that role. To do this, you...`

articles/scopes/current/index.md
```diff
@@ -1,23 +1,47 @@
...
 useCase:
   - development
+  - add-login
+  - secure-api
+  - call-api
 ---
 # Scopes
 
-The OAuth 2.0 protocol is a delegated authorization mechanism, where an application requests access to resources controlled by the user (the resource owner) and hosted by an API (the resource server). The authorization server issues the application a more restricted set of credentials than those of the user in the form of an Access Token.
+Scopes let you control the type of access your users need; they are a method of limiting the access given to a user by a [token](/tokens).
```
jeffreylees: I'm not sure I'd say "by a token" here.

`Scopes are a method by which you can control the specific types of access to resources that are granted to the application on the user's behalf.` or if you want to mention tokens, `Scopes are a method by which you can control the specific types of access a user's token will grant.` or something maybe? The token isn't the thing causing the limitation (although it is the tool by which they gain access).

articles/scopes/current/index.md
```diff
@@ -1,23 +1,47 @@
...
+When an app requests access to a resource through an authorization server, it uses the `scope` parameter to specify what type of access it desires, and the authorization server uses the `scope` parameter to respond with the type of access that was actually granted (if the granted access level was different from what was requested).
 
-You can use scopes to:
+Generally, you use scopes in two ways:
 
-- Let an application verify the identity of a user (by using [Open ID Connect](/protocols/oidc)) and get basic profile information about the user, such as their email or picture. For details, refer to [OpenID Connect Scopes](/scopes/current/oidc-scopes).
+* In an [application](/applications), to verify the identity of a user and get basic profile information about the user, such as their email or picture. In this scenario, the scopes available to you include those implemented by the [OpenID Connect](/protocols/oidc) protocol. For details, refer to [OpenID Connect Scopes](/scopes/current/oidc-scopes).
```
jeffreylees: So, the slight wording change at the start makes it sound like this use case occurs within an application. You're using OIDC, but you're still querying an API endpoint from the application, and then getting back standard claims with user info.

articles/scopes/current/index.md
```diff
@@ -1,23 +1,47 @@
...
 
-- Implement granular access control to your API. In this case, you need to define [custom scopes](/scopes/current/api-scopes) for your API and add these newly-created scopes to your `scope` request parameter: `scope=read:contacts`.
+* In an [API](/apis), to implement granular access control. In this case, you need to define custom scopes for your API and then identify these scopes so that calling applications can use them. For details, refer to [API Scopes](/scopes/current/api-scopes).
+
+
+## Best practices
+
+Understand your use case and choose the most restrictive scope possible. Are you establishing a user's identity or asking the user to allow you to interact with their data? There's a big difference between importing a user's Facebook profile information and posting to their wall. Only request what you absolutely need. By doing so, you are also more likely to gain user consent since users are more likely to grant access for limited, clearly-specified scopes.
+
```
jeffreylees, line 39: Should be single returns between sections

articles/scopes/current/index.md
```diff
@@ -1,23 +1,47 @@
...
+
+## Requested scopes versus granted scopes
+
+Remember that a user gets to consent to the access level you are requesting. While usually the scopes returned will be identical to the scopes you requested, users can edit their scopes (both during and after initial consent), thereby granting your app less access than you requested.
```
jeffreylees: This is a question, not a critique - but is this true? `...users can edit their scopes (both during and after initial consent)...`

I would assume that this would depend entirely on the application you had built? Surely in some cases they just have to approve (or deny) the application's request to access (whatever scopes)? WDYT?

jeffreylees (commented):
This is not an exhaustive review, and I'm not a scopes / authz expert, so feel free to discard suggestions. Just a few comments because I saw the PR.

articles/scopes/current/index.md
rkhoriander: I can actually authenticate using google, authorizing an app for certain scopes and then go back into Google and remove some of those scopes. I can do the same in Githhub. This is what I was thinking about--the fact that at any time, I could go back and be like, "NO, app, you are no longer allowed to look at my repos." So it's not really dependent on YOUR application. But it is dependent on the resource you're requesting access to. I will rephrase to make that clearer.

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
+Often, new applications will want to make use of the information that has already been created in an online resource. To do so, the application must ask for authorization to access this information on a user's behalf. Permissions (also known as _scopes_ in the OAuth 2.0 specs) define the specific actions third-party applications can be allowed to do on a user's behalf.
 
-You can use scopes to:
+## Ways to use permissions
 
-- Let an application verify the identity of a user (by using [Open ID Connect](/protocols/oidc)) and get basic profile information about the user, such as their email or picture. For details, refer to [OpenID Connect Scopes](/scopes/current/oidc-scopes).
+When an app requests permission to access a resource through an authorization server, it uses the `scope` parameter to specify what type of permission it needs, and the authorization server uses the `scope` parameter to respond with the type of permission that was actually granted (if the granted permission was different from what was requested).
```
twistedstream: The phrase "type of permission" here is a bit confusing. I think it should just be "permissions" as "type of" infers there's some higher level generalization, but there's not. For example, a type of permission might be _permissions for reading_ and another might be _permissions for writing_. But the `scope` parameter can contain any mix of "types" of permissions (eg. `scope=read:foo write:foo`).

```suggestion
When an app requests permission to access a resource through an authorization server, it uses the `scope` parameter to specify which permissions it needs, and the authorization server uses the `scope` parameter to respond with the permissions that were actually granted (if the granted permission was different from what was requested).
```

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
+
+Understand your use case and choose the most restrictive permissions possible. 
+
+If you are requesting permission, make sure you ask for enough access for your application to function, but only request what you absolutely need. Are you establishing a user's identity or asking the user to allow you to interact with their data? There's a big difference between importing a user's Facebook profile information and posting to their wall. By only requesting what you need, you are more likely to gain user consent since users are more likely to grant access for limited, clearly-specified permissions. 
+
+Similarly, when creating custom permissions for an API, consider what levels of granular access applications may need and design accordingly.
+
+## Requested scopes versus granted scopes
+
+Remember that a user gets to consent to the permissions being requested. While usually the permissions returned will be identical to those requested, users can edit granted permissions (both during initial consent and sometimes after, depending on the resource), thereby granting an app less access than it requested.
```
twistedstream: >Remember that a user gets to consent to the permissions being requested

This is really only the case with 3rd-party clients. If a 1st-party client is performing a call to `/authorize`, the user is never prompted (with an OAuth Consent page) to approve all the scopes being requested by the client. We should probably make this distinction here.

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
   - concept
   - index
 useCase:
   - development
+  - add-login
+  - secure-api
+  - call-api
 ---
-# Scopes
+# Permissions
```
twistedstream: I'm somewhat troubled by the almost blanket replacement of `scope` with `permission` in this doc page. I feel like we're conflating two separate concepts (see [Vittorio's blog post](https://auth0.com/blog/on-the-nature-of-oauth2-scopes/) for more details). 

Permissions in this conversation are really _user permissions_. They answer the question: 
_What can this user do with resource X?_

Whereas scopes are more of an OAuth construct designed to address _delegated access_. They attempt to answer the question: 
_What has the user allowed this application to do on its behalf with resource X?_

I feel like we really need two separate docs pages to address these concepts separately; otherwise, we're inviting a bit of a confusion mess.

I realize that we're potentially going to release an option with the new AuthZ feature where a customer could treat scopes as if they were permissions - a sort of short cut for the purposes of simplification. But I don't think that means we merge the concepts in the docs. We still need separate discussions and we need to provide proper guidance as to which option a customer should choose when modeling their API authorization.

articles/scopes/current/oidc-scopes.md
```diff
@@ -1,74 +1,58 @@
...
+  - permissions
 contentType:
-  - how-to
+  - concept
 useCase:
   - development
+  - add-login
 ---
-# OpenID Connect Scopes
+# OpenID Connect Permissions
```
twistedstream: Overall I think this doc page has received some nice updates and examples. But per my comments on the `/docs/scopes/current` docs page, I don't think introducing the term "permission" on this page is helpful. OIDC calls them "scopes" and not "permissions". I think this supports the notion that we need to treat scopes and permissions separately, not attempt to combine them.

articles/scopes/current/oidc-scopes.md
```diff
@@ -1,74 +1,58 @@
...
+:::
 
-The basic (and required) scope for OpenID Connect is the `openid` scope. This scope represents the intent of the application to use the OIDC protocol to verify the identity of the user.
+OpenID Connect (OIDC) permissions are used by an application during authentication to authorize access to a user's details, like name and picture. They are known as _scopes_ in the OIDC specification. Each scope returns a set of user attributes, which are called _claims_. The scopes an application should request depend on which user attributes the application needs. Once the user authorizes the requested scopes, the claims are returned in an ID Token and are also available through the [/userinfo endpoint](/api/authentication#get-user-info).
 
-In OpenID Connect (OIDC), we have the notion of __claims__. There are two types of claims:
+For example, let's say you have built a regular web application, registered it with Auth0, and have configured it to allow a user to log in using Google. Once a user logs in to your app, you want to auto-generate and send a personalized welcome email, including the user's name.
```
twistedstream: This is an interesting example, but it may be a bit confusing since it's mixing the collection of user attributes from Google with those stored by Auth0. The scopes that the client is requesting from Auth0 are specific to the Auth0 user profile, _regardless_ of how they were sourced. For example, the user could have chosen a different way to authenticate instead of Google (eg. sign in with un/pw). I'm not saying it's a bad example, but perhaps we need to be clear on the mechanics here so users don't think the prompt they receive from Google after they sign in is impacted by the OIDC scopes requested by the client. Instead they are impacted by the `scopes` property in the `google-oauth2` connection.

articles/scopes/current/index.md
jhiner: > and the authorization server uses the `scope` parameter

This is true, but in our case we might actually use different/additional claims. We'll have to update the doc if we change this during beta.

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
 
-- Implement granular access control to your API. In this case, you need to define [custom scopes](/scopes/current/api-scopes) for your API and add these newly-created scopes to your `scope` request parameter: `scope=read:contacts`.
+Generally, you use permissions in three ways:
+
+* From an [application](/applications), to verify the identity of a user and get basic profile information about the user, such as their email or picture. In this scenario, the permissions available to you include those implemented by the [OpenID Connect](/protocols/oidc) protocol. For details, refer to [OpenID Connect Permissions](/scopes/current/oidc-scopes).
+
+* In an [API](/apis), to implement granular access control. In this case, you need to define custom permissions for your API and then identify these permissions so that calling applications can use them. For details, refer to [API Permissions](/scopes/current/api-scopes).
```
jhiner: I would omit `granular`, since it is unclear and up to the API implementer how granular permissions should be (and there are appropriate levels of granularity in fact - background: https://auth0.com/blog/on-the-nature-of-oauth2-scopes/)

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
+
+## Requested scopes versus granted scopes
+
+Remember that a user gets to consent to the permissions being requested. While usually the permissions returned will be identical to those requested, users can edit granted permissions (both during initial consent and sometimes after, depending on the resource), thereby granting an app less access than it requested. 
+
+As an application developer, you should be aware of this possibility and handle these cases in your app. For example, your app could warn the user that they will see reduced functionality. It could also send the user back through the authorization flow to ask for additional permissions. But again, remember that users can always say no.
+
+## Keep reading
+
+- [OpenID Connect Permissions](/scopes/current/oidc-scopes)
```
jhiner: IMHO, I don't think we should call these `OpenID Connect Permissions` because the `scopes` terminology is embedded in the OIDC spec itself, so OpenID Connect Scopes seems to make sense in this context. But maybe for consistency we should use the same term, I don't know just expressing my 2 cents.

Note this comment applies throughout all the pages, so I won't repeat it elsewhere.

articles/scopes/current/sample-use-cases.md
```diff
@@ -0,0 +1,157 @@
...
+  "at_hash": "IoS3ZGppJKUn3Bta_LgE2A"
+}
+```
+
+Your app now can retrieve the user attributes and use them to personalize your UI.
+
+
+## Request custom API access
+
+In this example, we expand on our previous example requesting standard claims to also include a custom permission for a calendar API that will allow the calling application to read appointments for the user. To do this, we want to get an additional token--an Access Token containing the proper permission level to read appointments from the API.
```
jhiner: I think this is probably already on the docs team radar, but we are starting to advise against the implicit flow. https://auth0.com/blog/oauth2-implicit-grant-and-spa/ So perhaps this sample needs to be updated?

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
+Different pieces of user information are often stored across a number of online resources. Users may upload and store photos with a service like Flickr, keep digital files on Dropbox, and store contacts and events in Google Calendar or on Facebook.
 
-The permissions represented by the Access Token in OAuth 2.0 terms are known as scopes. The `scope` parameter allows the application to express the desired scope of the access request. The `scope` parameter can also be used by the authorization server in the response to indicate which scopes were actually granted (if they are different than the ones requested).
+Often, new applications will want to make use of the information that has already been created in an online resource. To do so, the application must ask for authorization to access this information on a user's behalf. Permissions (also known as _scopes_ in the OAuth 2.0 specs) define the specific actions third-party applications can be allowed to do on a user's behalf.
```
jhiner: `in the OAuth 2.0 specs`

Maybe link to the spec?

articles/scopes/current/index.md
```diff
@@ -1,23 +1,55 @@
...
+
+## Best practices
+
+Understand your use case and choose the most restrictive permissions possible. 
+
+If you are requesting permission, make sure you ask for enough access for your application to function, but only request what you absolutely need. Are you establishing a user's identity or asking the user to allow you to interact with their data? There's a big difference between importing a user's Facebook profile information and posting to their wall. By only requesting what you need, you are more likely to gain user consent since users are more likely to grant access for limited, clearly-specified permissions. 
+
+Similarly, when creating custom permissions for an API, consider what levels of granular access applications may need and design accordingly.
+
+## Requested scopes versus granted scopes
```
jhiner, line 42: I feel like the heading should reflect the `Permission` term.

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
 
-* `read:contacts` 
-* `delete:contacts`
+A calling application will request authorization from the user to access the requested permissions, and the user will approve or deny the request. The app may request read access by including the `read:appointments` permission in its request, write access by including the `write:appointments` permission in its request, or both read and write access by including both the `read:appointments` and `write:appointments` permissions in its request. 
 
-If you wanted to expand [our example on asking for standard claims](/scopes/current/oidc-scopes#example-asking-for-standard-claims) to include also the `read:contacts` permission, then you would using something like the following sample URL to initiate the authentication flow using the Implicit grant:
+Now, when the app calls your API, it will include a token which verifies that the user has provided authorization to access their content and also indicates which permissions the user has approved. Your API should respect the approved permissions and only release information that was authorized by the user to the calling application.
```
jhiner: I think we should add here that an API can use additional logic beyond the token itself to allow access or not. APIs may choose to enforce additional access control mechanisms.

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
-  nonce=YOUR_CRYPTOGRAPHIC_NONCE
-  state=YOUR_OPAQUE_VALUE
-```
+For an example showing how to request custom API access for your application, see [Sample Use Cases: Permissions and Claims](/scopes/current/sample-use-cases#request-custom-API-access).
 
-Note the differences between the example above and the example on asking for standard claims. In the example above, we want to get an Access Token, that will allow us to access the API, with the rights to do specific actions. To do so, we changed two parameters and added a new one:
+## Limit API permissions
 
-- `audience`: New parameter added for this example. Its value is the unique identifier of the API we want to get access to.
+An application can include any defined permission level in its request. Instead of including _all_ available permissions in every request, however, the permissions can be limited. For example, a user of your application can be given a role so that requests on their behalf are limited to just the permissions assigned to that role. To do this, you can use the [Authorization Extension](/extensions/authorization-extension) and a custom [Rule](/rules).
```
jhiner: I don't follow this block. Are we simply trying to say an app should limit its request only to what it needs?

articles/api-auth/tutorials/represent-multiple-apis.md
```diff
@@ -1,20 +1,23 @@
...
+  - scopes
+  - permissions
 contentType: tutorial
 useCase:
   - secure-api
   - call-api
 ---
 
-# How to Represent Multiple APIs Using a Single Auth0 API
+# Represent Multiple APIs Using a Proxy API in Auth0
```
jhiner: I'm not sure about the term Proxy. Maybe "Logical API" is better. The intent here I think is that a customer might have multiple API implementations (let's say a microservices architecture, so a calendar service and a contact service) so there are several distinct API implementations but they are all logically a part of the same API (Organizer service).

jhiner (commented):
Left a few comments.

To answer the two questions you posed:
1) Generally yes I think the docs are technically accurate
2) They still seem useful for a more technical user I'd say.

But I did leave some comments about some potential improvements.'

Also once we release our authorization feature, I'd say we want to remove all recommendations to use the Authorization Extension from the site.

articles/scopes/current/index.md
rkhoriander: Agreed. This will be a change we make during doc updates specifically related to the Authz release for when it comes out of beta. This doc will go out prior to that, so we can't make those changes just yet.

articles/scopes/current/index.md
rkhoriander: Removed.

articles/scopes/current/index.md
rkhoriander: Linked.

articles/scopes/current/index.md
rkhoriander: Fixed.

articles/scopes/current/api-scopes.md
rkhoriander: Added note.

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
 ---
-# API Scopes
+# API Permissions
 
-API scopes allow you to define the API data accessible to your applications. 
+As an [API](/apis) developer, you need to:
 
-## Background
+1. Decide which user information you would like external applications to be able to access on a user's behalf.
+2. Define these access levels as custom permission levels.
```
twistedstream: Using the term "permission levels" is confusing as it implies some variation of the concept of a permission. I would just use "permissions".

```suggestion
2. Define these access levels as custom permissions.
```

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
+3. Identify these permission levels so that calling applications can use them.
```
twistedstream: Same reasoning as the previous comment:

```suggestion
3. Identify these permissions so that calling applications can use them. 
```

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
+  - secure-api
 ---
-# API Scopes
+# API Permissions
 
-API scopes allow you to define the API data accessible to your applications. 
+As an [API](/apis) developer, you need to:
 
-## Background
+1. Decide which user information you would like external applications to be able to access on a user's behalf.
```
twistedstream: Calling this "user information" makes it sounds like its user profile information. Rather, this is data (information) within the API that can be _managed by_ or is _associated with_ the user.

```suggestion
1. Decide which information you would like external applications to be able to access on a user's behalf.
```

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
 
-When you [create an API in Auth0](/apis), you can use scopes to control how and what someone can access. For example, if you want users to be able to **read** and **delete** contact information, you would create the follow two scopes to reflect these two actions:
+For example, let's say you are building an API that provides data to a calendar application. At various times, the app may need to edit items on the calendar, read them, or both read and write to calendar items for a user. To do this, you create two permission levels for your API: one that authorizes write access (`write:appointments`) and one that authorizes read-only access (`read:appointments`). 
 
-* `read:contacts` 
-* `delete:contacts`
+A calling application will request authorization from the user to access the requested permissions, and the user will approve or deny the request. The app may request read access by including the `read:appointments` permission in its request, write access by including the `write:appointments` permission in its request, or both read and write access by including both the `read:appointments` and `write:appointments` permissions in its request.
```
twistedstream: I've already raised this issue in the main docs page (`/scopes/current`), but this language:
>A calling application will request authorization from the user to access the requested permissions

is again why I think we need to separate the concepts of "permissions" and "scopes". Permissions are about what the user themselves has been authorized to do. What we're talking about here is delegated access, which is what scopes are for. I personally think this document should be under the "scopes" umbrella, assuming we break out both concepts.

articles/scopes/current/api-scopes.md
```diff
@@ -1,98 +1,43 @@
...
-  response_type=id_token%20token&
-  client_id=${account.clientId}&
-  redirect_uri=${account.callback}&
-  nonce=YOUR_CRYPTOGRAPHIC_NONCE
-  state=YOUR_OPAQUE_VALUE
-```
+For an example showing how to request custom API access for your application, see [Sample Use Cases: Permissions and Claims](/scopes/current/sample-use-cases#request-custom-API-access).
 
-Note the differences between the example above and the example on asking for standard claims. In the example above, we want to get an Access Token, that will allow us to access the API, with the rights to do specific actions. To do so, we changed two parameters and added a new one:
+## Limit API permissions
```
twistedstream: This section attempts to blur the lines a bit between user permissions and scopes. If you attempt to marry the two concepts, then it makes sense that a user can only delegate access to the things they themselves have been given permission to access. While I think this is a valid simplification in some use cases, I think it breaks down in more complex ones. Again, I think we need to keep these concepts separate. In other words, an API may have both scopes and permissions.

articles/scopes/current/sample-use-cases.md
```diff
@@ -0,0 +1,157 @@
...
+Notice that in this example: 
+
+* the `response_type` parameter includes one value:
+  * `id_token` (to get an ID Token)
+* the `scope` parameter includes three values: 
+  * `openid` (to indicate that the application intends to use OIDC to verify the user's identity)
+  * `profile` (to get `name`, `nickname`, and `picture`)
+  * `email` (to get `email` and `email_verified`)
+
+2. After the user consents and Auth0 redirects back to your app, extract the ID Token from the hash fragment of the URL and decode it.
```
twistedstream: I think we need to be explicit here and say that user consent only occurs for 3rd-party applications - a type of app that _very few_ of our customers are using today. The vast majority are building 1st-party apps, which skip the OAuth consent prompt during authorization.

articles/scopes/current/sample-use-cases.md
twistedstream: Sorry to beat a dead horse, but it should be:
```suggestion
In this example, we expand on our previous example requesting standard claims to also include a custom scope for a calendar API that will allow the calling application to read appointments for the user. To do this, we want to get an additional token--an Access Token containing the proper scope to read appointments from the API.
```

articles/scopes/current/sample-use-cases.md
```diff
@@ -0,0 +1,157 @@
...
+  nonce=YOUR_CRYPTOGRAPHIC_NONCE
+  state=YOUR_OPAQUE_VALUE
+```
+
+Notice that in this example:
+
+* the `response_type` parameter now includes two values:
+  * `id_token` (to get an ID Token)
+  * `token` (to get an Access Token)
+* the `scope` parameter is used for both OpenID scopes and API permissions, so now includes four values:
```
twistedstream:
```suggestion
* the `scope` parameter is used for both OIDC scopes and our custom API scopes, so now includes four values: 
```

articles/api-auth/tutorials/represent-multiple-apis.md
```diff
@@ -23,92 +26,101 @@ The sample application contains:
...
-![](/media/articles/api-auth/tutorials/represent-multiple-apis/scopes-page.png)
+Permissions allow you to define which API actions will be accessible to calling applications. One permission (or scope) will represent one API/action combination. 
 
-Scopes allow you to define the API data accessible to your applications. You'll need one scope for each API represented and action. For example, if you want to `read` and `delete` from an API called `samples`, you'll need to create the following scopes:
+For example, if you want calling applications to be able to `read` and/or `delete` from one API called `samples` and another one called `examples`, you would need to create the following permissions:
 
 * `read:samples`
 * `delete:samples`
+* `read:examples`
+* `delete: examples`
```
twistedstream: There shouldn't be a space in the scope names:
```suggestion
* `delete:examples`
```

articles/api-auth/tutorials/represent-multiple-apis.md
twistedstream: In general I like the approach that this docs page is providing, but I'm not sure I agree with the use of the term "proxy". When it comes to API's, proxies usually refer to actual services that sit in front of downstream services (eg. an Nginx reverse-proxy in front of the Contacts and Calendar services). But we're just taking two _separate_ APIs and logically _combining_ them in the Auth0 configuration in order to simplify the authorization process. We're not telling customers to create an actual proxy service (although they certainly could). Personally, I think the term "single logical API" more clearly communicates the idea.

```suggestion
# Represent Multiple APIs Using a Single Logical API in Auth0
```

articles/api-auth/restrict-requests-for-scopes.md
```diff
@@ -1,22 +1,30 @@
...
 useCase:
   - secure-api
   - call-api
 ---
 
-# Restrict Application or User Requests for API Scopes
+# Restrict Application or User Requests for API Permissions
 
-By default, any user associated with an [Auth0 application](/applications) can request an API's [scope(s)](/scopes#api-scopes). If you would like to restrict access to the API's scopes based on the user's role, application association, location, and so on, you can do so via [rules](/rules). Then, if a restricted user attempts to request scopes not permitted to them, they will receive an `HTTP 401` response.
+By default, any user associated with an [Auth0 application](/applications) can request any [custom API permissions] (/scopes/current/api-scopes) that have been created. Sometimes you may not want to allow an application to request certain permissions, though. For example, you may want to restrict access to an API's permissions based on the calling application or a user's role or location. To do so, we use [rules](/rules).
```
twistedstream: Just fixing a broken MD link:

```suggestion
By default, any user associated with an [Auth0 application](/applications) can request any [custom API permissions](/scopes/current/api-scopes) that have been created. Sometimes you may not want to allow an application to request certain permissions, though. For example, you may want to restrict access to an API's permissions based on the calling application or a user's role or location. To do so, we use [rules](/rules).
```

articles/scopes/current/guides/customize-consent-prompt.md
```diff
@@ -0,0 +1,73 @@
...
+  - apis
+contentType:
+  - how-to
+useCase:
+  - development
+  - secure-api
+---
+# How to Customize the Consent Prompt
+
+When users are asked to authorize permissions, they see a consent prompt. By default, this prompt uses the permission **name** to generate text and groups all permissions for a resource, displaying the resource's actions in a single line.
```
twistedstream: >they see a consent prompt

Only for 3rd-party applications. We should maybe be more explicit about this behavior.

twistedstream (commented):
As I discuss in more detail within specific comments, I think the main theme of my PR review is that we shouldn't conflate the concepts of "scopes" and "permissions", nor should we replace "scopes" with "permissions". They are separate concepts and deserve separate discussions.

articles/scopes/current/index.md
rkhoriander: Reverted the terminology for this release as further review and convos have revealed a need to more thoroughly define when to use "scopes" and when to use "permissions". Spoke with Steve and agreed that terminology changes would be held for release with additional Authz docs changes.

articles/scopes/current/index.md
rkhoriander: Reverted the terminology for this release as further review and convos have revealed a need to more thoroughly define when to use "scopes" and when to use "permissions". Spoke with Steve and agreed that terminology changes would be held for release with additional Authz docs changes.

articles/scopes/current/oidc-scopes.md
rkhoriander: Removed Google from the example and replaced with user/password so as to make this less complicated since it's supposed to be a basic example.

jeffreylees (approved):

PR #7092, opened 2018-12-30
Add "email" to scope

ghost:
If you don't add "email" to the scope, email is never sent to application and has a value of null.


ghost:
If you download the sample project and open Startup.cs, you'll notice that it also adds "email" to the scope despite what the documentation shows.

albertoperdomo (approved):

albertoperdomo:
Thanks for the catch @andrewliang !

PR #7106, opened 2019-01-03
Fix broken links for Management API

rkhoriander:


rkhoriander:
Originally, thought there was more than one link broken, but there is just one:
https://docs-content-staging-pr-7106.herokuapp.com/docs/api/management/v2/get-access-tokens-for-production

khorne3 (approved):

PR #7125, opened 2019-01-09
Symfony example should not depend on Dotenv and AUTH0_DOMAIN env var

pitpit:

### DIFF
--- a/articles/quickstart/webapp/symfony/01-login.md
+++ b/articles/quickstart/webapp/symfony/01-login.md
@@ -66,8 +66,6 @@ Add this to your `src/AppBundle/Auth0ResourceOwner.php`
 
 namespace AppBundle;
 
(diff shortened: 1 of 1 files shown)


joshcanhelp:
Not enough info to make a decision here.

pitpit:
what info do you expect despite of the title "Symfony example should not depend on Dotenv and AUTH0_DOMAIN env var" and the content of the PR itself ?

Should I point out that Dotenv is not natively installed in SF which makes this example useless ?

joshcanhelp:
@pitpit - Without a description, I'm not sure what this is solving. Your comment helps with the context and that would have been good to have in the PR description. 

Either way, it looks like this quickstart has since been re-written and so I'm not seeing `Dotenv` showing up anywhere:

https://auth0.com/docs/quickstart/webapp/symfony

PR #7158, opened 2019-01-18
Update On the Quickstart Backend API for Spring

vladimirfomene:
I have made an update to the Quickstart for backend API with Spring Security. I changed the download instructions because they were initially Gradle instructions while the sample is using Maven. I also updated the sample, `https://github.com/auth0-samples/auth0-spring-security-api-sample/pull/35`. I updated the article to use Spring Security OAuth for resource server by changing the `AppConfig.java` file and the `auth0.properties.example` file.


joshcanhelp:
Appreciate the contribution but we've re-written both of these since your PR was submitted.

vladimirfomene:
Great!

PR #7159, opened 2019-01-18
Fixed email_verified authorization callback

mmogy:
Also applied spacing consistency on best practices article


articles/best-practices/rules.md
```diff
@@ -209,10 +209,11 @@ Whenever granting authorization based on an email address, always start by check
 
 ```js
-    // Access should only be granted to verified users.
-    if (!user.email || !user.email_verified) {
-        return callback(null, user, context);
-    }
+  // Access should only be granted to verified users.
+  if (!user.email || !user.email_verified) {
+    return callback(new UnauthorizedError('Access denied.'));
```
mmogy, line 214: This was the important change.

khorne3 (approved):

PR #7178, opened 2019-01-23
Re-write for the Laravel API quickstart

joshcanhelp:
The Laravel API quickstart is difficult to follow, missing key steps, and includes unnecessary information about user authentication. This is a ground-up re-write of the steps and code samples. 

**Notes to reviewer:**

This adds an intro section and one on obtaining tokens that I think these would be helpful on all the API quickstarts. If so, then we should parse them out into an include and apply in a separate PR.

### DIFF
--- a/articles/quickstart/backend/laravel/01-authorization.md
+++ b/articles/quickstart/backend/laravel/01-authorization.md
@@ -21,7 +21,7 @@ useCase: quickstart
 
 ### Install dependencies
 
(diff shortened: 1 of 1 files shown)


joshcanhelp:
Switching labels ... thought of a use case that we should support and want to include it here. Will fix this up and re-label.

joshcanhelp:
@albertoperdomo - Ready for you!

albertoperdomo (approved):

joshcanhelp:
@erjohnson - Ready for final review!

PR #7184, opened 2019-01-25
The path to remove a user from a role is incorrect

kaymaylove:
Updated to /users/{user_id}/roles from /users/{role_id}/roles


kaymaylove (commented):
Updating to include additional change to docs

khorne3 (approved):

PR #7199, opened 2019-01-29
Fixing broken note inside of logs docs

machuga:
We tried to use a ```::: info``` block, assuming it was acting like log levels, but clarified with Docs team that it needs to be ``` ::: note```. This has now been corrected on the Logs docs page.


machuga:
@jeffreylees I'm submitting this for review to fix that broken note section.

jeffreylees (approved):

PR #7205, opened 2019-01-30
Update required external connectivity with SMTP

bianchidotdev:
Very few Appliance customers block outbound access, but some of them do. One customer recently ran into an issue where they weren't able to send emails because the appliance wasn't able to communicate with their SMTP server. They pointed out that SMTP was not on the listed of required connectivity, though it has been required for as long as the Appliance has existed.

### DIFF
--- a/articles/appliance/infrastructure/ip-domain-port-list.md
+++ b/articles/appliance/infrastructure/ip-domain-port-list.md
@@ -197,7 +197,7 @@ Auth0 strives to keep these IP addresses stable, though this is not a given. Fro
     <td>SMTP</td>
     <td>Outbound</td>
     <td>SMTP Server(s)</td>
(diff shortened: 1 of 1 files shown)


sabues:
we might want to double check support for port `465` based on https://auth0.slack.com/archives/C1RC8M1MY/p1517368442000139

khorne3:
@michaeldbianchi The **Required** column for the row you added says **No**. Should that say yes instead?

bianchidotdev:
@khorne3  I don't think so. If a customer doesn't want to allow SMTP access, and thus block all emails, that's within their rights. I think we just need to state that it's require for email to work, but it's an ancillary function of the appliance.

khorne3 (approved):

PR #7212, opened 2019-02-01
PS-277 update authz scope validation rules

gkwang:
1) remove client check as the check provides no usefulness and may create
additional security issue when misconfigured.

2) update filteredScope logic as old logic will cause additional spaces in scopes that
does not contain ":"

3) Remove push.apply logic. We should only grant permissions that the client asks for.

4) Use arrays instead of strings for context.accessToken.scope


ubenzer (approved):
LGTM

tamigoodall (approved):

PR #7213, opened 2019-02-01
Change where to put Auth0 properties

KingIdee:
Putting your Auth0 configurations in `auth0.properties` does not work. I work for if you put it in `application.properties` hence the update in instruction.

### DIFF
--- a/articles/quickstart/backend/java-spring-security/01-authorization.md
+++ b/articles/quickstart/backend/java-spring-security/01-authorization.md
@@ -22,7 +22,7 @@ useCase: quickstart
 
 ## Configure the Sample Project
 
-The sample project has a `/src/main/resources/auth0.properties` file which configures it to use the correct Auth0 **Domain** and **API Identifier** for your API. If you download the code from this page it will be automatically filled. If you use the example from Github, you will need to fill it yourself.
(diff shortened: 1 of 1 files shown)


joshcanhelp:
@KingIdee - Apologies for the delay here!

@jimmyjames - This is a little confusing. There is an `application.properties` but I think you want to copy `auth0.properties.example` from [here](https://github.com/auth0-samples/auth0-spring-security-api-sample/tree/master/01-Authorization/src/main/resources)? Or maybe not?

lbalmaceda (commented):
Actually, any `.properties` file should be fine for Spring. We are explicitly "requiring" this file on [this lines](https://github.com/auth0-samples/auth0-spring-security-api-sample/blob/87268aedf21e258e0238c8f247c9d789bd9bf7db/01-Authorization/src/main/java/com/auth0/example/App.java#L11-L14). 

I think the problem is that the sample project by default has the file ending in a `.example` extension, which is renamed by the package processor when the user downloads the sample from the quickstart article. But users cloning the repo directly from github must manually rename it or this file won't be found. This is explained in the repo's [readme](https://github.com/auth0-samples/auth0-spring-security-api-sample/blob/87268aedf21e258e0238c8f247c9d789bd9bf7db/01-Authorization/README.md) but not on the quickstart article. Which is correct as is, IMO.

Closing this

PR #7215, opened 2019-02-01
Removed impersonation articles from search

tamigoodall:
Moved articles up one level
Added robots: noindex, nofollow


scottmathson-auth0:
> Moved articles up one level

Curious about this, wouldn't this effectively now create a _new_ article at that path? Not super familiar with the docs setup here so forgive me if that's incorrect.

Just seems like going from where it lives now, live + currently indexed `/docs/users/guides/impersonate-users-using-the-impersonation-api` to moving to >`docs/users/impersonate-users-using-the-impersonation-api` would cause/introduce other issues and we may need to then address some redirects here.

PR #7217, opened 2019-02-04
Appliance: Set up federated access to Dashboard

khorne3:
[Set Up Federated Login to the Manage Dashboard](https://docs-content-staging-pr-7217.herokuapp.com/docs/appliance/admin/federated-access-to-manage)

### DIFF
--- a/articles/appliance/admin/federated-access-to-manage.md
+++ b/articles/appliance/admin/federated-access-to-manage.md
@@ -1,0 +1,64 @@
+---
+description: How to set up federated login to the Manage Dashboard
+section: appliance
+topics:
+    - appliance
+    - admin
+    - signups
+contentType: how-to
+useCase: appliance
+applianceId: appliance76
+---
+# Set Up Federated Login to the Manage Dashboard
+
+If you use an Identity Provider (IdP) to handle your staff logins, you can use the IdP with your root tenant authority (otherwise known as the RTA or your config tenant) to provide federated access to your Dashboard.
+
+By using your IdP, this eliminates the need for you to:
+
+* *For customer-hosted PSaaS Appliance:* Manually create users in the root tenant authority (RTA)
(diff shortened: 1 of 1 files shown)


articles/appliance/admin/federated-access-to-manage.md
```diff
@@ -0,0 +1,64 @@
...
+In your Support ticket, you should include:
+
+* The **email domain(s)** that will be redirected to your Identity Provider
+* The Identity Provider's **single sign on URL**
+* The Identity Provider's **public key** (encoded in PEM or CER format)
+* The **sign out URL** (optional)
+
+You can find most of this information in the XML metadata file provided by your Identity Provider.
+
+Much of this information is included in your Identity Provider's XML metadata file. If you'd prefer, you can send Auth0 this file, along with the email domains that you will be redirecting to the IdP.
```
bianchidotdev: Duplicate of line above. I think I prefer the simpler, you can find the data in the XML metadata file.

articles/appliance/admin/federated-access-to-manage.md
```diff
@@ -0,0 +1,64 @@
...
+
+Once Auth0 receives all of the information we need, we will:
+
+* Create a Connection in your root tenant
+* Provide you with the metadata link containing the information you need to provide to your Identity Provider
```
bianchidotdev, line 62: Now that I've done this a fair few times, there's an additional step required by the customer which is in the docs but a lot of customers miss. It might be good to call it out specifically.

I can't even find it explicitly called out in our Generic Auth0 as a SP docs, but it's essentially mappings between the IdP and the SP in the form of Assertions. It's called out well in Okta. Maybe just note that mappings need to be made:
```
Name: email
Name format (optional): Unspecified
Value: ${user.email}
```

The DSE's have more specific information about custom mapping, but it's not something I've dealt with.

tamigoodall (approved):

PR #7220, opened 2019-02-05
Update user-profile-structure.md

rkhoriander:
Add data types to user profile info


rkhoriander:
https://docs-content-staging-pr-7220.herokuapp.com/docs/rules/references/user-object
https://docs-content-staging-pr-7220.herokuapp.com/docs/users/references/user-profile-structure

erjohnson:
The 3-column table looks odd, wonder if there's a way we can make it look less scrunched

rkhoriander:
@erjohnson - I was following the legacy format on this page:
https://auth0.com/docs/users/search/v3/query-syntax

So I've added it to the PR and fixed the column widths across all of the pages. See what you think now.

erjohnson (approved):

PR #7236, opened 2019-02-07
Restructure/rewrite Applications section.

rkhoriander:


rkhoriander:
https://docs-content-staging-pr-7236.herokuapp.com/docs/applications

articles/getting-started/the-basics.md
```diff
@@ -103,7 +103,7 @@ In our example, `ExampleCo` wants their users to be able to login with username/
 1. Enable all three connections for the mobile app
 
 ::: note
-For more information on the supported identity providers, refer to [Identity Providers Supported by Auth0](/identityproviders). For details on how to enable a connection for an application, refer to [Application Connections](/applications/connections).
+For more information on the supported identity providers, refer to [Identity Providers Supported by Auth0](/identityproviders). For details on how to enable a connection for an application, refer to [Connections](/connections).
```
erjohnson, line 106: `/identityproviders` and `/connections` are the same page

rkhoriander: Yes, but they have been the same thing for a long time (and the applications/connections doc that this link is replacing was more of the same). Also, identityproviders and connections won't be the same very shortly when my other sub-task on Connections moves out--it's one of the issues I've identified and resolved. When I do my search/replace on these links for that PR, I want to be able to locate this and keep the nuance. It's not introducing any new problems in the docs, so I left it alone for the time being since I know it will be resolved soon.

articles/reference/dashboard/settings-api.md
```diff
@@ -0,0 +1,31 @@
+---
```
erjohnson: Are we sure we want this to be /reference/dashboard vs /dashboard/reference?

rkhoriander: Changed to /dashboard/reference and updated referring links.

articles/api-auth/references/dashboard-reference.md
```diff
@@ -0,0 +1,96 @@
+---
```
erjohnson: Do we want to remove this for now?

erjohnson (approved):

articles/api-auth/references/dashboard-reference.md
rkhoriander: Yes, glad you removed this. Not sure what it is doing in this branch; I think it's something Tami was working on in another branch, so it shouldn't even be in here. But I will double-check.

PR #7237, opened 2019-02-07
Typo + some wording adjustments

charlau:
Proposed as part of my job application in the role of *R&D Content Engineer*. I picked the first file I could see would benefit from some improvements. These are very minor (and some might be subjective). __You guys are doing a good job!__ I added the mention [DO NOT MERGE], just because it was suggested and also I am not an employee (*yet!*), so you might or might not be interested in my changes.


jeffreylees (approved):
Thanks for the corrections, and good luck in your process! In the future, feel free to leave off the "Do not merge" - we usually notice when the PR is from outside of the organization, and simply hold it to a more stringent set of standards, you know, so it's ok to make suggestions.

I changed one wording change back, as the two audiences mentioned are not "as if they were the same" but literally are the same. Other than that, all good! 

Thanks!

PR #7245, opened 2019-02-11
[React Native Quickstart Guide] Remove --dev option from yarn install command

thetizzo:
For yarn, adding a `--dev` flag adds this dependency to `devDependencies`.  Since this dependency is for authentication to be used in a production environment eventually, I think we want to add this as a regular dependency, which is what `yarn add` will do.

### DIFF
--- a/articles/quickstart/native/react-native/00-login.md
+++ b/articles/quickstart/native/react-native/00-login.md
@@ -61,7 +61,7 @@ npm install react-native-auth0 --save
 ### yarn
 
 ```bash
-yarn add --dev react-native-auth0
+yarn add react-native-auth0
 ```
 
 ::: note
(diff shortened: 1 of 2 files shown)


thetizzo:
Is there anything else I can do on this PR to help it get accepted?  I can go through other pages with yarn examples and check them for the same issue if that helps.

albertoperdomo:
Can you please have a look @lbalmaceda ?

lbalmaceda (dismissed):
Sorry for the delay. Yes! this makes total sense. We want this as a regular dependency. I don't recall more react native quickstarts, but if you want to go through other pages I'm ok!

thetizzo:
I did a quick search and only found one other location where a yarn command had the `--dev` flag but the npm one didn't so I've updated that as well.  This should be ready to go. 😄

lbalmaceda (approved):
Thanks!

PR #7247, opened 2019-02-12
[DO NOT MERGE] Add OAuth2 current best practice.

gkwang:

### DIFF
--- a/articles/security/oauth2-bcp.md
+++ b/articles/security/oauth2-bcp.md
@@ -1,0 +1,61 @@
+---
+description: Current Best Practices in OAuth2
+toc: true
+crews: crew-security
+topics:
+  - security
+  - OAuth2
+  - Best Practice
+  - users
+contentType:
+  - reference
+useCase:
+  - development
+---
+
+# OAuth2 Current Best Practice
+
+We have compiled several security guidelines for the different grant types specified in the OAuth RFC.
+
+Please be aware that this is not an exhaustive list; further readings and references may be found in the Keep reading section.
+
+## Common Guidelines
+
+* `redirect_uri` should be as strict as possible; http urls should not be allowed in a production application.
+
+* [state parameter](/protocols/oauth2/oauth-state) should always be used to prevent CSRF attacks.
+
+* Access tokens should be stored in-memory rather than utilizing local storage.
+
+
+## Implicit Grant
+
+Per the most current best practice, implicit flows are generally not recommended due to tokens being passed through user agents.
+However, In many cases due to other restrictions, such as CORS not supported in the authorization server, implicit grants may be used.
+The current recommended guideline is to reduce the the impact by allowing only id_tokens returned back from the authorization server and not the access tokens.
+
+However, such recommendation is dependent on the risk profile of your application.
+
+
+## Authorization Code Grant
+
+Authorization code grant is the recommended method over implicit grant whenever possible. The security advantage of the grant prevents a user agent from accessing the tokens returned by the server, preventing an additional attack vector.
+
+The current recommendation is to utilize [PKCE protocol](https://auth0.com/docs/api-auth/tutorials/authorization-code-grant-pkce) wherever applicable to prevent authorization code interception attacks; despite the initial design was for native applications.
+
+
+## Resource Owner Password Grant
+
+Resource Owner Password Grant is generally not recommended for external use since it provides no ability to present the option for consent to the user. A user providing her credentials to the application is presumed to grant consent of all scopes to the application. However, due to reasons such as user experience or backwards compatibility, ROPG may be used.
+
+We recommend limiting the scopes to the minimal requirement to reduce the impact in case the application or tokens are compromised.  
+
+
+## Keep reading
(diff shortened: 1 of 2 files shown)


articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+
+# OAuth2 Current Best Practice
+
+We have compiled several security guidelines for the different grant types specified in the OAuth RFC.
+
+Please be aware that this is not an exhaustive list; further readings and references may be found in the Keep reading section.
+
+## Common Guidelines
+
+* `redirect_uri` should be as strict as possible; http urls should not be allowed in a production application.
```
mosspilot, line 24: Maybe define what strict as possible means -- scheme, domain, subdomain, port, path... 

Could also probably separate this bullet and mention that TLS is absolutely required per the spec. Or maybe add an additional bullet. It doesn't feel strongly worded enough.

articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+
+* [state parameter](/protocols/oauth2/oauth-state) should always be used to prevent CSRF attacks.
+
+* Access tokens should be stored in-memory rather than utilizing local storage.
```
mosspilot, line 28: 👍 , do the links below address how to do that?

articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+## Implicit Grant
+
+Per the most current best practice, implicit flows are generally not recommended due tokens being passed through user agents.
```
mosspilot: typo, should be "due to tokens..."

articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+However, In many cases due to other restrictions, such as CORS not supported in the authorization server, implicit grants may be used.
```
mosspilot, line 34: Aren't there some mitigations that one can take in this situation. Should we call these out or reference them?

articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+The current recommended guideline is to reduce the the impact by allowing only id_tokens returned back from the authorization server and not the access tokens.
```
mosspilot, line 35: Only applicable for SPA's that don't have external APIs to access? Should we discuss that?

articles/security/oauth2-bcp.md
```diff
@@ -0,0 +1,61 @@
...
+
+However, such recommendation is dependent on the risk profile of your application.
+
+
+## Authorization Code Grant
+
+Authorization code grant is the recommended method over implicit grant whenever possible. The security advantage of the grant prevents a user agent from accessing the tokens returned by the server, preventing an additional attack vector.
+
+The current recommendation is to utilize [PKCE protocol](https://auth0.com/docs/api-auth/tutorials/authorization-code-grant-pkce) wherever applicable to prevent authorization code interception attacks; despite the initial design was for native applications.
```
mosspilot, line 44: Is this true also for traditional server side web applications?

mosspilot (commented):
🎉  Looks great, it's nice to have all that data consolidated in one place. It probably couldn't hurt to have a little extra background in the document. It might be helpful to elaborate on the terms/jargon and expect that the reader may not have that background.

articles/security/oauth2-bcp.md
gkwang: One reason this isn't going to go out yet.  I'm waiting on a proper link, then I can add an example.

articles/security/oauth2-bcp.md
gkwang: Still holds true. Although they're not really susceptible to auth code interceptions, the rec is to still use PKCE. One way to think about it is simplifying workflows. Removing the option to allow/disallow and recommend the safer choice reduces confusion.

articles/security/oauth2-bcp.md
gkwang: Do you have any references you'd like to add?

articles/security/oauth2-bcp.md
mosspilot: 👍

articles/security/oauth2-bcp.md
gkwang: I didn't mention that because it's common practice allowing http for development purposes. 
Security isn't the main objective in development environments. 

As for schemes, it can differ from webapps to native apps, and can get a little complicated when discussing in detail. I'd like to keep this simple, but if you have a good reference I can link them here.

articles/security/oauth2-bcp.md
mosspilot: I think you touched on a few of them above. An oldie but a goodie: https://tools.ietf.org/html/rfc6819#section-4.4.2 

Things like shortening token expiration times could be used...

gkwang: good idea.

PR #7265, opened 2019-02-18
Remove all mentions to Lock Analytics add-on

khorne3:
The Lock Analytics library has been deprecated, so I removed all mentions to Facebook and Google Analytics. I removed the `/analytics` folder completely, as well as links from the sidebar.


erjohnson:
~maybe add a redirect from /analytics just for good measure~ scratch that, might conflict in the future

erjohnson (approved):

PR #7266, opened 2019-02-18
Removed window.location.hash = '';. It triggered an additional router…

trydalch:
… event which cancels the subsequent router.navigate call.

Without removing or commenting this line, I was unable to navigate to the login redirect url. It would begin navigating, but then would be interrupted by another router event that was triggered by the hash change.

### DIFF
--- a/articles/quickstart/spa/angular2/_includes/_centralized_login.md
+++ b/articles/quickstart/spa/angular2/_includes/_centralized_login.md
@@ -86,7 +86,6 @@ export class AuthService {
   public handleAuthentication(): void {
     this.auth0.parseHash((err, authResult) => {
       if (authResult && authResult.accessToken && authResult.idToken) {
-        window.location.hash = '';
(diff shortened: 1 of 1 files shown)


articles/quickstart/spa/angular2/_includes/_centralized_login.md
```diff
@@ -86,7 +86,6 @@ export class AuthService {
   public handleAuthentication(): void {
     this.auth0.parseHash((err, authResult) => {
       if (authResult && authResult.accessToken && authResult.idToken) {
-        window.location.hash = '';
```
trydalch, line 89: This line triggered a navigation event that canceled the `router.navigate(['/home']); ` call on line 90/91 made.

joshcanhelp:
Appreciate your contribution here but that quickstart has since been re-written and I'm not see this on the page any longer.

PR #7267, opened 2019-02-18
Remove URL hash fragment from browser's history

alexisluque:
Code side also updated auth0-samples/auth0-javascript-samples#57.

### DIFF
--- a/articles/quickstart/spa/vanillajs/_includes/_centralized_login.md
+++ b/articles/quickstart/spa/vanillajs/_includes/_centralized_login.md
@@ -80,7 +80,7 @@ window.addEventListener('load', function() {
   function handleAuthentication() {
     webAuth.parseHash(function(err, authResult) {
       if (authResult && authResult.accessToken && authResult.idToken) {
-        window.location.replace('#');
+        history.replaceState(null, null, '/');
         localLogin(authResult);
         loginBtn.style.display = 'none';
         homeView.style.display = 'inline-block';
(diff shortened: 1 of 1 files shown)


joshcanhelp (changes requested):
@alexisluque - This change does not affect the browser history, just changes the hash in the address bar, as it did before. I don't know if you can remove it from history once it's there but, if you can, it would probably be with something like `replaceState()`:

https://developer.mozilla.org/en-US/docs/Web/API/History_API

... making sure to do a feature check so you don't call that for browser that don't support it. 

Ping @luisrudge in case he knows of any black magic.

luisrudge:
I agree replaceState is the best to use here. Adoption is pretty good as well: https://caniuse.com/#search=replacestate

alexisluque:
@joshcanhelp I updated with your suggestion.

joshcanhelp:
@alexisluque - Do you have testing steps you can share? And a list of browsers this was tested on?

alexisluque:
@joshcanhelp I've updated the code to fallback on `location.href` if `replaceState` isn't available.
I've tested this in the latest version of Chrome, Firefox, Opera and Edge.
The test steps I used are the following:
* Proceed to login.
* After login is performed, hit the back button, and see that the URL with the access token was removed from the session history.

joshcanhelp:
Thanks @alexisluque! I'll review as soon as I can and get right back to you!

joshcanhelp (approved):
Just tested in the latest Chrome and it works as expected!

joshcanhelp:
Looks like this page has been replaced 👍

PR #7279, opened 2019-02-20
Initial release of docs for core authorization feature

rkhoriander:


rkhoriander:
https://docs-content-staging-pr-7279.herokuapp.com/docs/authorization

jeffreylees (approved):
Since this has had significant eyes on and reviews already, just giving it a cursory look over and approving so that we can get it live today.

PR #7289, opened 2019-02-22
Update set-access-token-lifetime.md

spkjess:
Make default value and maximum time consistent


khorne3:
@spkjess I'm going to leave the doc as is, since the definition calls for the time set in seconds (which is why seconds is listed first and hours second).

PR #7323, opened 2019-03-05
Patch 2

richteambs:
Fixed article to reflect updates to UI - the Rotate Secret CTA is now in the Danger Zone rather than next to the Client Secret field.

The existing image /media/articles/clients/change-client-secret/clients.png is also slightly out-of-date, but I didn't want to screenshot my own client ID etc. and the other screenshot was taken on a Mac, so my screenshot would be inconsistent.


khorne3:
See https://github.com/auth0/docs/pull/7349. Closing this PR since changes have been ported and rebased onto master.

PR #7357, opened 2019-03-13
error in content-type value.

vit100:
value must be: application/x-www-form-urlencoded

### DIFF
--- a/articles/api-auth/tutorials/authorization-code-grant.md
+++ b/articles/api-auth/tutorials/authorization-code-grant.md
@@ -74,7 +74,7 @@ Now that you have an Authorization Code, you must exchange it for an Access Toke
   "method": "POST",
   "url": "https://${account.namespace}/oauth/token",
   "headers": [
-    { "name": "Content-Type", "value": "application/json" }
+    { "name": "Content-Type", "value": "application/x-www-form-urlencoded" }
   ],
   "postData": {
     "mimeType": "application/json",


khorne3:
Confirmed w/ engineering that content-type is indeed json.

vit100:
This is wrong:
According to spec: [https://tools.ietf.org/html/rfc6749#section-4.1.3](https://tools.ietf.org/html/rfc6749#section-4.1.3)

>  Access Token Request
   The client makes a request to the token endpoint by sending the
   following parameters using the "application/x-www-form-urlencoded"
   format per Appendix B with a character encoding of UTF-8 in the HTTP
   request entity-body:

Also if value set to json, then server rejects request with error.
Please review again.

khorne3:
@vit100 Yes, you're correct, my mistake. We will review this and make the appropriate changes. Thank you!

vit100:
Btw sample code defective too. Will create or later




On Fri, Mar 15, 2019 at 1:34 PM -0400, "K Horne" <notifications@github.com> wrote:












@vit100 Yes, you're correct, my mistake. We will review this and make the appropriate changes. Thank you!

vit100:
Pls approve this PR, then I will create new one with reviewed sample, as this one is defective as well - unfortunately internal bug in library `request` make this sample to throw exception

khorne3 (approved):

PR #7375, opened 2019-03-16
Fix outdated documentation

itinsley:
The documentation is wrong/outdated and the implementation seems dubious (if you were to click disable RBAC by mistake it would open all your routes rather than close them, which seems the wrong way around).  I lost a lot of time on this, it's very confusing.

### DIFF
--- a/articles/quickstart/backend/nodejs/01-authorization.md
+++ b/articles/quickstart/backend/nodejs/01-authorization.md
@@ -19,6 +19,22 @@ useCase: quickstart
 
 ## Validate Access Tokens
 
+### Enable RBAC
+*Important* if you do not enable RBAC, the check scopes will merely validate that the scope itself is configured on the API but will not actually validate whether the user has been granted those permissions. 
+Note to Auth0: It is very easy to get a false positive from the examples given which indicate a route is secured by a scope when it is actully merely validating the user has been authenticated and the scope declared.
+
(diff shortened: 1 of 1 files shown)


jimmyjames:
Thanks for the input and PR! We understand that RBAC and its impacts on API authorization can be confusing. We are planning to address that in the future. 

Regarding this: “if you do not enable RBAC, the check scopes will merely validate that the scope itself is configured on the API but will not actually validate whether the user has been granted those permissions.”:

* If RBAC is not enabled, the scope claim should be used to determine if the user has authorized the requested permissions. If you are observing a case where the scopes claim contains the possible permissions on the API, but not the actual granted permissions, please raise an issue with support at https://support.auth0.com, including steps to reproduce.
* Conversely, if RBAC is enabled, and the API is configured to add the granted permissions to the access token, then there will be an additional claim on the token (`permissions`), which will contain all permissions granted to the authorized party. In this case, the `scopes` claim will be the intersection of the scopes requested on the authorization request and the permissions granted. You can learn more about this in [this document](https://auth0.com/docs/dashboard/guides/apis/enable-rbac). This Quickstart does not currently demonstrate that scenario, but it is one that is on our radar to address.

I’m going to close this PR, but please free to continue the conversation in a reported to issue for support if necessary, or in a GitHub issue if there are additional clarifications or questions between scopes and permissions related to RBAC and authorization in general. Thanks!

PR #7385, opened 2019-03-19
Change `user_metadata` to `app_metadata`

dctoon:


dctoon:
Seems like the references to `user_metadata` should also be changed in this doc to `app_metadata`. https://auth0.com/docs/users/guides/change-user-pictures

jeffreylees:
I'm not sure I follow why this would be changed - wouldn't these be items that one would store in user_metadata? The app_metadata is supposed to be for metadata that dictates functionality, and the user metadata for information about the user, correct?

dctoon:
Please close, makes sense if this is specifically to overwrite properties in the `id_token` from the `user_metadata` via a rule then it's fine. For `app_metadata` this is done automatically without the need for a rule.

PR #7409, opened 2019-03-23
Fix broken quickstart link and typo

xplainr:
original points to https://auth0.com/docs/docs/quickstart
now should point to https://auth0.com/docs/quickstarts


jeffreylees (approved):
Thanks!

PR #7412, opened 2019-03-25
removed user.logins_count from user object in rules

ShayMe21:
`user.logins_count` is not available in Rules (anymore?).

Users are expected to use `context.stats.loginsCount` to check for number of times a user has logged in: https://auth0.com/docs/rules/references/context-object


ShayMe21:
This is being [tracked](https://auth0team.atlassian.net/browse/ULP-730) by Engineering to check why `user.logins_count` is not shown anymore. We may not need to merge this PR if Engineering fix this.

ShayMe21:
@rkhoriander you can merge this PR now. Confirmed with @nicosabena that should be removed from docs.

rkhoriander (approved):

PR #7415, opened 2019-03-25
Update password-strength.md

bpogrebi:
password_complexity_options is a number and shouldn't have quotes around it.

### DIFF
--- a/articles/connections/database/password-strength.md
+++ b/articles/connections/database/password-strength.md
@@ -91,7 +91,7 @@ You'll need to add `password_complexity_options` to leverage the new parameter. 
     csrf_token:                   '{{csrf_token}}',                       // DO NOT CHANGE THIS
     ticket:                       '{{ticket}}',                           // DO NOT CHANGE THIS
     password_policy:              '{{password_policy}}',                  // DO NOT CHANGE THIS
-    password_complexity_options:  '{{password_complexity_options}}'       // DO NOT CHANGE THIS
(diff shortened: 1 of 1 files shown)


jeffreylees:
Hi, thanks for the contribution! The value of `password_complexity_options` is actually an object (such as `password_complexity_options:  '{"minLength":8}'`, however, so closing this PR.

bpogrebi:
@jeffreylees, In this case you have a bug. With: password_complexity_options:  '{{password_complexity_options}}', and
"password_complexity_options": { "min_length": 12 }, 
resulting script is: {message:"At least **%d** characters in length",format:[a.minLength],code:"lengthAtLeast"} 
And:  

password_policy:              'excellent',                  // DO NOT CHANGE THIS
password_complexity_options:  '{"minLength":12}',      // DO NOT CHANGE THIS

Results in the default value 10 instead of specified 12. The only way i can make it work correctly is by removing quotes from   password_complexity_options:  **'{{password_complexity_options}}'**:
image

jeffreylees:
Ohk, I was just following on the statement that the value was a number.
Yeah let me test and confirm and can reopen here. Thanks for being
tenacious :)

jeffreylees (approved):
Interestingly, I can get it to work both with and without the quotation marks, on two separate tenants. However, if I reset my template to default, the default template has it without, so I'm going to just leave it at that :)

Thanks again.

PR #7424, opened 2019-03-26
Provide warning on TLS usage

mosspilot:
The OAuth 2 and OIDC specification both rely upon Transport Layer Security. This PR creates a warning for the user to ensure that our customers are not using plaintext HTTP for their production applications. The use of http in the sample application is to help in starting development.


mosspilot:
Closing in favor of a more generic warning

PR #7425, opened 2019-03-27
Update download.md for TLS language

mosspilot:
The OAuth 2.0 specification requires TLS to ensure that the access token and other sensitive information is not leaked. This PR updates the language to warn our users not to use plaintext http for their production applications.


mosspilot:
Plan is to make this more generic across the quick starts. Closing for now.

PR #7431, opened 2019-03-28
Change access method of user properties to by string

nagamoto:
If you access ```http://localhost:3000/dashboard```, will occur error.
Because the reason that ```@user``` doesn't have symbolized keys.

### DIFF
--- a/articles/quickstart/webapp/rails/03-user-profile.md
+++ b/articles/quickstart/webapp/rails/03-user-profile.md
@@ -50,20 +50,20 @@ Add a template which displays the user's profile data.
 ```html
 <!-- app/views/dashboard/show.html.erb -->
 <section class="jumbotron  text-center">
-  <h2><img class="jumbo-thumbnail img-circle" src="${ '<%= @user[:info][:image] %>' }"/></h2>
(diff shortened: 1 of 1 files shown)


joshcanhelp:
@nagamoto - Apologies for the long delay here. On our sample, we use the symbols:

https://github.com/auth0-samples/auth0-rubyonrails-sample/tree/master/03-User-Profile

I just ran that locally to make sure and it works fine. I don't see anything in the sample that converts or allows accessing both ways. That said, it does work both ways. I'd rather the code be consistent but I also don't want anyone running into trouble on the quickstart here. 

Any ideas?

joshcanhelp (commented):
@nagamoto - Apologies for the long delay here. On our sample, we use the symbols:

https://github.com/auth0-samples/auth0-rubyonrails-sample/tree/master/03-User-Profile

I just ran that locally to make sure and it works fine. I don't see anything in the sample that converts or allows accessing both ways. That said, it does work both ways. I'd rather the code be consistent but I also don't want anyone running into trouble on the quickstart here.

Any ideas?

joshcanhelp:
We'll add a backlog task to verify strings or symbols for keys here.

PR #7446, opened 2019-04-03
Update add-permissions-roles.md

tomhundley:
Typo in the docs or a bug in the endpoint- either way, this uses the API identifier (audience) not the API_ID.


jeffreylees (approved):
Thanks!

PR #7452, opened 2019-04-04
Fix single-quote string typo

Jefferson-Faseler:
String was using single quotes, but contained a contraction causing highlighting issue.

Smallest possible PR.


jeffreylees (approved):
Thanks!

PR #7473, opened 2019-04-09
DAE and SSO-dashboard: custom domains support

zxan1285:
New secret - `AUTH0_CUSTOM_DOMAIN` for delegated-admin and `AUTH0_ISSUER_DOMAIN` for sso-dashboard.


charlesjohnson:
"Custom domain for end-users to login (without protocol)." is pretty opaque. Can you use language that provides more context to the end user about why this is useful?

erjohnson:
Maybe: "Your custom domain (without protocol). For example: `login.exampleco.com`." ?

charlesjohnson:
Let's go to "If you have a custom domain name configured, enter it here. For example: login.exampleco.com"

erjohnson (approved):

PR #7483, opened 2019-04-11
Update part-3.md

officeguy:
currently the tutorial cannot be completed due to this error:
JwksError: The JWKS endpoint did not contain any keys


officeguy:
resolved. user error :) had swapped TOKEN_ISSUER and JWKS_URI in .env

PR #7522, opened 2019-04-22
Fix call to Date.now() in isAuthenticated

thepsi:
The `new` was accidentally left-over from when this used to be `new Date().getTime()`.


joshcanhelp:
Statement no longer exists 👍

PR #7524, opened 2019-04-22
Standardize titles/references to flows and use of "single-page" terminology

rkhoriander:


rkhoriander:
For main testing, go to review app and look at microsites, then follow links to Flow, "Add Login using..." flow, and "Call API using..." flow. Entire PR is basically just replacing the way these flows are referred to across all docs using judicious search-and-replace.

tamigoodall (approved):

PR #7527, opened 2019-04-23
Correct role claim in Auth0 rule

bazile:
Samples use 'https://schemas.quickstarts.com/roles' role claim but JS rule is using incorrect 'https://example.com/roles' claim.


joshcanhelp:
Already complete 👍

PR #7537, opened 2019-04-25
Update weibo.md

sacksn:
corrected reference to _*shopify*_ to instead say _*Weibo*_


tamigoodall:
There are about 25 other instances of "shopify" appearing in other places on the docs site, specifically mentioning Liquid syntax and the GH shopify templates and parameters. Do those need to change as well?

tamigoodall (approved):

PR #7543, opened 2019-04-26
Logout returnTo URL missing case

tamigoodall:
Describe what happens if the Logout returnTo URL is missing.


tamigoodall:
https://auth0team.atlassian.net/projects/DR/queues/custom/100/DR-163

jeffreylees (approved):

PR #7494, opened 2019-04-16
Update logs query to find User Search v2 usage

mmaddex:
The query here doesn't work for me, I have to use wild cards to find relevant logs. 
Screen Shot 2019-04-16 at 9 31 19 AM
Screen Shot 2019-04-16 at 9 30 41 AM

PR #7549, opened 2019-04-26
Update migrate-search-v2-v3.md

tamigoodall:
Fixed log search string


mmaddex:
This does not work for tenants using the new logs search. We would still need to also include a query like `type:w AND description:*search_engine*` for `us` tenants (similar to what we had in https://github.com/auth0/docs/pull/7494/files), I don't think there's any way around that (its listed as a breaking change of v3 https://auth0.com/docs/logs/query-syntax#search-engine-v3-breaking-changes). 

Given that we are in the middle of User Search deprecation I think it's most important to get something accurate out there.

mmaddex:
actually it seems `type:w AND description:*search_engine*` works for both my `us` and `eu` tenant. so maybe just update this PR to include the wild cards and we're in business.

khorne3 (approved):

PR #7593, opened 2019-05-06
Update Workday SAML SSO requirements

thameera:
The new format was confirmed by a customer in SF ticket 00427556.

### DIFF
--- a/articles/protocols/saml/saml-apps/workday.md
+++ b/articles/protocols/saml/saml-apps/workday.md
@@ -23,6 +23,7 @@ useCase:
   "passthroughClaimsWithNoMapping": false,
   "mapUnknownClaimsAsIs": false,
   "mapIdentities":        false,
+  "signResponse":         true,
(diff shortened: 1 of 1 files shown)


khorne3:
Hey @thameera! Thanks for bringing this to our attention. Unfortunately, we no longer have access to Salesforce tickets. In spite of that, have we confirmed this configuration on our end? I'd like to double check customer-requested changes to make sure they're not unique situations, etc and unfortunately, this isn't a setup I can easily recreate on my end.

thameera:
FYI I need to collect some further evidence + collect proof, so pls keep the PR open for now.

thameera:
@khorne3 I couldn't reproduce this, unfortunately, because Workday does not provide a self-service trial. 

This is not a widely used integration, and I found only 3 customers who are actively using this (using Kibana). I then checked their tenant configurations and they conform with the new proposed updates. If you think that is enough evidence we could merge this, otherwise let's ignore and close this for now.

khorne3 (approved):

PR #7611, opened 2019-05-09
Adding fully qualified url

amovsesy:
Without the fully qualified URL, the token seemed invalid always. I used https://jwt.io to decode the token and then copied the issuer from the decoded value. Now my APIs work


joshcanhelp (approved):

joshcanhelp:
Already complete 👍

PR #7628, opened 2019-05-14
Android Quickstart: logout from Auth0 authorization server

alexisluque:
Logout from Auth0 authorization server using `/v2/logout` endpoint.


lbalmaceda (changes requested):
I'll block this PR for now. This should be provided by the SDK.

PR #7662, opened 2019-05-23
Add logout docs to Java Quickstart and update common callback URL QS docs to be consistent

jimmyjames:
Addresses QS-124.

### DIFF
--- a/articles/_includes/_logout_url.md
+++ b/articles/_includes/_logout_url.md
@@ -2,7 +2,7 @@
 
 A logout URL is a URL in your application that Auth0 can return to after the user has been logged out of the authorization server. This is specified in the `returnTo` query parameter.
 
-The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, the users will be unable to log out from the application and will get an error.
+The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, users will be unable to log out from the application and will get an error.
 
 <% if (typeof(returnTo) !== "undefined") { %>
   ::: note
--- a/articles/_includes/_logout_url.md
+++ b/articles/_includes/_logout_url.md
@@ -2,10 +2,10 @@
 
 A logout URL is a URL in your application that Auth0 can return to after the user has been logged out of the authorization server. This is specified in the `returnTo` query parameter.
 
-You can set the logout URL for your app in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If you don't set a logout URL, your users will be unable to log out of the application and will get an error.
+The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, the users will be unable to log out from the application and will get an error.
 
 <% if (typeof(returnTo) !== "undefined") { %>
   ::: note
-  If you are following along with the sample project you downloaded from the top of this page, you should set the **Logout URL** to `${returnTo}`.
+  If you are following along with the sample project you downloaded from the top of this page, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `${returnTo}`.
   :::
 <% } %>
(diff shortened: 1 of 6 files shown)


damieng (dismissed):

lbalmaceda (changes requested):
Logout URLs need to be whitelisted in the dashboard. I think that should be part of the guidance as well

jimmyjames:
> Logout URLs need to be whitelisted in the dashboard. I think that should be part of the guidance as well

Yes, but it is brought in by the include `<%= include('../../../_includes/_logout_url') %>`, you can see [here](https://auth0.com/docs/quickstart/webapp/java#configure-logout-urls).

I notice that the deploy for this branch failed, is that something that should be investigated? When I go to the details it brings me to login to Heroku.

jimmyjames:
@lbalmaceda on closer look, while the included `_logout_url` snippet does include information about allowed logout URLs in general, there's no specific inclusion of what URL to specify in the specific context of this quick start.

I could add some info about this to the Logout section of the QS that was added in this PR, but since we already have a common logout include, I propose that we update `includes/_logout_url.md` to be parameterized and display the info, similar to how `_getting_started.md` does it. That way it is consistent and the info is contained in one common place.

Because `_logout_url` is used by many quickstarts, I think we can conditionally render a note until others are updated:

```
<% if (returnTo) { %>
  ::: note
  If you are following along with the sample project you downloaded from the top of this page, you should set the **Logout URL** to `${returnTo}`.
  :::
<% } %>
```

Then from this quickstart can do this, similar to how `_getting_started` is configured:

```
<%= include('../../../_includes/_logout_url', { returnTo: 'http://localhost:3000/login' }) %>
```

Thoughts?

albertoperdomo:
> could add some info about this to the Logout section of the QS that was added in this PR, but since we already have a common logout include, I propose that we update `includes/_logout_url.md` to be parameterized and display the info, similar to how `_getting_started.md` does it. That way it is consistent and the info is contained in one common place.

+1!

lbalmaceda:
@jimmyjames Yup! That will work for future QS and not break existing ones. Thanks

articles/_includes/_logout_url.md
```diff
@@ -3,3 +3,9 @@
 A logout URL is a URL in your application that Auth0 can return to after the user has been logged out of the authorization server. This is specified in the `returnTo` query parameter.
 
 You can set the logout URL for your app in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If you don't set a logout URL, your users will be unable to log out of the application and will get an error.
```
lbalmaceda:
```suggestion
The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, the users will be unable to log out from the application and will get an error.
```

articles/quickstart/webapp/java/_includes/_login.md
```diff
@@ -122,6 +122,40 @@ protected void doGet(HttpServletRequest req, HttpServletResponse res) throws Ser
 }
 ```
 
+## Handle Logout
+
+To properly handle logout, we need to clear the session and log the user out of Auth0. This is handled in the `LogoutServlet` of our sample application.
+
+First, we clear the session by  calling `request.getSession().invalidate()`. We then construct the logout URL, being sure to include the `returnTo` query parameter, which is where the user will be redirected to after logging out. Finally, we redirect the response to our logout URL.
```
lbalmaceda:
```suggestion
First, we clear the session by calling `request.getSession().invalidate()`. We then construct the logout URL, being sure to include the `returnTo` query parameter, which is where the user will be redirected to after logging out. Finally, we redirect the response to our logout URL.
```

articles/_includes/_logout_url.md
```diff
@@ -3,3 +3,9 @@
...
+
+<% if (typeof(returnTo) !== "undefined") { %>
+  ::: note
+  If you are following along with the sample project you downloaded from the top of this page, you should set the **Logout URL** to `${returnTo}`.
```
lbalmaceda:
```suggestion
  If you are following along with the sample project you downloaded from the top of this page, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `${returnTo}`.
```

lbalmaceda (dismissed):

lbalmaceda (approved):
LGTM!

PR #7664, opened 2019-05-23
PHP Web App Quickstart: Add logout implementation

alexisluque:

### DIFF
--- a/articles/quickstart/webapp/php/01-login.md
+++ b/articles/quickstart/webapp/php/01-login.md
@@ -15,7 +15,7 @@ github:
 ---
 <%= include('../_includes/_getting_started', { library: 'PHP', callback: 'http://localhost:3000/' }) %>
 
-<%= include('../../../_includes/_logout_url') %>
+<%= include('../../../_includes/_logout_url', { returnTo: 'http://localhost:3000' }) %>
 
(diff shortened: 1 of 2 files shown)


jimmyjames (commented):
Developers need to add the allowed URL(s) to `returnTo` as part of the app settings. Are we documenting in this quickstart what to set that allowed URL to?

albertoperdomo:
> Developers need to add the allowed URL(s) to `returnTo` as part of the app settings. Are we documenting in this quickstart what to set that allowed URL to?

@jimmyjames Is it technically feasible to pass the logout URL the corresponding partial template or include? We haven't been doing this so far, but I agreed it would be desirable. If I recall correctly, it was due to a technical limitation.

jimmyjames:
@albertoperdomo it should be possible now that #7662 is merged.

You can pass in a `returnTo` parameter to the `_logout_url` template like this:

```
<%= include('../../../_includes/_logout_url', { returnTo: 'http://localhost:3000/login' }) %>
```

If the `returnTo` parameter is specified, it will be used by the partial similar to how it works for login. If it's not specified, it just happily ignores it (don't want to break all existing usages).

lbalmaceda (approved):

PR #7681, opened 2019-05-29
B2C-Phase2-Authorization

PeterGFernandez:


tamigoodall (dismissed):

tamigoodall (dismissed):

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+It’s important to start by distinguishing between Authentication, Authorization, and Access Control. Your Auth0 tenant (your Authorization Server) is typically responsible for Authentication and some or all of Authorization. Access Control however must be the responsibility of the Application or API itself, because access control is almost always contextual: 
 
-* **Authentication**: the process of determining if the user is who they say they are.
-* **Authorization**: the proceess of determining what the user is allowed to do in the system.
-* **Access Control**: the process of limiting a user to only the actions permitted, based on a combination of who the user is, what they are allowed to do in the system, and their consent.
+* **Authentication**: the process of determining if a user is who they say they are.
+* **Authorization**: the proceess of determining what is allowed, based on who a user is and the consent they provide.
```
mostekcm:
```suggestion
* **Authorization**: the proceess of determining what is allowed, based on who a user is, what permissions they have been given, and/or the consent they provide.
```
Not 100% sure if I like the and/or, but I think it applies

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+* Should my applications and/or APIs be enforcing role or permission based access control?
+ 
+Auth0 provides coarse-grained authorization by restricting access to certain applications or APIs based on certain conditions. In a coarse-grained authorization scenario, a [Rule](/rules) can be built that returns an `UnauthorizedError` when, for example, a user attempts access to an application or an API at an incorrect time (as described in this [example](/authorization/concepts/sample-use-cases-rules#allow-access-only-on-weekdays-for-a-specific-application)) - or if the user doesn’t have the right claim(s) contained in their [Metadata](/users/concepts/overview-user-metadata). For an _application_ using OpenID Connect ([OIDC](/protocols/oidc)), this would prevent the allocation of the [ID Token](/tokens/id-token) used to authorize access. Similarly, for an _API_, allocation of any OAuth2 [Access Tokens](/tokens/overview-access-tokens) (used when [calling the API](/api-auth/why-use-access-tokens-to-secure-apis)), could be prevented as described in this [example](/api-auth/restrict-access-api#example-deny-access-to-anyone-calling-the-api).
```
mostekcm, line 29: We should point at app-metadata instead of user-metadata for a user not having the right claims(s), right?  Or are you saying if they haven't set a phone number yet, they can't access a particular app.

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+::: panel Best Practice
+In the same way that [OIDC](/protocols/oidc) is the most commonly used industry-standard protocol for authentication in customer facing applications, we find that [OAuth2](protocols/oauth2) is the most commonly used industry-standard protocol for authorization.
+:::
+
+Auth0 also provides fine-grained access control at both an application level and an API level. For [application level integration](#application-integration), Auth0 allows you to add [custom claims](#id-token-claims) to an ID Token, which your application can then verifiy and subsequently use to enforce access control. In this case you will need to decide what information you require for your application to make access control decisions.
```
mostekcm:
```suggestion
Auth0 also can provide the authorization information needed so the application or API can apply fine-grained access control. For [application level integration](#application-integration), Auth0 allows you to add [custom claims](#id-token-claims) to an ID Token, which your application can then verify and subsequently use to enforce access control. In this case you will need to decide what information you require for your application to make access control decisions.
```

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: My statement was referring to the fact that any metadata associated with a user could be used. We could qualify that this with something like: `app_metadata` for the stuff a user can't update and `user_metadata` for stuff they can?

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: Agree with the `and/or` - especially in the case of first-party applications

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
mostekcm: Honestly, I'm still not sure I would call this "fine-grained" access control.  Fine-grained is usually more deeply around things like I'm not allowed to access your github repo.  But this is also more fine than coarse grained.  Maybe mid-grained 😂

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+::: warning
+When deciding what data to include in OIDC tokens, consider token size, especially if you are passing the token in the URL. Even if you are not passing tokens in the URL, you will also need to consider the potential of exposing sensitive PII (Personally Identifiable Information).
```
mostekcm:
```suggestion
When deciding what data to include in OIDC tokens, consider token size, especially if you are passing the token in the URL. Even if you are not passing tokens in the URL, you will also need to consider the potential of exposing sensitive PII (Personally Identifiable Information).  Token information is not encrypted, so, though it isn't generally a security issue for an ID token to be leaked, it can be a privacy issue depending on the data that is included in the token.
```

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+:::
+
+For [API level integration](#api-integration), Auth0 supports both [custom claims](#access-token-claims) as well as [scope](#access-token-scopes) re-configuration, both within the context of an Access Token. Again, you will need to decide what information will be required in order for your API to make access control decisions, and then your API will need to enforce that access control by validating the contents of the Access Token.
```
mostekcm: Should we comment here about scopes vs custom claims for access tokens?  We need to make the distinction somewhere.  I'll come back to this after reviewing the rest...

PeterGFernandez: We do. It's in the `API level integration` section :-)

PeterGFernandez: Under scopes

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+Additionally, Auth0 has out-of-box support for Role Based Access Control ([RBAC](/authorization/concepts/rbac)). RBAC refers to assigning permissions to users based on their role within an organization. RBAC provides for simpler fine-grained access control by offering a more manageable approach that is often less prone to error.
```
mostekcm:
```suggestion
Additionally, Auth0 has out-of-box support for Role Based Access Control ([RBAC](/authorization/concepts/rbac)). RBAC refers to assigning permissions to users based on their role within an organization. RBAC provides for simpler fine-grained access control by offering a more manageable approach that is less prone to error. 
```

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+## Application integration
+
+In this scenario, your Auth0 tenant provides a token as an indicator of authorized access to an application. For applications utilizing OpenID Connect ([OIDC](/protocols/oidc)) - the industry-standard protocol we typically find most utilized when it comes to customer facing applications - this would be an ID Token expressed as a [JWT](/jwt).
+
+### ID Token claims 
+
+Using Rule extensibility, Auth0 allows you to easily [add custom claims to an ID Token](/tokens/add-custom-claims) based on, for example, a user’s [Metadata](/users/concepts/overview-user-metadata) content. Your application can then verify the ID Token for the necessary claims, and either allow or prevent access to certain functionality as required. Note that though the process of adding custom claims via Rule is streamlined, the Rule engine is flexible and allows you to write custom code that may have negative effects. Therefore it’s important to follow our [rules best practice](/best-practices/rules) guidance anytime you use this extensibility feature.
```
mostekcm, line 51: I'm wondering if we should point at app_metadata directly, you would almost certainly only make access decisions based on claims that were populated from app_metadata.  And not making that explicit here seems like it is likely to result in people doing it wrong.

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+
+## API integration
+
+In this scenario, Auth0 provides support for both first-party and third-party applications. Acting as the authorization server, and with the approval of the user (the resource owner), your Auth0 tenant can provide an Access Token (typically expressed as a [JWT](/jwt)) to an application (client) so that it can access a protected resources hosted by a resource server on behalf of the resource owner. The issued Access Token typically is passed as the Bearer token in the HTTP Authorization header sent to an [API](/api-auth/why-use-access-tokens-to-secure-apis).
```
mostekcm:
```suggestion
In this scenario, Auth0 provides support for both first-party and third-party applications. Acting as the authorization server, and with the approval of the user (the resource owner), your Auth0 tenant can provide an Access Token (typically expressed as a [JWT](/jwt)) to an application (client) so that it can access a protected resources hosted by a resource server, via an API call, on behalf of the resource owner (user). The issued Access Token typically is passed as the Bearer token in the HTTP Authorization header sent to the [API](/api-auth/why-use-access-tokens-to-secure-apis).
```

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: Good suggestion. I'll take a look at that :-)

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+
+Whether you have a single API, or perhaps a suite of logically related [microservice APIs](/api-auth/tutorials/represent-multiple-apis), you can leverage the Access Tokens that Auth0 provides in order to [secure access to your service(s)](/api-auth/why-use-access-tokens-to-secure-apis). Though relatively easy to set this up in the [Auth0 Dashboard](/apis) or through the [Auth0 Management API](/api/management/v2#!/Resource_Servers/post_resource_servers), it's important to review the different application scenarios and API layouts to determine the best architecture for your system.
+
```
mostekcm, line 69: I feel like a callout that clarifies the fact that this only needs to be done if the API you are accessing is not part of the application itself.  If the API is just a backend for the application, cookie based authentication has many benefits and should be considered as an alternative to protecting the API with tokens.

PeterGFernandez: Agreed. I was wondering how to address that, but as you say, a callout would be a great way to do it :-)

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
 
 ::: warning
-When deciding what data to include in OIDC tokens, you need to consider token size, especially if you are passing the token in the URL. Even if you are not passing tokens in the URL, there are other things that you will also need to consider - such as the potential of exposing sensitive PII (Personally Identifiable Information).
+Though you can configure your applications to be first-party and subsequently configure your APIs to allow first-party clients to ignore consent, if you are using `localhost` then Auth0 cannot verify that the application is truly a first-party app so your users will be prompted for consent anyway. To work around this constraint, when testing on your local machine during development, create a [fake local hostname and use that instead](https://community.auth0.com/t/how-do-i-skip-the-consent-page-for-my-api-authorization-flow/6035).
 :::
```
mostekcm, line 81: I know we aren't doing "how-to" documentation, but it still seems like we want to mention that you need to create an API in Auth0 if you want to fetch an access token, and maybe even do a callout that if you are getting an opaque token instead of a JWT for your access token, you probably forgot to pass the audience for your API to /authorize...

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
-When you are considering adding custom claims, we recommend that you choose to store any data you may need to include within the claims in the user's `user` or `app` [Metadata](/users/concepts/overview-user-metadata). Doing so prevents you from needing to call out to an external API to fetch the data, which can negatively impact the performance and scalability of the login sequence. Remember to check out our [metadata best practices](architecture-scenarios/implementation/b2c/b2c-profile-mgmt#metadata) too.
+Even though you have the ability to fully manipulate Access Token Scopes via Auth0 extensibility, as a security best practice you should only remove scopes which are not authorized and refrain from adding scopes that were not requested.
 :::
 
+Though scopes are often used as a way to enforce access permissions for a user, there are situations where it can become [tricky when you use them in this manner](https://auth0.com/blog/on-the-nature-of-oauth2-scopes/). We therefore recommend that you use scopes for their intended purpose (i.e. delegating permission to an application) and use [custom claims](#access-token-claims) for your role-based or other access control scenarios.
```
mostekcm, line 100: Maybe make this a best practice.

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: I do actually mention somewhere that you can use the Dashboard or the Management API to create a Resource Server definition (API). Let me re-view that and refactor to make it clearer :-)

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
```diff
@@ -12,26 +12,84 @@ useCase:
...
+As is the case with ID Tokens, you can [add custom claims to Access Tokens](/tokens/add-custom-claims) using Auth0 Rule extensibility. Once added, your API can then verify an Access Token for the necessary claims and either allow or prevent access to certain functionality as required. 
+
+### Access Token scopes
+
+[OAuth2 Scopes](/scopes/current/api-scopes) are typically used as the mechanism by which an API can determine what actions can be performed on behalf of a user. Scopes can be added on a per API basis to [define specifc access permissions](/dashboard/guides/apis/add-permissions-apis) in the Auth0 Dashboard or through the Auth0 Management API). Scopes can also be manipulated via Auth0 extensibility (e.g. via a Rule, as in this [example](/architecture-scenarios/spa-api/part-2#create-a-rule-to-validate-token-scopes)). The scopes an application requests for accessing an API should depend on what functionality the application needs to perform. Once the requested scopes are authorized, they will be returned in the Access Token which can be subsequently verified by said [API](/api-auth/tutorials/verify-access-token).
```
mostekcm:
```suggestion
[OAuth2 Scopes](/scopes/current/api-scopes) are typically used as the mechanism by which an API can determine what actions can be performed on behalf of a user. Scopes can be added on a per API basis to [define specifc access permissions](/dashboard/guides/apis/add-permissions-apis) in the Auth0 Dashboard or through the Auth0 Management API). Scopes can also be manipulated via Auth0 extensibility (e.g. via a Rule, as in this [example](/architecture-scenarios/spa-api/part-2#create-a-rule-to-validate-token-scopes)). The scopes an application requests for accessing an API should depend on what functionality the application needs the user to give permission for the application to use. Once the requested scopes are authorized, they will be returned in the Access Token which can be subsequently verified by said [API](/api-auth/tutorials/verify-access-token). A good example of this is when you log in to an application that is using a social provider for login: the social provider API requires that the application specifies whether the user will want the application to post items on your behalf. This allows the user to accept or deny this request. This example demonstrates how the user is delegating permission to the application - which is different than the API restricting access based on a user's role and should be handled differently.
```

mostekcm: Maybe include the next (non-best practice paragraph) here...

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: Hmmm. I may have misunderstood, but doesn't our core RBAC implementation utilize scopes in the way we *don't* recommend here? Isn't that then going to mean that our own implementation goes against best practice advice?

PeterGFernandez: I think I'll leave this as-is for now. We can always review at a later date is necessary :-)

articles/architecture-scenarios/implementation/b2c/b2c-authorization.md
PeterGFernandez: I've clarified using `app` metadata via best practice callout.

tamigoodall (approved):

PR #7693, opened 2019-05-31
IUM-512: Update User Search Best Practices to reflect the latest recommendations

panga:
https://auth0team.atlassian.net/browse/IUM-512

### DIFF
--- a/articles/best-practices/search-best-practices.md
+++ b/articles/best-practices/search-best-practices.md
@@ -22,10 +22,12 @@ When running user searches:
   * Avoid dynamic property names.
   * Avoid large schema sizes and deep structures.
   * Avoid storing data you do not need for authentication and authorization purposes.
+* Don't use a search criteria that returns a large data set (more than 1000 results).
 * Don't use existence queries (for example, "give me all users with a property regardless of its value").
-* Don't use full text search or partial searches.
 * Don't poll the search APIs.
 * Don't use large metadata fields (try to keep metadata fields to 2 KB or less).
+* Avoid using wildcard or range operators on searches.
+* Avoid using only metadata fields on searches.
 * If you are using [user search engine v2](/api/management/v2/user-search), check out the section on [migrating from v2 to v3](/users/search/v3/migrate-search-v2-v3).
 
 ::: note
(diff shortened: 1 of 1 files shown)


articles/best-practices/search-best-practices.md
```diff
@@ -22,10 +22,12 @@ When running user searches:
   * Avoid dynamic property names.
   * Avoid large schema sizes and deep structures.
   * Avoid storing data you do not need for authentication and authorization purposes.
+* Don't use a search criteria that returns a large data set (more than 1000 results).
 * Don't use existence queries (for example, "give me all users with a property regardless of its value").
-* Don't use full text search or partial searches.
 * Don't poll the search APIs.
 * Don't use large metadata fields (try to keep metadata fields to 2 KB or less).
+* Avoid using wildcard or range operators on searches.
```
siacomuzzi: is there any way to improve current support of wildcards/ranges operators instead of put this in the doc?
why are we supporting it if it's not recommended?

panga: It is supported, but not a best practice. It causes slow queries that timeout and adds extra load on our servers.

Everything that is in this document is supported, but not recommended.

To give you an example, there are customers doing searches like this: `q=user_id:auth0|*` or `q=user_id:auth0|xxxxxxx*` or `q=email:myemail@mydomain.com*`.

The worst I saw is this one: 
`q=created_at:[1970-01-01 to *]`

panga: Note that I removed the full-text search and partial search that means the same thing, but they are not clear and explicit.

siacomuzzi: I understand your point, I'm just trying to take care of the v3's reputation :)

maybe something like this: https://docs.splunk.com/Documentation/Splunk/7.2.6/Search/Wildcards ?

panga: Agree, I'll rewrite the text.

panga: Text changed.

siacomuzzi (approved):

tamigoodall (approved):

PR #7742, opened 2019-06-11
[QS-94] Small - Updated iOS Objective-C quickstart with logout info

benlizar:


joshcanhelp (approved):

articles/quickstart/native/ios-objc/00-login.md
```diff
@@ -18,3 +18,4 @@ The preferred way to integrate Auth0 in iOS apps is using Swift. Read about it i
 :::
 
 <%= include('_includes/_login_centralized') %>
+<%= include('../../../_includes/_logout_url') %>
```
lbalmaceda, line 21: This adds the whitelisting of the URL. What about the call to `clearSession` that was added to the sample? I think both should be part of the next chapter (session handling) instead of login. Login already has a lot of information (sdk setup, dashboard setup, use of the lib to log in)

lbalmaceda (changes requested):

articles/quickstart/native/ios-objc/00-login.md
joshcanhelp: @benlizar 👆

benlizar: This PR is outdated. Closed the PR

PR #7773, opened 2019-06-14
initial refactor of base content and addition of sidebar

mostekcm:


PeterGFernandez:
🤔 Looks like there was a problem deploying to staging. Might be a syntax error in your conditional logic?

PR #7777, opened 2019-06-14
Only imports are supported for rule configs

pravinady:
Based on the info at the github repo, It looks like only import of rule configs is supported.


nicosabena (approved):
Looks good to me. Management API v2 does not allow exporting rule configuration values (see https://auth0.com/docs/api/management/v2#!/Rules_Configs/get_rules_configs).

PR #7793, opened 2019-06-19
correction to installation text for deploy cli

ShayMe21:
There's no need to create a M2M Application manually when installing via the Dashboard extension. The extension does that for you.

I've moved the scopes required section further down to the section about installing the tool manually.


ShayMe21:
@tamigoodall Could you double check the anchor link "[scopes](#scopes-required)" would work as expected?

dgobaud:
1. i don't think npm i -g auth0-deploy-cli is equivalent of the other two. you must install it in addition to the dashboard app.

2. why does the dashboard auto install step make a "general" app if manual is m2m? they should be same

tamigoodall (approved):

PR #7800, opened 2019-06-19
Fix cURL commands

twistedstream:
My assumption is that the original cURL commands were crafted for a  content type of `application/x-www-form-urlencoded`. However, the POST data is in JSON format (instead of `NAME=VALUE`), so the current commands don't work. I figured JSON format would be more popular, so I fixed the commands to favor that.

### DIFF
--- a/articles/flows/guides/device-auth/includes/request-tokens.md
+++ b/articles/flows/guides/device-auth/includes/request-tokens.md
@@ -11,10 +11,10 @@ To avoid errors due to network latency, you should start counting each interval 
   "method": "POST",
   "url": "https://${account.namespace}/oauth/token",
   "headers": [
-    { "name": "Content-Type", "value": "application/x-www-form-urlencoded" }
+    { "name": "Content-Type", "value": "application/json" }
   ],
   "postData": {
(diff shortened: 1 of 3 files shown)


articles/flows/guides/device-auth/includes/request-tokens.md
```diff
@@ -11,10 +11,10 @@ To avoid errors due to network latency, you should start counting each interval
   "method": "POST",
   "url": "https://${account.namespace}/oauth/token",
   "headers": [
-    { "name": "Content-Type", "value": "application/x-www-form-urlencoded" }
+    { "name": "Content-Type", "value": "application/json" }
   ],
   "postData": {
-    "text": "{\"grant_type\":\"urn%3Aietf%3Aparams%3Aoauth%3Agrant-type%3Adevice_code\",\"device_code\": \"YOUR_DEVICE_CODE\", \"client_id\": \"${account.clientId}\" }"
```
twistedstream, line 17: This URL-encoded version of the grant type was probably necessary with the `application/x-www-form-urlencoded` content type. But for JSON, the URL-encoding isn't necessary, and in fact, it results in an error.

rkhoriander:
The endpoint requires a Content-Type of x-www-form-urlencoded. HAR snippets reformatted with proper mimeType included in [PR 8073](https://github.com/auth0/docs/pull/8073), which is pending merge. So I'm closing this PR. Thanks for bringing this to our attention!

PR #7863, opened 2019-06-26
Create new Rules Best Practice guidance

tamigoodall:
Draft rules best practices - up to Error Handling


PeterGFernandez:
Re-factored to complete Rules Best Practice as a replacement for current guidance in `master` branch. Would be good to get this sanity reviewed before merging.

PeterGFernandez (changes requested):
Requesting final sanity review before merging 👍

articles/best-practices/rules.md
```diff
@@ -1,129 +1,571 @@
...
+contentType: reference
 useCase:
   - best-practices
   - rules
 ---
 
 # Rules Best Practices
 
-This article covers best practices for using [rules](/rules). Before you start writing rules, we recommend that you:
+This article covers some best practices when using [rules](/rules). Rules can be used in a [variety of situations](/rules#what-can-i-use-rules-for-) as part of the pipeline where artifacts for authenticity are generated - i.e. an [ID Token](/tokens/id-token) in [OpenID Connect (OIDC)](/protocols/oidc), an [Access Token](/tokens/overview-access-tokens) in [OAuth 2.0](/protocols/oauth2), or an [assertion in SAML](/protocols/saml/saml-configuration/saml-assertions#use-rules). A new pipeline is created for each authentication request, in which rules execute.
```
tamigoodall: Too many links in this paragraph make it hard to read. Maybe change the OIDC and Access Token links to glossary term links instead (see below). 

Also, this one (/rules#what-can-i-use-rules-for-) ends in a dash. Not sure if it's a typo.

Glossary links to replace other links: 
- `<dfn data-key="openid">OpenID Connect (OIDC)</dfn>`
- `<dfn data-key="access-token">Access Token</dfn>`

articles/best-practices/rules.md
```diff
@@ -1,129 +1,571 @@
...
+
+## Order
+
+The order in which rules are displayed in the [Auth0 Dashboard](/dashboard) (see image above) dictate the order in which the rules will be executed. This is important, as one rule may make one or more definitions within the [environment](#environment) associated with execution that another rule may depend upon. In this case, the rule making the definition(s) should execute before the rule that makes use of them.
+
+## Environment
+
+Rules execute as a series of called JavaScript functions, in an instance of a [Webtask container](https://www.webtask.io/docs/containers). As part of this, a specific environment is provided together with a number of artefacts supplied by both Webtask, and the Auth0 platform. 
+
+### NPM modules
```
tamigoodall: Is "NPM" supposed to be in all caps? Should it be "`npm` modules"?

articles/best-practices/rules.md
```diff
@@ -1,129 +1,571 @@
...
+
+By default, a large list of publicly available npm modules are [supported out-of-the-box](https://auth0-extensions.github.io/canirequire/). This list has been compiled and vetted for any potential security concerns. If you require an npm module that is not supported out-of-the-box, then a request can be made via the [Auth0 support](https://support.auth0.com/) portal or via your Auth0 representative. Auth0 will evaluate your request to determine suitability. There is currently no support in Auth0 for the use of npm modules from private repositories.
+
+::: Best practice
+When using NPM modules to access external services it’s recommended best practice to [keep API requests to a minimum](#minimize-api-requests), [avoid excessive calls to paid services](#limit-calls-to-paid-services), and avoid potential security exposure by [limiting what is sent](don-t-send-entire-context-object-to-external-services). For more information on this see the [performance](#performance) and [security](#security) sections below.
```
tamigoodall: This link seems wrong: `[limiting what is sent](don-t-send-entire-context-object-to-external-services)`

articles/best-practices/rules.md
```diff
@@ -1,129 +1,571 @@
...
+          
+        default:
+          callback(null, user, context);
+          break;
+      
+    } break; 
+```
+
+::: warning
+It’s recommended best practice to [avoid using contextual logic for Multi-Factor Authentication(MFA)](?). For example, **serious security flaws** can surface if use of MFA is predicated on `context.request.query.prompt === 'none'`. Additionally, the contents of the `context` object is **security sensitive**, so you should [**not** directly pass the object to any external or 3rd party service](don-t-send-entire-context-object-to-external-services).
```
tamigoodall: Missing link: `[avoid using contextual logic for Multi-Factor Authentication(MFA)](?)`

This link might be missing `#`: `[**not** directly pass the object to any external or 3rd party service](don-t-send-entire-context-object-to-external-services)`

3rd party => third-party

articles/best-practices/rules.md
```diff
@@ -1,129 +1,571 @@
...
+:::
+
+#### Redirection
+
+[Redirect from rule](/rules/guides/redirect) provides the ability for implementing custom authentication flows that require additional user interaction (i.e. beyond the standard login form) and is triggered via use of [context.redirect](/rules/references/context-object#properties-of-the-context-object). Redirect from rule can only be utilized when using the [`\authorize`](/api/authentication#login) endpoint
```
tamigoodall: Missing period at the end.

articles/best-practices/rules.md
PeterGFernandez: Updated to address this

articles/best-practices/rules.md
PeterGFernandez: Updated to correct link

articles/best-practices/rules.md
PeterGFernandez: Updated links and content syntax

articles/best-practices/rules.md
PeterGFernandez: Updated to address this

articles/best-practices/rules.md
PeterGFernandez: Re-factored to use glossary reference for Access Token. It would be useful to have a glossary reference for ID Token too - which currently seems to be missing :-(

Hyperlink format (with dash at end) is correct; good observation though, thank you :-)

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
 
-To improve performance, write rules that exit as soon as possible.
+![Rules Pipeline](/media/articles/rules/rules-best-practice-pipeline.png)
 
-For example, if a rule has three checks to decide if it should run, the first check should eliminate the most cases. It should then be followed by the check that eliminates the second-highest number of cases for the rule to run, and so on.
+## Size
 
-### Reduce API requests
+Webtasks currently have a [maximum execution limit](https://webtask.io/docs/limits) of 100 kB of code per instance. So the total size of implementation for all enabled rules must not exceed 100 kB. Doing so will have unpredictable results. Note that the 100 kB limit does not include any [`npm`](https://www.npmjs.com/) modules referenced as part of any [`require`](https://nodejs.org/api/modules.html#modules_require_id) statements.
```
mostekcm: Are you sure on the sizes here?  I thought it was 150kB or more now.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+## Size
 
-### Reduce API requests
+Webtasks currently have a [maximum execution limit](https://webtask.io/docs/limits) of 100 kB of code per instance. So the total size of implementation for all enabled rules must not exceed 100 kB. Doing so will have unpredictable results. Note that the 100 kB limit does not include any [`npm`](https://www.npmjs.com/) modules referenced as part of any [`require`](https://nodejs.org/api/modules.html#modules_require_id) statements.  
 
-Try to minimize the number of API calls you make in your rules. Too many can slow down login response time, as well as cause failures during a timeout.
+## Order
 
-Avoid calling the Management API if at all possible, especially in high traffic environments.
+The order in which rules are displayed in the [Auth0 Dashboard](/dashboard) (see image above) dictate the order in which the rules will be executed. This is important, as one rule may make one or more definitions within the [environment](#environment) associated with execution that another rule may depend upon. In this case, the rule making the definition(s) should execute before the rule that makes use of them.
```
mostekcm, line 53: May want to add a comment here about how expensive rules (rules that call out to API's, including the management API) should run as late as possible if you have other rules that can reject the request as unauthorized.  This can prevent unauthorized access from making expensive calls...

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
 
-### Avoid calls to Management API to get Connection-related details
+## Environment
 
-We have recently expanded the connection properties available in rules [context object](/rules/references/context-object). You should now obtain connection info from the context object instead of calling the Auth0 Management API.
+Rules execute as a series of called JavaScript functions, in an instance of a [Webtask container](https://www.webtask.io/docs/containers). As part of this, a specific environment is provided together with a number of artefacts supplied by both Webtask, and the Auth0 platform.
```
mostekcm:
```suggestion
Rules execute as a series of called JavaScript functions, in an instance of a [Webtask container](https://www.webtask.io/docs/containers). As part of this, a specific environment is provided together with a number of artifacts supplied by both Webtask, and the Auth0 platform. 
```

Can leave it, but American spelling might be more recognized, and this spelling is widely used in British English as well.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+### `global` object
 
-To run a rule for only specific applications, check for a [client metadata field](/rules/references/context-object) instead of comparing the client.
+Webtask containers provide the `global` object, which can be accessed across all rules executing within a Webtask container instance. The `global` object acts as a global variable and can be used to cache information, or even define to functions, that can be used across all rules that run.    
 
-This can improve performance as the rule only executes for clients with a particular metadata field, rather than checking Client IDs. It also makes adding new clients and reading the rule code easier.
+Rules can run more than once when a pipeline is executed, and this depends on the [context](#context-object) of operation. For each context in which a rule is run, a new Webtask container *may* be instantiated, and for each instantiation of a new Webtask container the `global` object is reset. Thus any declaration in the `global` object should also include provision for initialization - with that declaration ideally being made as early as possible (i.e. in a rule that runs early in the execution [order](#order)), e.g:
```
mostekcm: It seems that we lost the cache recommendation between rule executions (for caching access tokens from client credentials for example).  This is an important use of the global object.  You talk about resetting here, but we kind of missed the fact that you can cache tokens or other expensively generated items between calls...

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+      });
+    });
+   };
+```
 
-### Reduce calls to paid services
+### `auth0` object
 
-If you have rules that call a paid service, such as sending SMS messages using Twilio, make sure that you only use the service when necessary to avoid extra charges. To help reduce calls to paid services:
+The `auth0` object is an instance of the [Management API Client](https://github.com/auth0/node-auth0#management-api-client) (defined in the [node-auth0](https://github.com/auth0/node-auth0) Node.js client library), and provides limited access to the [Auth0 Management API](/api/management/v2). It is primarily used for [updating metadata](/rules/guides/metadata#update-metadata) associated with the [`user`](#user-object) object from within a rule.
```
mostekcm: We should mention that this is a special instance of the Management API Client.  It doesn't have full functionality.

mostekcm: Also, if you want other logic (like automatic retries on 429 errors), you need to force a more recent version and use that instead.  Though you can initialize it with the auth0.accessToken...

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
-* Ensure that the rule only gets triggered for an authorized subset of users or under other appropriate conditions. For example, you may want to add logic that checks if a user has a particular email domain, <dfn data-key="role">role</dfn>/group, or subscription level before triggering the call to the paid service.
+::: note
+The [access token](/tokens/overview-access-tokens) associated with the `auth0` object has scopes limited to `read:users` and `update:users` only. Typically, these are sufficient for the majority of operations we recommend being performed from within a rule. However, if you require additional scope(s) then you will need to employ an alternative means of [access to the Management API](/api/management/v2/tokens). 
+:::
 
-## Security recommendations
+Like the [`context`](#context-object) object (described below), the `auth0` object contains security sensitive information, so you should not pass it to any external or 3rd party service. Further, the Auth0 Management API is both [rate limited](/policies/rate-limits#management-api-v2) and subject to latency, so you should be judicious regarding [how often calls are made](#minimize-api-requests).
```
mostekcm: Maybe add some ideas for how often is acceptable.

For example:
Good => the first time someone logs in, I have to sync data with an external system, and update the app_metadata to specify that that was done.  (NOTE: only happens once per user, not on every login)
Not good => I update app_metadata before and after I redirect a user to my own page using a redirect rule and this happens on every login!  (NOTE: one call to the management API per login is too many, more than one is definitely a problem.  All calls should be only occasional per user, not on every login).

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
-### Store sensitive values in settings
+## Execution
 
-Sensitive information, such as credentials or API keys, should be stored in your [rules settings](${manage_url}/#/rules) where they will be encrypted. Do not store these values in your rules code. For example, do not write code like this:
+Each rule is executed as a JavaScript function; these functions are called in the [order](#order) that the rules are defined. Rules execute sequentially - that is to say the next rule in order won’t execute until the previous rule has completed.
 
-```js
-const myApiKey = 'abc123';
-```
+In pipeline terms, a rule completes when the [`callback`](#callback-function) function supplied to the rule is called. Failure to call the function will result in a stall of pipeline execution, and ultimately in an error being returned. Each rule must call the `callback` function at least once.
```
mostekcm:
```suggestion
In pipeline terms, a rule completes when the [`callback`](#callback-function) function supplied to the rule is called. Failure to call the function will result in a stall of pipeline execution, and ultimately in an error being returned. Each rule must call the `callback` function exactly once.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+ 
+Rule execution supports the asynchronous nature of JavaScript, and constructs such as [`Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise) objects and the like can be used. Asynchronous processing effectively results in suspension of a pipeline pending completion of the asynchronous operation. A webtask container typically has a [30 second execution limit](https://webtask.io/docs/limits), after which the container may be recycled. A recycle of a container will prematurely terminate a pipeline (suspended or otherwise), ultimately resulting in an error in authentication being returned - as well as resulting in a reset of the [`global`](#global-object) object.
```
mostekcm: Can we double check the timeout on rules?  I'm almost positive it is less than 30.  It doesn't timeout by webtask, but has a specific rule timeout.  I want to say it is 20 seconds.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+::: note
+Setting `context.redirect` will trigger a [redirection](#redirection) once all rules have completed (the redirect is not forced at the point it is set). Whilst all rules must complete within the execution limit of the Webtask container for the redirect to occur, the time taken as part of redirect processing *can* extend beyond that limit. Redirection back to Auth0 via the `/continue` endpoint will cause the creation of a new Webtask container, in the context of the current pipeline, in which all rules will again be run.
```
mostekcm: Would be nice if we new what the /continue timeout is.  I'm pretty sure it is measured in hours, but there is still a limit.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
-Instead, store secrets in the rules settings and access them using the [configuration object](/rules/guides/configuration):
+### `context` object
+
+The [`context`](/rules/references/context-object) object provides information about the context in which a rule is run (such as client identifier, connection name, session identifier, request context, protocol, etc). Using the context object, a rule can determine the reason for execution. For example, as illustrated in the sample fragment below, [`context.clientID`](/rules/references/context-object#properties-of-the-context-object) as well as [`context.protocol`](/rules/references/context-object#properties-of-the-context-object) can be used to implement conditional processing to determine when rule logic is executed. The sample also shows some best practices for [exception handling](#exceptions), use of [`npm` modules](#npm-modules) (for `Promise` style processing), and the [`callback`](#callback-object) object. 
 
 ```js
-const myApiKey = configuration.myApiKey;
+  switch (context.protocol) {
+    case 'redirect-callback':
+      callback(null, user, context);
```
mostekcm: I always type `return callback` and never just `callback`, it helps protect against multiple callback.  It is a good habit...

```suggestion
      return callback(null, user, context);
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+              .
+              .
+            })
+          )
+          .then(function () {
+              .
+              .
+          })
+          .catch(function (error) {
+            callback(new UnauthorizedError(“unauthorized”), user, context);
```
mostekcm:
```suggestion
            return callback(new UnauthorizedError(“unauthorized”), user, context);
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+          });
+        } break;
+          
+        default:
+          callback(null, user, context);
```
mostekcm:
```suggestion
          return callback(null, user, context);
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+:::
 
-For rules that send information to an external service, make sure you are not sending the entire [context object](/rules/references/context-object), since this object may contain tokens or other sensitive data.
+#### Redirection
 
-Instead of sending the context object to an external service to debug your rules, you should use the **Auth0 Real-Time Webtask Logs Extension** instead. For rules that send information to external services, you should only send a *subset* of the less sensitive attributes from the context object.
+[Redirect from rule](/rules/guides/redirect) provides the ability for implementing custom authentication flows that require additional user interaction (i.e. beyond the standard login form) and is triggered via use of [context.redirect](/rules/references/context-object#properties-of-the-context-object). Redirect from rule can only be utilized when using the [`\authorize`](/api/authentication#login) endpoint.
```
mostekcm:
```suggestion
[Redirect from rule](/rules/guides/redirect) provides the ability for implementing custom authentication flows that require additional user interaction (i.e. beyond the standard login form) and is triggered via use of [context.redirect](/rules/references/context-object#properties-of-the-context-object). Redirect from rule can only be utilized when using the [`/authorize`](/api/authentication#login) endpoint.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
 
-### Use HTTPS
+Redirection to your own hosted user interface is performed before a pipeline completes, and can be triggered *only once* per `context.clientID` context. Redirection should only [use HTTPS](#use-https) when executed in a production environment, and additional parameters should be kept to a minimum in order to help mitigate [common security threats](/security/common-threats). Preferably the Auth0 supplied `state` is the only parameter supplied.  
+
+Once redirected, your own hosted user interface will execute in a user authenticated context. You can obtain authenticity artefacts - e.g. an ID Token in [OpenID Connect (OIDC)](/protocols/oidc), and/or an <dfn data-key="access-token">Access Token</dfn> in [OAuth 2.0](/protocols/oauth2) - for a `context.clientID` context **that is not** the one which triggered redirect, and this can be achieved via the use of [silent authentication](/libraries/auth0js/v9#using-checksession-to-acquire-new-tokens). This will create a new pipeline which will cause all rules to execute again, and you can use the `context` object within a rule to perform conditional processing (as discussed above).
```
mostekcm: You really only need to do prompt=none if you have an SPA without a backend.  If they are implementing their redirect rule in a regular web app space, they can just redirect as normal.  This should *not* prompt them unless they don't have an SSO session.  But prompt=none is really only needed in a situation where you are doing it in the background and don't have a way to prompt the user.  The reason I bring this up is that they should *not* use checkSession unless they need implicit grant, and that should really only be done if they don't need an access token.  many of those things are not true here. 

I would recommend that we just state that the redirect rule can redirect users to /authorize to get a user based context.  Since the user has already authenticated they should be automatically be logged in and not need to re-enter credentials.  They also need to be careful about having rules not redirect them again for the same reason or they will get an error after they get redirected twice to their redirect landing page.

Maybe give an example:
1) User logs in with an unverified email
2) Rule redirects user to unverified email landing page (only does this if clientMetadata does not state that this client is the unverified email landing page
3) Landing page redirects to /authorize with clientID for email landing page (this can be checkSession if an SPA, though we should point at our new SDK for SPA's?)
4) User is logged into email landing page
5) User can resend themselves the email for verifying their email from that page 
6) Page can poll (very slowly) the mgmt API to check if verified and page can provide a button for someone to click once they've verified if they don't want to wait for polling to catch it.
7) Page redirects to /continue
8) Rules run again and see that user has now verified their email (NOTE: do they still have to use the mgmt API to re-fetch the user?  I think they do...)
9) User is logged into original application
5) User

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+### `user` object
+
+The [`user`](/rules/references/user-object) object provides access to a cached copy of the user account (a.k.a. [user profile](/users/concepts/overview-user-profile)) record in Auth0. The object provides access to information regarding the user without the need to access the Auth0 Management API - access which is both [rate limited](/policies/rate-limits) and subject to latency.
 
-Always use HTTPS, not HTTP, connections when making calls to external services in your rules code.
+Whilst the contents of the `user` object can be modified - for example, one rule could make a change which another rule could use to influence it’s execution - any changes made will not be persisted. There may be occasions when it becomes necessary to persist, say, [updates to metadata](/rules/guides/metadata#update-metadata) associated with a user and the [`auth0`](#auth0-object) object can be used to perform such operations where required. 
 
-### Enforce timeouts when making external HTTP calls
+::: note
```
mostekcm, line 198: You need to say this a lot, an improvement would be to create an include for this...

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+The [`context`](#context-object) object contains the `primaryUser` property which refers to the user identifier of the primary user. This user identifier will typically be the same as `user_id` property in the root of the `user` object. The primary user is the user that is returned to the Auth0 engine when the rule pipeline completes, and the `user_id` is a unique value generated by Auth0 to uniquely identify the user within the Auth0 tenant. This `user_id` should be treated as an opaque value.
 
-### Don’t use conditional logic for MFA based on silent authentication, device fingerprint, or geolocation
+There are occasions when `primaryUser` must be updated as the primary user may change - i.e. the user returned to the Auth0 engine will be different from the user on rule pipeline entry. [Automatic account linking](/link-accounts#automatic-account-linking) being one of those occasions. On these occasions, a rule must update `primaryUser` to reflect the new primary user identifier. Note that this change *will not* affect any subsequent rule executed in the current instance of the pipeline; the `user` object will remain unchanged.
```
mostekcm: Should we just state that account linking is the *only* appropriate use of this feature?

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+    // Roles should only be set to verified users.
+    if (!user.email || !user.email_verified) {
+      return callback(null, user, context);
+    } else {
+      getRoles(user.email, (err, roles) => {
+        if (err) return callback(err);
+
+        context.idToken['https://example.com/roles'] = roles;
+
+        callback(null, user, context);
```
mostekcm:
```suggestion
        return callback(null, user, context);
```

Again, even though it is not *needed* here, I would always recommend doing it as it can result in maintenance bugs in the future if someone adds more code and forgets to add the return.  Also, it is just a good habit to always type `return callback`

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+      });
+    }
   }
-  // ...
-}
 ```
 
-Device fingerprint:
+As can be seen in the example provided (above), the `callback` function can be called with up to 3 parameters. The first parameter is mandatory and provides an indication of the status of rule operation. The second and third parameters are optional, and represent the user and the context to be supplied to the next rule in the pipeline. If these are specified, then it is a recommended best practice to pass the [`user`](#user-object) and [`context`](#context-object) object (respectively) as supplied to the rule. Passing anything else will have unpredictable results, and may lead to an [exception](#exceptions) or [error](#error-handling) condition.
```
mostekcm: This seems to say that you should not modify the user or context objects.  But it is quite appropriate to add values to user if needed and/or change the context.idToken, context.accessToken, context.redirect, etc.  Might be worth making a statement about the fact that it is okay to modify those values as needed based on the recommendations in this document, but should not recreate new versions of them from scratch as some internal fields and values exist in those objects when they are passed in.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+The status parameter should be passed as either `null`, an instance of an `Error` object, or an instance of an `UnauthorizedError` object. Specifying null will permit the continuation of pipeline processing, whilst any of the other values will terminate the pipeline; an `UnauthorizedError` signalling [denial of access, and allowing information to be returned to the originator of the authentication operation](/rules/references/legacy#deny-access-based-on-a-condition) (regarding the reason why access is denied). Passing any other value for any of these parameters will have unpredictable results, and may lead to an exception or error condition.
```
mostekcm: I think it is worth a callout here to mention that Authentication has already occurred at this point, so any early exit is an "authorization error" and subsequent redirects to /authorize will result in automatic login.  They can mitigate this by automatically redirecting to /v2/logout if desired, but this does not guarantee that it happens because the call to logout can be interrupted.  This is important because any condition that causes an unauthorized, must be re-checked in the next rule execution and should not be able to be bypassed through other conditions (like prompt===none checks).

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
-    };
-  }
-  function getDeviceFingerPrint() {
-    var shasum = crypto.createHash('sha1');
-    shasum.update(context.request.userAgent);
-    shasum.update(context.request.ip);
-    return shasum.digest('hex');
-  }
-}
+  callback(new Error('some description'), user, context);
```
mostekcm:
```suggestion
  callback(new Error('some description'));
```

PeterGFernandez (dismissed):

mostekcm (changes requested):

articles/best-practices/rules.md
PeterGFernandez: I'll check with #dx-extensibility and see what they have to say on the matter :-)

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
 ```
 
-Geolocation:
+Alternatively, an instance of the Auth0 specific `UnauthorizedError` can be returned which will cause an `unauthorized` error condition, with the supplied error description, to be returned to the application that initiated authentication - i.e. the application from which redirect to the `/authorize` end-point, say, was initiated. This allows an application to offer (contiional) retry capability, and additionaly provides capability to implement rule(s) which can be used to [deny access based on certain conditions](/rules/references/legacy#deny-access-based-on-a-condition). For a description of other common authentication error conditions in Auth0, see the [Auth0 SDK library documentation](/libraries/error-messages):
```
mostekcm:
```suggestion
Alternatively, an instance of the Auth0 specific `UnauthorizedError` can be returned which will cause an `unauthorized` error condition, with the supplied error description, to be returned to the application that initiated authentication - i.e. the application from which redirect to the `/authorize` end-point, say, was initiated. This allows an application to offer (conditional) retry capability, and additionally provides capability to implement rule(s) which can be used to [deny access based on certain conditions](/rules/references/legacy#deny-access-based-on-a-condition). For a description of other common authentication error conditions in Auth0, see the [Auth0 SDK library documentation](/libraries/error-messages): 
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
-  // Inadequate verification check
-  if (user.app_metadata.last_location !== context.request.geoip.country_code) {
-    user.app_metadata.last_location = context.request.geoip.country_code;
-    context.multi-factor = {
-      // ...
-    };
-  }
-  // ...
-}
+  callback(new UnauthorizedError('some description'), user, context);
```
mostekcm:
```suggestion
  return callback(new UnauthorizedError('some description'), user, context);
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+:::
 
-Don’t use rules that redirect users to custom multi-factor authentication providers based on silent authentication. Doing so can lead to edge cases where the user can skip the MFA process. For example, we **do not recommend** the following:
+## Debugging
 
-Silent authentication or “prompt === none”
+Out of the box, [runtime debugging](/rules/guides/debug) of a rule is typically achieved via the use of console logging by using the [console.log](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) facility. These is no interactive debugging of a rule available within the Auth0 platform (though one could employ the testing [automation](#automation) technique described below in conjunction with some external interactive source-debugging facility).
```
mostekcm:
```suggestion
Out of the box, [runtime debugging](/rules/guides/debug) of a rule is typically achieved via the use of console logging by using the [console.log](https://developer.mozilla.org/en-US/docs/Web/API/Console/log) facility. There is no interactive debugging of a rule available within the Auth0 platform (though one could employ the testing [automation](#automation) technique described below in conjunction with some external interactive source-debugging facility).
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+          identity.profileData.family_name);
+      }).map(function(identity) {
+        return identity.profileData.family_name;
+      })[0];
+    DEBUG(LOG_TAG, "Computed user.family_name as '", user.family_name, "'");
+      .
+      .
+  
+    //
+    callback(null, user, context);
```
mostekcm:
```suggestion
    return callback(null, user, context);
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
   }
-  //redirect to custom MFA
-  context.redirect = {
-    url: "https://example.com/"
+```
+
+In the example above, a `DEBUG` environment configuration variable has been created, which can be set to `true` or `false`, depending on the execution environment (i.e. production, testing, development etc). The setting of this variable is used to determine when debug logging is performed. Further, a `DEBUGLEVEL` environment` configuration` variable, say, could be created, which could be used to control the debugging log level - e.g. verbose, medium, sparse, etc. 
+
+The example (above) also demostrates declaration of a named funtion. For convenience, providing a function name - using some compact and unique naming convention, say - can assist with diagnostic analysis. Anonymous functions make it hard in debugging situations to interpret the call-stack generated as a result of any [exceptional error](#exceptions) condition, and providing a unique function name addresses this.
```
mostekcm:
```suggestion
The example (above) also demonstrates declaration of a named function. For convenience, providing a function name - using some compact and unique naming convention - can assist with diagnostic analysis. Anonymous functions make it hard in debugging situations to interpret the call-stack generated as a result of any [exceptional error](#exceptions) condition, and providing a unique function name addresses this.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+## Testing
+
+The Auth0 Dashboard provides the facility to [`TRY`](/rules/guides/debug) a rule for the purpose of testing and debugging. This facility allows a mock [`user`](#user-object) and [`context`](#context-object) object to be defined, which is then passed to the rule as part of its execution. The resulting output from the rule (including any console logging) being displayed upon completion. Whilst this provides an immediate at-a-glance way to unit test a rule, it is very much a manual approach, and one which is unable to leverage the use of automated testing tools such as [Mocha](https://mochajs.org/) or [rewire](https://www.npmjs.com/package/rewire).
+
+::: panel Best Practice
+As a best practice, and as part of the recommended [support for the Software Development Life Cycle](https://auth0.com/docs/architecture-scenarios/implementation/b2c/b2c-architecture#sdlc-support), the use of a seperate test Tenant in Auth0 should be employed in order to test any rule/rule changes before deploying to production.
```
mostekcm:
```suggestion
As a best practice, and as part of the recommended [support for the Software Development Life Cycle](https://auth0.com/docs/architecture-scenarios/implementation/b2c/b2c-architecture#sdlc-support), the use of a separate test Tenant in Auth0 should be employed in order to test any rule/rule changes before deploying to production.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+:::
+
+### Automation
+With the help of a little boilerplate however, it’s possible to implement so that a rule can be deployed and executed in an Auth0 Tenant *and*, without modification, be consumed in any Continuous Integration/Continuous Deployment (CI/CD) automated (unit) testing environment:
```
mostekcm:
```suggestion
With the help of a little boilerplate however, it’s possible to implement in a way that enables a rule to be deployed and executed in an Auth0 Tenant *and*, without modification, be consumed in any Continuous Integration/Continuous Deployment (CI/CD) automated (unit) testing environment: 
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+::: note
+The options object passed on the call to `runInThisContext` provides information that can be used to assist with [debugging](#debugging) in the case where any exceptional error condition(s) may arise. Please See the Node.js [documentation](https://node.readthedocs.io/en/stable/api/vm/) for further information regarding this function call.
+:::
+
+The first two objects passed to the rule during testing represent the [`user`](#user-object) and [`context`](#context-object), and these can be mocked in a fashion similar to that employed in the Auth0 Dashboard `TRY` functionality. The [`callback`](#callback-function) function, supplied as the third parameter, can be implemented to simulate pipeline continuation, subsequently performing execution of the next rule in [order](#order). 
+
+::: panel Best Practice
+The `callback` function supplied can be used to ensure execution of callback is performed *at least* once by it (the function) executing test of the next rule in the chain. Implementing test(s) in the supplied function to also ensure that multiple execution of the callback is not performed by a rule is also a recommended best practice.
```
mostekcm:
```suggestion
The `callback` function supplied can be used to ensure execution of the callback is performed *at least* once, by having the (callback) function complete the test and/or provide an assertion. Providing implementation to also ensure that multiple execution of the callback is not performed by a rule is also a recommended best practice.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+Out of the box, Auth0 provides a number of facilities for automated deployment of rule extensibility assets between Auth0 tenant environments. The Auth0 [GitHub](/extensions/github-deploy), [GitLab](/extensions/gitlab-deploy) and [Bitbucket](/extensions/bitbucket-deploy) extensions provide the ability to update rule assets from the respective 3rd party version control system - both manually, and in many instances automatically too (i.e. when a change in the version control system is detected).
+
+In addition, the Auth0 [Deploy CLI](/extensions/deploy-cli) tool can be used to automate deployment between Auth0 tenants. Deploy CLI works with files stored in the file system together with the Auth0 Management API, and provides capability to allow the export of rule assets from an Auth0 tenant, as well as import of them into an Auth0 tenant. Further the tool provides for programmatic control over rule ordering and rule environment [configuration](#environment-variables), as part of deployment automation. In many ways, the Deploy CLI is like a Swiss Army Knife when it comes to rule extensibility deployment in Auth0.
```
mostekcm:
```suggestion
In addition, the Auth0 [Deploy CLI](/extensions/deploy-cli) tool can be used to automate deployment to Auth0 tenants. If you don't use one of the version control systems that are supported with extensions, and/or you want to integrate your deployment with automated tests as part of your continuous integration/deploy system, then this tools is a good alternative to the extensions.  The Deploy CLI works with files stored in the file system together with the Auth0 Management API, and provides capability to allow the export of rule assets from an Auth0 tenant, as well as import of them into an Auth0 tenant. Further the tool provides for programmatic control over rule ordering and rule environment [configuration](#environment-variables), as part of deployment automation. In many ways, the Deploy CLI is like a Swiss Army Knife when it comes to rule extensibility deployment in Auth0.  
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+:::
+
+### Versioning
+There is no version management when it comes to rules in Auth0: changes made to a rule deployed to an Auth0 tenant will be made live immediately as any change written instantly overwrites what's already there. It's therefore recommended that (a) use of version control - such as Git via GitHub, or the like - is employed to provide change management capability, and that (b) use of a seperate [Test Tenant](/dev-lifecycle/setting-up-env) in Auth0 is employed (as part of testing strategy) in order to provide safe testing of any rule/rule changes before deploying to production.  
+
+## Performance
+
+Rules execute as part of a pipeline where artifacts for authenticity are generated, as described in the [Anatomy](#anatomy) section above. As such, an enabled rule will execute for every login operation (interactive or otherwise), every silent authentication, and every time an access token is generated for an API call. This means that even in small scale deployments performance can be a concern, which will only be exacerbated as the scale of deployment increases.
```
mostekcm:
```suggestion
Rules execute as part of a pipeline where artifacts for authenticity are generated, as described in the [Anatomy](#anatomy) section above. As such, an enabled rule will execute for every login operation (interactive or otherwise), every silent authentication, and every time a user credentials related access token is generated for an API call. This means that even in small scale deployments performance can be a concern, which will only be exacerbated as the scale of deployment increases.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+### Avoid unecessary execution
+
+Prefer to implment execution based on conditional logic. For example, to run a rule for only specific applications, check on a specific [`clientID`](/rules/references/context-object) or for specific [`clientMetadata`](/rules/references/context-object) - especially when checking against a single `clientMetadata` value, common across multiple applications. Using `clientMetadata` can also make adding new clients (as well as reading rule code) easier, especially if you have a large number of applications defined.
```
mostekcm:
```suggestion
Prefer to implement execution based on conditional logic. For example, to run a rule for only specific applications, check on a specific [`clientID`](/rules/references/context-object) or for specific [`clientMetadata`](/rules/references/context-object) - especially when checking against a single `clientMetadata` value, common across multiple applications. Using `clientMetadata` can also make adding new clients (as well as reading rule code) easier, especially if you have a large number of applications defined, by reducing the code changes or configuration values needed between environments.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+:::
+
+### Exit early
+For optimal performance, prefer to write rules that complete as soon as possible. For example, if a rule has three checks to decide if it should run, use the first check to eliminate the majority of cases, followed by the check to eliminate the next largest set of cases, and so on and so forth. At the end of each check remember to execute the [callback](#callback-function) function, ideally combined with a (JavaScript) `return` in order to exit the (rule) function. 
+
+### Minimize API requests
+Calls to APIs, especially calls to 3rd party APIs, can slow down login response time, and can cause rule timeout failures due to call latency - ultimately leading to authentication error situations. We recommended keep API requests to a minimum wherever possible within a rule, and to [avoid excessive calls to paid services](#limit-calls-to-paid-services). We also recommend you avoid potential security exposure by [limiting what is sent]() to any API - 3rd party or otherwise.
```
mostekcm:
```suggestion
Calls to APIs, especially calls to third party APIs, can slow down login response time, and can cause rule timeout failures due to call latency - ultimately leading to authentication error situations. We recommended keeping API requests to a minimum wherever possible within a rule, and to [avoid excessive calls to paid services](#limit-calls-to-paid-services). We also recommend you avoid potential security exposure by [limiting what is sent]() to any API - third party or otherwise. 
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+### Minimize API requests
+Calls to APIs, especially calls to 3rd party APIs, can slow down login response time, and can cause rule timeout failures due to call latency - ultimately leading to authentication error situations. We recommended keep API requests to a minimum wherever possible within a rule, and to [avoid excessive calls to paid services](#limit-calls-to-paid-services). We also recommend you avoid potential security exposure by [limiting what is sent]() to any API - 3rd party or otherwise. 
+
+::: panel Best Practice
+The [global](#global-object) object can be used to cache information from API calls, which can subsequently be used across all rules that execute in the pipeline. Prefer to use this to store information instead of repeatedly calling an API.
```
mostekcm: Should mention it can also be used from rule execution to rule execution in some instances...

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+### Minimize API requests
+Calls to APIs, especially calls to 3rd party APIs, can slow down login response time, and can cause rule timeout failures due to call latency - ultimately leading to authentication error situations. We recommended keep API requests to a minimum wherever possible within a rule, and to [avoid excessive calls to paid services](#limit-calls-to-paid-services). We also recommend you avoid potential security exposure by [limiting what is sent]() to any API - 3rd party or otherwise. 
+
+::: panel Best Practice
+The [global](#global-object) object can be used to cache information from API calls, which can subsequently be used across all rules that execute in the pipeline. Prefer to use this to store information instead of repeatedly calling an API. 
+:::
+
+#### Limit calls to paid services
+If you have rules that call paid services - such as sending SMS messages via Twilio - make sure that you only use those servicess when necessary. This not only provides performance enhancement but helps to avoid extra charges too. To help reduce calls to paid services:
```
mostekcm:
```suggestion
If you have rules that call paid services - such as sending SMS messages via Twilio - make sure that you only use those services when necessary. This not only provides performance enhancement but helps to avoid extra charges too. To help reduce calls to paid services:
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+* Disallow public sign-ups to reduce the number of users who can sign up and trigger calls to paid services.
+* Ensure that a rule only gets triggered for an authorized subset of users, or other appropriate conditions. For example, you may want to add logic that checks if a user has a particular email domain, role/group, or subscription level before triggering the call to the paid service.
+
+#### Limit calls to the Management API 
+Outside use of the [`auth0`](#auth0-object) object, prefer to avoid calls to the Auth0 Management API. The Auth0 Management API is [rate limited](/policies/rate-limits#management-api-v2) - which will still be a consideration even when using the `auth0` object too (so be sure to use it sparingly). In addition, Management API functions take varying degrees of time to perform, so will incur varying degrees of latency; executing [user search](/api/management/v2#!/Users/get_users) functionality, for example, should be kept to a minimum and performed only where absolutely necessary - even when executed via the `auth0` object.
```
mostekcm:
```suggestion
Prefer to avoid calls to the Auth0 Management API. The Auth0 Management API is [rate limited](/policies/rate-limits#management-api-v2) - which will still be a consideration even when using the `auth0` object (so be sure to use it sparingly). In addition, Management API functions take varying degrees of time to perform, so will incur varying degrees of latency; executing [user search](/api/management/v2#!/Users/get_users) functionality, for example, should be kept to a minimum and performed only where absolutely necessary - even when executed via the `auth0` object.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+#### Avoid calls to the Management API for Connection-related details
+
+We've recently expanded [connection related properties](/rules/references/context-object) available to the rules [`context`](#context-object) object. You should now be able to obtain connection info from the `context` object instead of needing to call the Auth0 Management API. 
+
+By way of example, if you are using the **Check if user email domain matches configured domain** rule template, check out the latest version [on Github](https://github.com/auth0/rules/blob/master/src/rules/check-domains-against-connection-aliases.js) or on the [Auth0 dashboard](${manage_url}/#/rules/new) to see this in action. Note, the changes won't alter functionality but will improving the performance of rules that had once relied on calls to the Management API.
```
mostekcm:
```suggestion
By way of example, if you are using the **Check if user email domain matches configured domain** rule template, check out the latest version [on Github](https://github.com/auth0/rules/blob/master/src/rules/check-domains-against-connection-aliases.js) or on the [Auth0 dashboard](${manage_url}/#/rules/new) to see this in action. Note, the changes won't alter functionality but will improve the performance of rules that had once relied on calls to the Management API.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+Removing calls to the Management API (as well as the extra call required to get the appropriate <dfn data-key="access-token">Access Token</dfn>) will make your rule code perform better as well as be more reliable. 
+
+### Consider use of explicit timeouts when making API calls
+
+When calling APIs or accessing external services consider specifiying explicit timeout(s). The specific timeout value you choose will typically vary based on your user case, but in general, choosing a one that's as low as possible - whilst bearing in mind the performance charisteristics of the external service - is advised.
```
mostekcm:
```suggestion
When calling APIs or accessing external services consider specifying explicit timeout(s). The specific timeout value you choose will typically vary based on your user case, but in general, choosing a one that's as low as possible - whilst bearing in mind the performance characteristics of the external service - is advised.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+
+::: panel Best Practice
+Whether you choose to employ use of explicit timeouts, or you choose to opt for implicit timout processing, always be sure to cater for [error](#error-handling) and/or [exception](#exceptions) conditions that may occur as a rusult of any timeout period expiration.
```
mostekcm:
```suggestion
Whether you choose to employ use of explicit timeouts, or you choose to opt for implicit timeout processing, always be sure to cater for [error](#error-handling) and/or [exception](#exceptions) conditions that may occur as a result of any timeout period expiration.
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
 ```
 
-### Always check if an email is verified
+### Don’t send entire [`context`](#context-object) object to external services
+
+For rules that send information to an external service, make sure you are not sending the entire [context](#context-object), since this object may contain tokens or other sensitive data. For rules that send information to external services, you should only send a *subset* of the less sensitive attributes from the `context` object when and where necessary.
 
-Whenever granting authorization based on an email address, always start by checking if the user has verified the email address:
+::: warning
+In a similar fashion, avoid passing **any** aspect of the [`auth0`](#auth0-object) object outside of a rule too.
```
mostekcm:
```suggestion
In a similar fashion, avoid passing **any** aspect of the [`auth0`](#auth0-object) object outside of a rule.  
```

articles/best-practices/rules.md
```diff
@@ -1,247 +1,565 @@
...
+For rules that send information to an external service, make sure you are not sending the entire [context](#context-object), since this object may contain tokens or other sensitive data. For rules that send information to external services, you should only send a *subset* of the less sensitive attributes from the `context` object when and where necessary.
 
-Whenever granting authorization based on an email address, always start by checking if the user has verified the email address:
+::: warning
+In a similar fashion, avoid passing **any** aspect of the [`auth0`](#auth0-object) object outside of a rule too.  
+:::
+
+### Check if an email is verified
+
+Whenever granting authorization predicated on email address or email address charisteristics, always start by checking if the email address is verified e.g:
```
mostekcm:
```suggestion
Whenever granting authorization predicated on email address or email address characteristics, always start by checking if the email address is verified e.g:
```

mostekcm (changes requested):

articles/best-practices/rules.md
PeterGFernandez: Re-factored this to reflect recommended best practice guidance, based on recent conversations with the #dx-extensibility team et al: https://auth0.slack.com/archives/C9170558S/p1562775894309800

articles/best-practices/rules.md
PeterGFernandez: Re-factored this based on container definition and re-introduced missing content.

articles/best-practices/rules.md
PeterGFernandez: Re-factored content based on this observation

articles/best-practices/rules.md
PeterGFernandez: Refactored code based on comment

articles/best-practices/rules.md
PeterGFernandez: Re-factored based on comment

articles/best-practices/rules.md
PeterGFernandez: Re-factored content based on comment

articles/best-practices/rules.md
PeterGFernandez: Re-factored based on comment

articles/best-practices/rules.md
PeterGFernandez: Re-factored to add a Best Practice callout regarding this

articles/best-practices/rules.md
PeterGFernandez: I've re-worded this to reflect automatic account linking as the primary use case. However, there _may_ be other use cases so I've not expressed it as the **only** use case: Automatic Account Linking for example essentially deletes the primary user when a new primary user is selected - so any other delete user scenario would need to change `primaryUser`.

articles/best-practices/rules.md
PeterGFernandez: I'm going to leave the "too" as this sentence is about an object "in addition" to `context`

articles/best-practices/rules.md
PeterGFernandez: Re-factored and accepted suggested change

articles/best-practices/rules.md
PeterGFernandez: Re-factored to include suggested update

articles/best-practices/rules.md
PeterGFernandez: Re-factored to include suggested update

articles/best-practices/rules.md
PeterGFernandez: Your re-factor seems to suggest that using the Auth0 Extensions as part of a CI/CD pipeline is acceptable practice. Whilst there's nothing to prevent a customer from doing so, I would argue that it's going to make life a whole less flexible - and potentially harder - if you **don't** use the CLI tooling instead (i.e not just as a good alternative). 

For example, customers may want to add static analysis checking of their rules as part of CI/CD pipeline workflow, and possible gate deployment to an Auth0 tenant (e.g. via something like Gerrit). And that's going to be tricky to do if you use the Extension, as push deployment via extension will trigger *before* analysis tooling is run in the pipeline.

articles/best-practices/rules.md
PeterGFernandez: Re-factored based on comment

articles/best-practices/rules.md
PeterGFernandez: Re-factored to update guidance accordingly

articles/best-practices/rules.md
PeterGFernandez: I'm going to be re-factoring for common includes once the next set of guidance is published - i.e. Hooks or (Custom) Connections, whichever. So I will definitely be looking to re-factor for commonality then. Thanks 👍

articles/best-practices/rules.md
PeterGFernandez: Agreed. I think we can add this to Vnext documentation

articles/best-practices/rules.md
PeterGFernandez: Re-factored this to provide guidance based on the lower timeout value.

articles/best-practices/rules.md
PeterGFernandez: Re-factored this based on recommended guidance, and also the conversation here: https://auth0.slack.com/archives/CED2N848Z/p1562951792106600

articles/best-practices/rules.md
PeterGFernandez: Re-factored guidance in line with this suggestion

articles/best-practices/rules.md
PeterGFernandez: Re-factored guidance based on suggestion

articles/best-practices/rules.md
PeterGFernandez: Re-factored guidance based on this

articles/best-practices/rules.md
PeterGFernandez: Re-factored guidance based on suggestions. 

For the example suggested, I believe we should leave that to Vnext iteration.

articles/best-practices/rules.md
PeterGFernandez: For now I'm going to leave the statement as it stands

PeterGFernandez (approved):
Numerous re-factors based on comments. Some suggestions being deferred until Vnext.

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release.

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (commented):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (approved):
Numerous changes submitted. Some suggestions deferred until Vnext release

PeterGFernandez (dismissed):
Numerous changes submitted. Certain suggestions deferred to Vnext release

rkhoriander:
@PeterGFernandez, screenshots should follow our contributing guide, specifically they need to have no personally identifiable information (extra plugins, etc) and use a standard avatar. We have a script you can run before taking your screenshot that will temporarily change your avatar and tenant name:
https://auth0team.atlassian.net/wiki/spaces/CG/pages/17826032/Replace+Account+Name+Avatar+For+Screenshots

Also, all images should be a max of 1500px wide and compressed using tinypng.com, or they can cause problems with the deployment. I don't mind doing this part for you if you can add the standardized screenshot to the PR.

Lastly, was the diagram approved by Design? Generally, they have specific graphic standards for the docs site and often standardized, branded icons, and we have to get all diagrams created and/or approved by them.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,581 @@
...
+:::
+
+### Exit early
+For optimal performance, prefer to write rules that complete as soon as possible. For example, if a rule has three checks to decide if it should run, use the first check to eliminate the majority of cases, followed by the check to eliminate the next largest set of cases, and so on and so forth. At the end of each check, remember to execute the [callback](#callback-function) function, ideally combined with a (JavaScript) `return` in order to exit the (rule) function. 
+
+### Minimize API requests
+Calls to APIs, especially calls to third-party APIs, can slow down login response time and can cause rule timeout failures due to call latency, ultimately leading to authentication error situations. We recommend keeping API requests to a minimum wherever possible within a rule and avoiding excessive calls to paid services](#limit-calls-to-paid-services). We also recommend you avoid potential security exposure by [limiting what is sent]() to any API&mdash;third party or otherwise.
```
rkhoriander: Last sentence is missing the link.

articles/best-practices/rules.md
```diff
@@ -1,247 +1,581 @@
...
+contentType: reference
 useCase:
   - best-practices
   - rules
 ---
 
 # Rules Best Practices
 
-This article covers best practices for using [rules](/rules). Before you start writing rules, we recommend that you:
+This article covers some best practices when using [rules](/rules). Rules can be used in a [variety of situations](/rules#what-can-i-use-rules-for-) as part of the pipeline where artifacts for authenticity are generated &mdash;i.e., an ID Token in [OpenID Connect (OIDC)](/protocols/oidc), an <dfn data-key="access-token">Access Token</dfn> in [OAuth 2.0](/protocols/oauth2), or an [assertion in SAML](/protocols/saml/saml-configuration/saml-assertions#use-rules). A new pipeline is created for each authentication request, in which rules execute.
```
rkhoriander: Last sentence. If the pipeline is what the rules execute in, then the comma is appropriate. If the authentication request is what the rules execute in, then it is not appropriate. Regardless, I'd suggest simplifying this sentence a bit, as you're likely to have some user confusion either way.

PeterGFernandez:
@rkhoriander: I've updated the dashboard screenshot by running the script you suggested before taking the screenshot, and also sanitising to remove any custom items. In addition I've reconfigured the image to 1500px wide and compressed using tinypng.com :-)

As for the diagram of the Rule pipeline, that's a section taken from the official "Inside the Auth0 Engine diagram" (https://cdn.auth0.com/blog/auth0-raises-100m-to-fuel-the-growth/inside-the-auth0-engine-high-res.jpg). So I'm assuming this is standardized and branded correctly? If this is a problem for the Docs specifically, then feel free to remove the diagram and the link to it: I've re-worded the content so that whether the diagram is present or not makes no difference to the guidance provided 👍

articles/best-practices/rules.md
PeterGFernandez: Updated content to address missing link

articles/best-practices/rules.md
PeterGFernandez: Updated content to remove ambiguity.

mostekcm (dismissed):

rkhoriander (approved):

PR #7873, opened 2019-06-27
Minor changes to Laravel Login quickstart

joshcanhelp:

### DIFF
--- a/articles/quickstart/webapp/laravel/01-login.md
+++ b/articles/quickstart/webapp/laravel/01-login.md
@@ -28,7 +28,7 @@ If you are installing Auth0 to an existing app, you can skip this section. Other
 2. **[Configuration](https://laravel.com/docs/5.7/configuration)**
     * Create a .env file, used later for critical and sensitive Auth0 connection values
     * Make sure `APP_DEBUG` is set to `true`
-    
+
 By the end of those 2 sections, you should have a Laravel application up and running locally or on a test server.
 
 ## Integrate Auth0 in your application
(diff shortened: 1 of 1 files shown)


articles/quickstart/webapp/laravel/01-login.md
```diff
@@ -118,21 +118,7 @@ AUTH0_CLIENT_ID=${account.clientId}
...
-    // ...
-    'domain'        => env( 'AUTH0_DOMAIN' ),
-    // ...
-    'client_id'     => env( 'AUTH0_CLIENT_ID' ),
-    // ...
-    'client_secret' => env( 'AUTH0_CLIENT_SECRET' ),
-    // ...
-);
-```
+In `laravel-auth0.php`, the global helper `env()` is used to retrieve these values and load them in your Laravel app.
```
jimmyjames, line 121: Should this be `getenv()` instead of `env()`? Looking at the sample code it is calling `getenv()`.

articles/quickstart/webapp/laravel/01-login.md
```diff
@@ -148,7 +134,7 @@ Route::get( '/auth0/callback', '\Auth0\Login\Auth0Controller@callback' )->name(
 
 If you load this callback URL now, you should be immediately redirected back to the homepage rather than getting a 404 error. This tells us that the route is setup and being handled. 
 
-Now we need to add this URL to the **Allowed Callback URLs** field in the Application settings screen for the Application used with this app. Add the complete URL, like `https://myapp.com/auth0/callback`.
+Now we need to add this URL to the **Allowed Callback URLs** field in the Application settings screen for the Application used with this app. This should be your Laravel app's `APP_URL` followed by `/auth0/callback`. If you're using one of our samples, this was already configured at the beginning of this tutorial.
```
jimmyjames: > If you're using one of our samples...

Does this mean if the user downloaded the sample from the Quickstart? If so maybe be specific so as not to confuse developers thinking if they just cloned the sample instead downloading it from the Quickstart.

jimmyjames (commented):
Just a couple questions, otherwise looks good.

articles/quickstart/webapp/laravel/01-login.md
joshcanhelp: `env()` is built into Laravel and `getenv()` is from a package (which isn't needed).

jimmyjames (approved):

PR #7897, opened 2019-07-01
React Native Quickstart: Add logout and update download instructions

alexisluque:
Code sample also updated at: auth0-samples/auth0-react-native-sample#24.

### DIFF
--- a/articles/quickstart/native/react-native/00-login.md
+++ b/articles/quickstart/native/react-native/00-login.md
@@ -204,7 +204,7 @@ For more information on the `accessToken`, refer to [Access Token](/tokens/overv
 
 ### Log the User Out
 
-To log the user out, you remove the Access Token from state, and then log the user out from Auth0 authorization server.
+To log the user out, redirect them to the Auth0 log out endpoint by calling `clearSession`. This will remove their session from the authorization server. After this happens, remove the Access Token from the state. 
 
 ```js
  auth0.webAuth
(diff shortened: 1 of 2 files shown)


stevehobbsdev (dismissed):

lbalmaceda (changes requested):
Hi @alexisluque I've updated the sample code now that the latest SDK release has support for logout on android. Can you please update this article to reflect those changes? The code PR (already merged!) is here for your reference https://github.com/auth0-samples/auth0-react-native-sample/pull/25. 
Thanks

articles/quickstart/native/react-native/00-login.md
```diff
@@ -11,7 +11,7 @@ github:
 contentType: tutorial
 useCase: quickstart
 ---
-<%= include('../_includes/_getting_started', { library: 'React Native') %>
+<%= include('../_includes/_getting_started', { library: 'React Native'}) %>
```
lbalmaceda, line 14: ⚡

articles/quickstart/native/react-native/00-login.md
```diff
@@ -195,3 +201,21 @@ Upon successful authentication the user's `credentials` will be returned, contai
 ::: note
 For more information on the `accessToken`, refer to [Access Token](/tokens/overview-access-tokens).
 :::
+
+### Log the User Out
+
+To log the user out, you remove the Access Token from state, and then log the user out from Auth0 authorization server.
```
lbalmaceda: This is confusing. I'd reverse it, because if the step of calling clearSession fails, we should not remove the token from the state. What about...?

```suggestion
To log the user out, redirect them to the Auth0 log out endpoint by calling `clearSession`. This will remove their session from the authorization server. After this happens, remove the Access Token from the state. 
```

lbalmaceda (changes requested):

lbalmaceda (approved):

PR #7916, opened 2019-07-08
Fix it so the B2C code points at the B2C planning instead of B2B plan…

mostekcm:
…ning, also fix one typo


tamigoodall (approved):

PeterGFernandez (commented):
Looks goods 👍

PR #7934, opened 2019-07-10
Update cards.yml

tamigoodall:
Fixed broken "More" links on cards


tamigoodall:
@erjohnson Can you take a look at the link for the "Guides" card - I just added back in the /topics/guides/index.md file that originally was linked from there and added links. Maybe we should do a different card now that we don't have a guides-tutorials main page.

erjohnson (approved):

PR #7936, opened 2019-07-11
Adding tip to best practices about not storing credentials in metadata

asmalser:
A number of customers have either inquired about or are actually storing access tokens and refresh tokens in app_metadata or user_metadata. This is not a good practice, as it enables any dashboard administrator to view and copy those tokens. So I am suggesting this update to explicitly state this is not recommended.


asmalser:
Additional comment: Our existing docs just say that refresh tokens should be "stored securely", and that app_metadata should be used to store read-only information that impacts what a user can access. I suspect this may be leading some of these customers to think that storing Refresh Tokens in app_metadata is OK.

tamigoodall (approved):

PR #7958, opened 2019-07-15
Add paid feature warning for custom domain

rkhoriander:
Per [DR-252](https://auth0team.atlassian.net/projects/DR/queues/custom/97/DR-252).


damieng:
Hey @khorne3 do we want to use the same format in the API docs for consistency?

khorne3:
@damieng I'm inclined to say no, since these banners are designed to be quite eye catching. I think we should note which features are paid or not in API docs, but I don't think large banners are appropriate in that setting. What do you think?

khorne3 (approved):

PR #7984, opened 2019-07-19
Update root profile endpoint reference link

lbalmaceda:
Introduced here https://github.com/auth0/docs/pull/7954. Fixed on this PR

### DIFF
--- a/articles/libraries/lock-android/v2/custom-fields.md
+++ b/articles/libraries/lock-android/v2/custom-fields.md
@@ -79,6 +79,6 @@ CustomField fieldName = new CustomField(R.drawable.ic_field_person, FieldType.TY
 ```
 
 ::: note
-For the fields to be saved at the root level of the user's profile, their keys must match the ones listed in the [endpoint documentation](https://auth0.com/docs/api/authentication#signup).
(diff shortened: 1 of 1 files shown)


lbalmaceda:
@jeffreylees do you know if using the full URL is OK or if I need to use a relative path? Cheers

articles/libraries/lock-android/v2/custom-fields.md
```diff
@@ -79,6 +79,6 @@ CustomField fieldName = new CustomField(R.drawable.ic_field_person, FieldType.TY
 ```
 
 ::: note
-For the fields to be saved at the root level of the user's profile, their keys must match the ones listed in the [endpoint documentation](https://auth0.com/docs/api/management/v2#!/Users/patch_users_by_id).
+For the fields to be saved at the root level of the user's profile, their keys must match the ones listed in the [endpoint documentation](https://auth0.com/docs/api/authentication#signup).
```
jeffreylees:
```suggestion
For the fields to be saved at the root level of the user's profile, their keys must match the ones listed in the [endpoint documentation](/api/authentication#signup).
```

jeffreylees: Relative like so ^ is how they're supposed to be, but if you ever forget and do a full one they usually work anyway :D

khorne3 (approved):

PR #7989, opened 2019-07-22
Add prefix and backslash to Django QS issuer template

lbalmaceda:


articles/quickstart/backend/django/01-authorization.md
```diff
@@ -156,7 +156,7 @@ JWT_AUTH = {
     'JWT_PUBLIC_KEY': publickey,
     'JWT_ALGORITHM': 'RS256',
     'JWT_AUDIENCE': '${apiIdentifier}',
-    'JWT_ISSUER': '${account.namespace}',
+    'JWT_ISSUER': 'https://' + '${account.namespace}' + '/',
```
damieng, line 159: Can't this just be `'https://${account.namespace}/',` ?

lbalmaceda: @damieng Thought about it. But when the user is not logged in, I think it has more visual impact to keep 

`'https://' + YOUR_DOMAIN + '/'` than `'https://YOUR_DOMAIN/'`

We need to be explicit that the backslash at the end IS required.

damieng (approved):

PR #7994, opened 2019-07-22
Update Quickstart authors

alexisluque:

### DIFF
--- a/articles/quickstart/backend/aspnet-core-webapi-v1_1/index.yml
+++ b/articles/quickstart/backend/aspnet-core-webapi-v1_1/index.yml
@@ -7,6 +7,10 @@ language:
   - C#
 framework:
   - ASP.NET Core
+author:
+  name: Damien Guard
+  email: damien.guard@auth0.com
+  community: false
 thirdParty: false
 image: /media/platforms/asp.png
 topics:
--- a/articles/quickstart/backend/aspnet-core-webapi/index.yml
+++ b/articles/quickstart/backend/aspnet-core-webapi/index.yml
@@ -7,6 +7,10 @@ language:
   - C#
 framework:
   - ASP.NET Core
+author:
+  name: Damien Guard
+  email: damien.guard@auth0.com
+  community: false
 thirdParty: false
 image: /media/platforms/asp.png
 topics:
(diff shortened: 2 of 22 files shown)


articles/quickstart/config.yml
```diff
@@ -1,4 +0,0 @@
-author:
```
lbalmaceda: What about those articles without a specific author entry. This was here to show a default in that case. Can we use something like "Auth0" or "SDKs Team" and have it display an auth0 avatar? 🤔

alexisluque: @lbalmaceda Yeah, agreed. Should we provide and email? If the answer is "yes", which one?

albertoperdomo: @davidpatrick Is an email required for the fallback? Maybe to load the gravatar?

lbalmaceda: Yes, I think we should because of gravatar. But no clue which one. I know `support@auth0.com` has a valid gravatar image associated (with the Auth0 logo). Maybe that's enough?

albertoperdomo: @davidpatrick Is the email field actually required? What happens if it's not provided.

davidpatrick: The author isn't necessarily required - https://github.com/auth0/auth0-docs/blob/master/test/schemas/author.js

But I believe it is assumed one is there https://github.com/auth0/auth0-docs/blob/master/client/components/quickstarts/Author/index.js#L27

Not sure what happens when one isn't supplied 😅.  Could test in a review app, or we could just use `support@auth0.com` as @lbalmaceda has mentioned

albertoperdomo: OK, let's use the support address then @alexisluque. Thanks @davidpatrick!

davidpatrick: Would be best to change this file to https://github.com/auth0/docs/pull/8033

lbalmaceda: Please rebase @alexisluque . This file should remain, as we recently merged this https://github.com/auth0/docs/pull/8033

lbalmaceda (changes requested):
Please, re-add the default config file that is currently being used. (And rebase)

lbalmaceda (approved):

PR #8004, opened 2019-07-25
Always use the latest version of the dependency

lukabratos:
It's usually easy to forget and update the dependency version when new version of the SDK is released.

By not specifying the version, CocoaPods and Carthage will always use the latest available version.

### DIFF
--- a/articles/quickstart/native/_includes/_ios_dependency_centralized.md
+++ b/articles/quickstart/native/_includes/_ios_dependency_centralized.md
@@ -5,7 +5,7 @@
 If you are using Carthage, add the following to your `Cartfile`:
 
 ```ruby
-github "auth0/Auth0.swift" ~> 1.13
+github "auth0/Auth0.swift"
 ```
 
 Then, run `carthage bootstrap`.
@@ -20,7 +20,7 @@ If you are using [Cocoapods](https://cocoapods.org/), add the following to your 
 
 ```ruby
 use_frameworks!
-pod 'Auth0', '~> 1.13'
+pod 'Auth0'
 ```
 
 Then, run `pod install`.


articles/quickstart/native/_includes/_ios_dependency_centralized.md
```diff
@@ -5,7 +5,7 @@
 If you are using Carthage, add the following to your `Cartfile`:
 
 ```ruby
-github "auth0/Auth0.swift" ~> 1.13
+github "auth0/Auth0.swift"
```
luisrudge, line 8: @lukabratos can you specify only 1.x versions? Because allowing any version might open the doors to unexpected breaking changes in case of a major release.

luisrudge (changes requested):

articles/quickstart/native/_includes/_ios_dependency_centralized.md
lukabratos: Hi @luisrudge 

If we leave it as it is: `github "auth0/Auth0.swift" ~> 1.13`

Running `carthage update` will resolve to version `1.13` or higher (minor version) as you have suggested.

Currently, [in the docs](https://github.com/auth0/docs/blob/master/articles/quickstart/native/_includes/_ios_dependency_centralized.md) command specified is `carthage bootstrap` which will install dependency with the version specified, which is `1.13`. 

My reason for completely removing the version is the following; these instructions are useful for customers who are integrating your SDK in their project for the first time. So why not using the latest version. Customers who already configured the project will know if the dependency version changes and will be able to handle it.

lbalmaceda: I agree that having a dynamic version would help to maintain _less_ a given quickstart sample and/or article. However, I also agree with @luisrudge that the major version should be fixed to a value. I think that using this is correct for carthage dependencies:

```
github "auth0/Auth0.swift" ~> 1
```

That would be translated to `>=1.0.0 && < 2.0.0`

lukabratos: Sounds good @lbalmaceda 👍

joshcanhelp:
No activity

PR #8010, opened 2019-07-25
Fix API link in Authz how-to

rkhoriander:
Per concierge issue. Link was a temporary placeholder until we pulled the separate doc out. Was only supposed to be referring to a section of the doc, but it appears it was confusing. Separate doc has now been pulled out, so I linked to it instead.


rkhoriander:
https://docs-content-staging-pr-8010.herokuapp.com/docs/authorization/guides/how-to

Step #1.

khorne3 (approved):

PR #8059, opened 2019-08-05
Fix jwt dependency import on DJANGO QS

lbalmaceda:


joshcanhelp (approved):

jhansen8:
@lbalmaceda These changes seem to have broke the import. pyJWT does not have the get_unverified_claims method that jose.jwt has. Is there a different version of pyJWT that you are using? I would appreciate the clarification because these changes have seem to broke the tutorial, at least on windows 10.

PR #8065, opened 2019-08-06
Bumped react-native min version QS-247

benlizar:
There was an outdated minimum requirement for react-native

--- [QuickStart | React-native] minimum requirements are not the same as the project

### DIFF
--- a/articles/quickstart/native/react-native/index.yml
+++ b/articles/quickstart/native/react-native/index.yml
@@ -30,7 +30,7 @@ github:
   org: auth0-samples
   repo: auth0-react-native-sample
 requirements:
-  - React Native 0.59.10
+  - React Native 2.0.1
   - NodeJS 10.16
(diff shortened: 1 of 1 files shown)


lbalmaceda:
@benlizar I know version 0.60 of RN has support for "Android X". Our SDK doesn't support that yet and people has already run into these issues (you can check the react-native-auth0 repo). If we move the "requirements" to the latest version, we would be saying "this sample works with the latest version" which is half-true, because when combined with Android X dependencies, some things break. Do you think it's worth to change this now?

benlizar:
@lbalmaceda Thanks for the feedback, let me check to see if I can improve/investigate a bit how this can affect our QS. 

Added the "Do-not-merge" Label, so is back to "in progress"

damieng:
Any progress on this PR?

PR #8095, opened 2019-08-10
Update troubleshooting section

joshcanhelp:
Add a note about reverse proxies. Ref [Jira SDK-916](https://auth0team.atlassian.net/browse/SDK-916)


lbalmaceda (approved):
I'm not entirely convinced to link to an external site. Specially because you already have added the snippet here. If you think it adds any value, go ahead.

joshcanhelp:
@lbalmaceda - Mostly for attribution. Seems wrong to just pull the code sample and not include where it was pulled from. There is also more information there.

PR #8101, opened 2019-08-13
[Quickstart] Added server side logout Obj-c

benlizar:


articles/quickstart/native/ios-objc/_includes/_login_centralized.md
```diff
@@ -114,6 +114,27 @@ To learn more about the `credentials` object, read the [Credentials](https://git
...
+To clear on server side the session add the following snipped to the HybridAuth.swift file:
+
+```swift
+// HybridAuth.swift
+
+    @objc
+    func logOutUser(callback: @escaping(Bool, String) -> Void){
+        Auth0
+            .webAuth()
+            .clearSession(federated: true){
```
lbalmaceda: I don't think this parameter is necessary. Can you remove it please?

articles/quickstart/native/ios-objc/_includes/_login_centralized.md
```diff
@@ -114,6 +114,27 @@ To learn more about the `credentials` object, read the [Credentials](https://git
...
+                switch $0 {
+                case true:
+                    callback($0, "Logged out successfully")
+                case false:
+                    callback($0, "An error occurred")
+                }
+            }
+    }
+```
+After the logout, the callback will return a BOOL with the logout status.
```
lbalmaceda:
```suggestion
After the call, the callback will receive a BOOL with the logout status.
```

articles/quickstart/native/ios-objc/_includes/_login_centralized.md
```diff
@@ -114,6 +114,27 @@ To learn more about the `credentials` object, read the [Credentials](https://git
 
 <%= include('../../../../_includes/_logout_url') %>
 
+To clear on server side the session add the following snipped to the HybridAuth.swift file:
```
lbalmaceda: The file is already clarified on the snippet and we can remove it👍 

```suggestion
To clear the session on the server side you need to invoke the `clearSession` method. Add the following snippet:
```

articles/quickstart/native/ios-objc/_includes/_login_centralized.md
```diff
@@ -114,6 +114,27 @@ To learn more about the `credentials` object, read the [Credentials](https://git
...
+
+```swift
+// HybridAuth.swift
+
+    @objc
```
lbalmaceda: See if you can remove all these spaces on the left so that the **starting** indentation matches the `// HybridAuth.swift` line and so, all the snippets above.

lbalmaceda (changes requested):
Thanks benja! Left some comments

articles/quickstart/native/ios-objc/_includes/_login_centralized.md
benlizar: @lbalmaceda the parameter is required 
image

benlizar (commented):
Fixed and corrected all suggestions

damieng:
@cocojoe I wonder if we should make clearSessions federated parameter default to false?

https://github.com/auth0/Auth0.swift/blob/34b9258412184faa8b9e81d21c51f8cb5c0d12ec/Auth0/WebAuth.swift#L234 indicates it should but the definition at doesn't make it so https://github.com/auth0/Auth0.swift/blob/34b9258412184faa8b9e81d21c51f8cb5c0d12ec/Auth0/WebAuth.swift#L248

FWIW Auth0.Net's logout specifying federated is optional and defaults to false.

articles/quickstart/native/ios-objc/_includes/_login_centralized.md
albertoperdomo: If it's required please set to false, which is what we are doing across all QS @benlizar (default behavior).

PR #8128, opened 2019-08-20
Doc Update: Using custom SMS Gateways

khorne3:


articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
     - passwordless
     - sms
 contentType: how-to
 useCase: customize-connections
 ---
-# Send One-Time Codes via Your Own SMS Gateway
+# Set Up SMS Gateway for Passwordless Connections
 
-By default, the SMS connection uses Twilio to send the one-time code. If you already have your own infrastructure to send text messages, you cannot or might not want to use Twilio.
+If you are using a Passwordless connection, this guide will show you how you can use custom infrastructure that you've built (instead of Twilio) to send out one-time use codes via SMS messages to users who are logging in.
```
rkhoriander: Kind of a long sentence. User might get a little lost toward the end. :)

rkhoriander: Provide a little more context and a link to the how-to that shows them how to set-up SMS Passwordless Connections. (To help provide context and link to the main docs since they could stumble across this page in a search and it would be orphaned.)

articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
 
-Instead, you'll want your own infrastructure to handle the delivery of one-time codes.
+1. Get an [Access Token for Management API](/api/management/v2/tokens). You will need this to make calls to the Management API to update your Passwordless connection.
 
-## Configure your SMS Gateway
+2. Obtain information about the Passwordless SMS connection.
 
-You will need to configure your own SMS Gateway through the Management API. 
+    To do this, make a GET call to the [Connections](/api/management/v2#!/Connections/get_connections) endpoint. The response from the endpoint will be an array of objects. Each object represents one connection affiliated with your tenant.
```
rkhoriander: Maybe add a description of how to make a call as in this guide:
https://auth0.com/docs/api/management/guides/roles/create-roles

I don't know. I'm on the fence about including these because our API Explorers should take care of this for people, but until our API docs are a little better, I've been told they are necessary. Thoughts?

rkhoriander: Maybe add a description of how to make a call as in this guide:
https://auth0.com/docs/api/management/guides/roles/create-roles

I don't know. I'm on the fence about including these because our API Explorers should take care of this for people, but until our API docs are a little better, I've been told they are necessary. Thoughts?

rkhoriander: Feels like this and the little section below are slightly disorganized. Feels like the second sentence (about the response) deserves to be a separate paragraph before the response example you provide below.

articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
+    
+    Find the connection used for your Passwordless SMS connection (you can search for the `"name": "sms"` property). Notice that the connection currently displays the Twilio information you provided during the setup process.
```
rkhoriander: Feels like this might deserve its own step. Or at least be grouped with the note made below the response block. Seems like it's a bit easy to miss that note otherwise?

articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
+# Set Up SMS Gateway for Passwordless Connections
 
-By default, the SMS connection uses Twilio to send the one-time code. If you already have your own infrastructure to send text messages, you cannot or might not want to use Twilio.
+If you are using a Passwordless connection, this guide will show you how you can use custom infrastructure that you've built (instead of Twilio) to send out one-time use codes via SMS messages to users who are logging in.
 
-Instead, you'll want your own infrastructure to handle the delivery of one-time codes.
+1. Get an [Access Token for Management API](/api/management/v2/tokens). You will need this to make calls to the Management API to update your Passwordless connection.
 
-## Configure your SMS Gateway
+2. Obtain information about the Passwordless SMS connection.
```
rkhoriander: Maybe give a little context in this step so they don't do their own thing and miss the note about copying all the options below? This doc isn't in the how-to format, but it does something similar (tells them up front why they're doing the step):
https://auth0.com/docs/connections/pass-parameters-to-idps#example-wordpress

articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
+        "realms": [
+            "sms"
+        ],
+        "enabled_clients": []
+    }
 ]
 ```
 
-2. Modify the options of the connection:
+     **Make anote of the `id` value of this connection, since you will need it to update your connection with your SMS Gateway information.**
```
rkhoriander: a note

articles/connections/passwordless/sms-gateway.md
```diff
@@ -1,66 +1,74 @@
...
 
- - `provider`: Set this to `sms_gateway`
- - `gateway_url`: Set this to the URL of your SMS Gateway. Note that Auth0 must be able to reach it for this to work.
+3. Update the connection. You can do this by making a PATCH call to the [Update a Connection](/api/management/v2#!/Connections/patch_connections_by_id) endpoint. More specifically, you'll be updating the connections `options` object to provide information about the SMS Gateway.
 
-```
+    **You must send the entire `options` object with each call, otherwise, your new payload will overwrite the existing data.**
```
rkhoriander: each call; otherwise,

rkhoriander: Avoid the word payload? Good for tech users, maybe not so much for people who haven't studied JWTs?

articles/connections/passwordless/sms-gateway.md
```diff
@@ -72,32 +80,48 @@ This will return your SMS connection with your Twilio settings:
...
       }
-    }
+    },
+    "is_domain_connection": false,
+    "enabled_clients": []
 }
 ```
 
-3. Send the updated configuration to the Management API using the [PATCH connections endpoint](/api/v2#!/Connections/patch_connections_by_id).
+    If your SMS Gateway accepts authenticated requests that are token-based, you can add the following to your `options` object:
```
rkhoriander: Seems like this section (authenticated requests) deserves a subheading?

articles/connections/passwordless/sms-gateway.md
```diff
@@ -72,32 +80,48 @@ This will return your SMS connection with your Twilio settings:
...
 
-After updating the connection for any user that signs up or authenticates using the <dfn data-key="passwordless">Passwordless</dfn> SMS connection, the following payload will be sent to your SMS gateway:
+```json
 
+"gateway_authentication": {
+    "method": "bearer",
+    "subject": "urn:Auth0",
+    "audience": "urn:MySmsGateway",
```
rkhoriander: Is this correct?

articles/connections/passwordless/sms-gateway.md
```diff
@@ -72,32 +80,48 @@ This will return your SMS connection with your Twilio settings:
...
+    "secret": "MySecretToSignTheToken"
+}
 ```
+
+    With this configuration, a JWT will be added to the `Authorization` header which contains a token with the `subject` and `audience` configured in the connection and signed with the `secret` when the payload is sent to your SMS Gateway.
```
rkhoriander: If you're a new user, this is confusing. What subject and audience configured in the connection? Could get away with simply referencing and linking to the how-to where this is configured.

What's a JWT? Maybe link to that section. Or maybe add a tiny bit of context.

Also, which secret? App secret? Password for their authenticated gateway that does not reside in Auth0? If I'm a newbie, I might be confused.

articles/connections/passwordless/sms-gateway.md
```diff
@@ -72,32 +80,48 @@ This will return your SMS connection with your Twilio settings:
...
+```json
 {
   "recipient": "+1 399 999",
   "body": "Your verification code is: 12345",
   "sender": "+1 234 567"
 }
 ```
 
-If you set the `forward_req_info` property to **true**, the gateway will also receive information from the HTTP request that initiated the Passwordless process. This includes the IP address of the client calling `/passwordless/start` and its User Agent.
+If you set the `forward_req_info` property to `true`, the gateway will also receive information from the HTTP request that initiated the Passwordless process. This includes the IP address of the client calling `/passwordless/start` and its User Agent.
```
rkhoriander: Where? We've just been talking about authenticated requests and are switching out of that now. Might be good to point them to where they should include this. (It makes sense to me, but I can see someone making a mistake here.)

articles/connections/passwordless/sms-gateway.md
```diff
@@ -117,39 +141,4 @@ If the SMS Gateway returns an HTTP code other than 200, the `/passwordless/start
...
-        "length": 6
-      }
-    }
-}
-```
-
-With this configuration, when the payload is sent to the SMS Gateway, a JWT will be added to the `Authorization` header which contains a token with the `subject` and `audience` configured in the connection and signed with the `secret`.
-
-Additionally, if your secret is base64-url-encoded, you can set `options.gateway_authentication.secret_base64_encoded` to `true`.
+**The `error_description` field is not part of the endpoint contract and is subject to change.**
```
rkhoriander: What is an endpoint contract? What are we actually telling them? Not to use error_description? If so, why are we telling them what it is? Is there something more useful we should suggest instead? I'd be frustrated if I saw this at the end of a doc because I would be reading to be told what to do for error catching, and that isn't really what we're doing here right now.

rkhoriander:
@khorne3, can you please check why this review app is failing to build? Rebase from master maybe?

rkhoriander (approved):

PR #8190, opened 2019-09-03
Edge Case: Custom DB  with import mode enabled and leaked password pr…

kaymaylove:
…otection enabled

### DIFF
--- a/articles/users/guides/configure-automatic-migration.md
+++ b/articles/users/guides/configure-automatic-migration.md
@@ -48,6 +48,10 @@ If an un-migrated user confirms a password change, their user profile will be cr
 You may see unexpected behavior if you return differing user profiles in the `login` and `get_user` scripts.
 :::
 
+:::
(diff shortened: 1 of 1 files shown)


articles/users/guides/configure-automatic-migration.md
```diff
@@ -48,6 +48,10 @@ If an un-migrated user confirms a password change, their user profile will be cr
 You may see unexpected behavior if you return differing user profiles in the `login` and `get_user` scripts.
 :::
 
+:::
+If you are migrating users from your legacy db into Auth0's database, if the password policy isn't consistent from one to the other, users won't be able to log in. For example, if you have users in the legacy database using leaked passwords and you have leaked password protection enabled, users who are trying to sign in with leaked passwords with not be able to sign in. They will have to reset their password to match the new password policy.
```
khorne3, line 52:
```suggestion
When migrating users from your legacy database to Auth0's database, users won't be able to log in if your password policies aren't consistent across the two databases.

For example, if you've enabled breached password protection for your Auth0 database, and users in your legacy database attempt to sign in using leaked passwords, they will not be able to do so. The users will need to reset their passwords.
```

khorne3: @kaymaylove What do you think ☝️

PR #8208, opened 2019-09-05
Updated Angular2 quickstart code samples

davidmkershaw:
Documentation doesn't mention a config.json, so source as written does not compile. Also, generating a service via Angular doesn't place code in new folder, so `auth` folder designation does not work since service is created in `app` folder.


stevehobbsdev (approved):
Thanks for the contribution, @davidmkershaw!

PR #8245, opened 2019-09-17
Update _centralized_login.md

vhothur:
Updated typo in auth.service.ts file path


stevehobbsdev (commented):
@vhothur We've just merged a PR which does the opposite of what you're proposing.

The command `ng generate service auth` should generate the file `src/app/auth.service.ts` - is that not what you're seeing?

I realise however that the associated sample does have this file inside the `auth` subfolder, but purely following the steps in the tutorial, the current path should be correct.

vhothur:
@stevehobbsdev Makes sense. I'll close the request

PR #8293, opened 2019-10-01
Duplicate note

landitus:
There are 2 very similar notes next to each other referring to not being able to use a SparkPost EU account. I removed the softer one.

Screen Shot 2019-10-01 at 14 33 44


erjohnson (approved):
Thanks 😄

PR #8348, opened 2019-10-09
Update react-native quickstarts

lbalmaceda:
Updates the quickstart to reflect _requirements_ for latest major and use of the `pod install` command. 
In addition, it re-orders the article showing first how to set up the linking on each platform and then showing how to set up the dashboard with those values defined on the linking section for callback and logout urls.

Lastly, there were 2 small snippets not being reused anywhere that I decided to move into the article for better readability. And some "includes" files that were not being used anywhere.

Note for the reviewer: Check the deployment as it will be easier to read ;) 

Sample PR: https://github.com/auth0-samples/auth0-react-native-sample/pull/26


articles/quickstart/native/react-native/00-login.md
```diff
@@ -86,17 +32,22 @@ yarn add react-native-auth0
...
-react-native link react-native-auth0
+npm install react-native-auth0 --save
 ```
 
-::: note
-For further reference on linking libraries, check [the official documentation](https://facebook.github.io/react-native/docs/linking-libraries-ios.html).
-:::
+### Additional iOS step: Install the Module Pod
+
+CocoaPods is the package management tool for iOS that React Native framework uses to install itself into your project. For the iOS native module to work on your iOS app you must first install the library Pod. If you're familiar with older React Native SDK versions this is similar to what was called _linking a native module_. Currently, the process is simpler:
```
jimmyjames: > CocoaPods is the package management tool for iOS that React Native framework uses to install itself into your project.

Either add "the" before "React Native framework..." or remove the word "framework".

> For the iOS native module to work on your iOS app you must first install the library Pod.

Instead of "to work on your iOS app..." consider "to work with your iOS app".

> If you're familiar with older React Native SDK versions this is similar to what was called _linking a native module_.

Add a comma after the word "versions": If you're familiar with older React Native SDK versions, this is similar to what was called _linking a native module_.

> Currently, the process is simpler:

Consider: "The process is now simplified:"

articles/quickstart/native/react-native/00-login.md
```diff
@@ -181,16 +136,64 @@ and then register a URL type entry using the value of `CFBundleIdentifier` as th
...
 ```
 
 ::: note
-The value org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier) is the default for apps created with React Native CLI, you may have a different value.
+If your application was generated using the React Native CLI, the default value of `$(PRODUCT_BUNDLE_IDENTIFIER)` dynamically matches `org.reactjs.native.example.$(PRODUCT_NAME:rfc1034identifier)`. For the sample app, this value matches `com.auth0samples`.
 :::
 
+Take note of this value as you'll be using it to define the callback URLs below. If desired, you can change it using XCode in the following way:
+
+- Open the `ios/TestApp.xcodeproj` file replacing 'TestApp' with the name of your app or run `xed ios` on a Terminal from the app root.
```
jimmyjames: Consider: "Open the `ios/TestApp.xcodeproj` file and replace 'TestApp' with the name of your app or run `xed ios` on a Terminal from the app root."

jimmyjames (changes requested):

articles/quickstart/native/react-native/00-login.md
lbalmaceda: Good feedback!

articles/quickstart/native/react-native/00-login.md
lbalmaceda: I think your suggestion changes a bit the reading. My intention is to say open `ios/{your app name}.xcodeproj`. That's why I'm saying "open ios/myapp replacing 'myapp' with your app name". Now that you have the context, would you rewrite it in a different way @jimmyjames ?

lbalmaceda: I've pushed a suggestion using `ios/<YOUR PROJECT>.xcodeproj`, as I saw it was being used above in the doc. LMK what you think.

jimmyjames (approved):

PR #8355, opened 2019-10-10
NodeJS EA QS

davidpatrick:
## Description
This PR is to add a Quickstart for the Early Access express oidc library

👉 [Deployed here](https://docs-content-staging-pr-8355.herokuapp.com/docs/quickstart/webapp/nodejs-ea)

### DIFF
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -20,7 +20,7 @@ github:
 <%= include('../../../_includes/_callback_url') %>
 
 :::note
-If you are running this project on your local machine, the callback URL you need to whitelist in the **Allowed Callback URLs** field is `http://localhost:3000/callback`.
+If you are running this project on your local machine, your application's callback URL is `http://localhost:3000/callback`
 :::
 
 <%= include('../../../_includes/_logout_url' }) %>
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -80,7 +80,7 @@ app.use(auth({
 }));
 ```
 
-Now that your application has the middleware installed, you can use [various helpers and values](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the `express-openid-connect` library.  Here is an example of using `req.openid.isAuthenticated`
+Now that your application has the middleware installed, you can use [the helper functions and contextual data](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the Express OpenID Connect library.  Here is an example of how to use `req.openid.isAuthenticated()`:
 
 ```js
 app.get('/', (req, res) => {
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -31,7 +31,7 @@ If you are running this project on your local machine, the logout URL you need t
 
 ## Integrate Auth0
 ### Install Dependencies
-Install the following dependencies
+Your application will need the following packages:
 
 - [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
 - [`express-session`](https://github.com/auth0/express-session) - session middleware for Express.
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -105,7 +105,7 @@ app.use(auth({
   baseURL: 'http://localhost:3000',
   // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
   issuerBaseURL: 'https://${account.namespace}',
-  // clientID - This is your Auth0 Client ID found in your Auth0 Application Settings
+  // clientID - This is the Client ID found in your Auth0 Application settings
   clientID: '${account.clientId}',
   // clientSecret - This is your Auth0 Client Secret found in your Auth0 Application Settings
   clientSecret: '${account.clientSecret}'
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -16,8 +16,18 @@ github:
 ---
 
 <%= include('../../../_includes/_new_app', { hideDownloadSample: true, hideDashboardScreenshot: true }) %>
+
 <%= include('../../../_includes/_callback_url') %>
+
+:::note
+If you are running this project on your local machine, the callback URL you need to whitelist in the **Allowed Callback URLs** field is `http://localhost:3000/callback`.
+:::
+
 <%= include('../../../_includes/_logout_url' }) %>
+
+:::note
+If you are running this project on your local machine, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `http://localhost:3000/logout`.
+:::
 
 ## Integrate Auth0
 ### Install Dependencies
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -104,7 +104,7 @@ app.use(auth({
   // baseURL - There is where your /callback, /login & /logout URLs will be registered
   baseURL: 'http://localhost:3000',
   // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
-  issuerBaseURL: '${account.namespace}',
+  issuerBaseURL: 'https://${account.namespace}',
   // clientID - This is your Auth0 Client ID found in your Auth0 Application Settings
   clientID: '${account.clientId}',
   // clientSecret - This is your Auth0 Client Secret found in your Auth0 Application Settings
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -72,19 +72,21 @@ app.use(auth({
   issuerBaseURL: 'https://${account.namespace}',
   clientID: '${account.clientId}'
 }));
-
-app.get('/', (req, res) => {
-  if (req.openid.isAuthenticated()) {
-    // when a user is logged in you can access the user object via `req.openid.user`
-    res.send(JSON.stringify(req.openid.user));
-  } else {
-    res.send('No user is logged in');
-  }
-})
 ```
 
-## Protect a Route
-Now that your application has the required middleware to integrate Auth0 in your application, you can start protecting routes.
+Now that your application has the middleware installed, you can use [various helpers and values](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the `express-openid-connect` library.  Here is an example of using `req.openid.isAuthenticated`
+
+```js
+app.get('/', (req, res) => {
+ res.send('You are logged ' + req.openid.isAuthenticated() ? 'In' : 'Out')
+});
+```
+
+## Login
+To log a user in visit the `/login` route provided by the library, or if you are testing your application locally [`localhost:3000/login`](http://localhost:3000/)
+
+## Profile
+To retrieve the user profile you will first want to protect the route where you display that users information.
 
 Add the `requiresAuth` middleware supplied from `express-openid-connect` for routes that require authentication.  Any route using this middleware will check for a valid user session and, if one does not exist, it will redirect the user to log in.
 
@@ -97,13 +99,8 @@ app.get('/profile', requiresAuth(), (req, res) => {
 });
 ```
 
-## Login
-To log a user in visit the `/login` route registered by the `express-oidc-connect.auth` middleware, 
-or if you are testing your application locally [`localhost:3000/login`](http://localhost:3000/)
-
 ## Logout
-To log a user out visit the `/logout` route registered by the `express-oidc-connect.auth` middleware, 
-or if you are testing your application locally [`localhost:3000/logout`](http://localhost:3000/)
+To log a user out visit the `/logout` route provided by the library, or if you are testing your application locally [`localhost:3000/logout`](http://localhost:3000/)
 
 ## What's next?
 This is an Early Access version of the Auth0 Express OIDC library.  You can further explore this library and its configuration options on [GitHub](https://github.com/auth0/express-openid-connect).
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -24,7 +24,7 @@ github:
 Install the following dependencies
 
 - [`express-session`](https://github.com/auth0/express-session) - simple session middleware for Express (any session middleware for express will work).
-- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - an [OIDC](/protocols/oidc) compliant authentication library.
+- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
 
 ```sh
 # Using NPM
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -141,7 +141,7 @@ If you are testing your application locally, you can login by visiting `http://l
 Once a user is logged in it is attached to the `request` via the `req.openid` object. You can access the full user object via `req.openid.user`
 
 ### Logout
-To log a user out send a `GET` request to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
+To log a user out, redirect them to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
 
 ::: note
 If you are testing your application locally, you can logout by visiting `http://localhost:3000/logout`
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -129,7 +129,7 @@ app.get('/profile', requiresAuth(), (req, res) => {
 ## Login and Logout
 ### Login
 To log a user in you have 3 options:
-- Send a `GET` request to the `/login` route registered by the `express-oidc-connect.auth` middleware.
+- Redirect to the `/login` route registered by the `express-oidc-connect.auth` middleware.
 - Use the `express-oidc-connect.requiresAuth` middleware to protect a route, as described above.
 - When initializing the `express-oidc-connect.auth` middleware, pass in `required: true` to force authentication on all routes.
 
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -120,7 +120,7 @@ Add the `requiresAuth` middleware supplied from `express-openid-connect` for rou
 ```js
 const { requiresAuth } = require('express-openid-connect');
 
-app.get('/logged_in_user', requiresAuth(), (req, res) => {
+app.get('/profile', requiresAuth(), (req, res) => {
   // when a user is logged in you can access the user object via `req.openid.user`
   res.send(JSON.stringify(req.openid.user));
 });
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -43,7 +43,7 @@ app.use(express.urlencoded({
 ```
 
 ### User sessions
-In order to keep the user logged in across requests you will need to use session middleware.  The session will allow you to recognize the user on concurrent requests. In this example, you will use `express-session` which stores the session in memory.
+To keep the user logged in across multiple requests, your application will use the `express-session` middleware, which stores the user's data in memory. The user's session will allow your application to recognize them from request to request.
 
 ```js
 const session = require('express-session');
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -27,11 +27,7 @@ Install the following dependencies
 - [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
 
 ```sh
-# Using NPM
 npm install express-openid-connect express-session
-
-# Using Yarn
-yarn add express-openid-connect express-session
 ```
 
 ### Handling server responses
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -26,7 +26,7 @@ If you are running this project on your local machine, your application's callba
 <%= include('../../../_includes/_logout_url' }) %>
 
 :::note
-If you are running this project on your local machine, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `http://localhost:3000/logout`.
+If you are running this project on your local machine, your application's logout URL is `http://localhost:3000`
 :::
 
 ## Integrate Auth0
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -95,6 +95,7 @@ app.get('/profile', requiresAuth(), (req, res) => {
   // when a user is logged in you can access the user object via `req.openid.user`
   res.send(JSON.stringify(req.openid.user));
 });
+```
 
 ## Login
 To log a user in visit the `/login` route registered by the `express-oidc-connect.auth` middleware, 
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -130,7 +130,7 @@ app.get('/profile', requiresAuth(), (req, res) => {
 ### Login
 To log a user in you have 3 options:
 - Send a `GET` request to the `/login` route registered by the `express-oidc-connect.auth` middleware.
-- Use the `express-oidc-connect.requiresAuth` middleware to protect a route.
+- Use the `express-oidc-connect.requiresAuth` middleware to protect a route, as described above.
 - When initializing the `express-oidc-connect.auth` middleware pass in `required: true` to force required authentication on all routes.
 
 ::: note
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -83,6 +83,19 @@ app.get('/', (req, res) => {
 })
 ```
 
+## Protect a Route
+Now that your application has the required middleware to integrate Auth0 in your application, you can start protecting routes.
+
+Add the `requiresAuth` middleware supplied from `express-openid-connect` for routes that require authentication.  Any route using this middleware will check for a valid user session and, if one does not exist, it will redirect the user to log in.
+
+```js
+const { requiresAuth } = require('express-openid-connect');
+
+app.get('/profile', requiresAuth(), (req, res) => {
+  // when a user is logged in you can access the user object via `req.openid.user`
+  res.send(JSON.stringify(req.openid.user));
+});
+
 ## Login
 To log a user in visit the `/login` route registered by the `express-oidc-connect.auth` middleware, 
 or if you are testing your application locally [`localhost:3000/login`](http://localhost:3000/)
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -74,7 +74,7 @@ Now that your application has the middleware installed, you can use [various hel
 
 ```js
 app.get('/', (req, res) => {
- res.send('You are logged ' + req.openid.isAuthenticated() ? 'In' : 'Out')
+ res.send(req.isAuthenticated() ? 'Logged in' : 'Logged out'))
 });
 ```
 
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -17,7 +17,7 @@ github:
 
 <!-- markdownlint-disable MD002 -->
 ## Getting Started
-In this quickstart you will learn how to implement authentication in an Express application using the OIDC Compliant library [Express OIDC](https://github.com/auth0/express-openid-connect) authored and maintained by Auth0.
+In this quickstart you will learn how to implement authentication in an Express application using the [Express OpenID Connect](https://github.com/auth0/express-openid-connect) library, authored and maintained by Auth0.
 
 ::: warning
 Express OIDC is still in an [early access stage](/).
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -16,6 +16,8 @@ github:
 ---
 
 <!-- markdownlint-disable MD002 -->
+<!-- markdownlint-disable MD034 -->
+
 ## Getting Started
 In this quickstart you will learn how to implement authentication in an Express application using the [Express OpenID Connect](https://github.com/auth0/express-openid-connect) library, authored and maintained by Auth0.
 
@@ -34,56 +36,28 @@ yarn add express-openid-connect express-session
 ```
 
 ## Configure Auth0
-### Configure Callback URLs
-
-A callback URL is a URL in your application where Auth0 redirects the user after they have authenticated. 
-
-The callback URL for your app must be whitelisted in the **Allowed Callback URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, users will be unable to log in to the application and will get an error.
-
-::: note
-If you are testing your application locally, you should set the **Allowed Callback URL** to `http://localhost:3000/callback`
-:::
-
-### Configure Logout URLs
-
-A logout URL is a URL in your application that Auth0 can return to after the user has been logged out of the authorization server. This is specified in the `returnTo` query parameter.
-
-The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, users will be unable to log out from the application and will get an error.
-
-::: note
-If you are testing your application locally, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `http://localhost:3000`.
-:::
+<%= include('../../../_includes/_callback_url') %>
+<%= include('../../../_includes/_logout_url' }) %>
 
 ## Integrate Auth0
-### Auth0 Application Keys
-
-You will need to retrieve some details for your application to communicate with Auth0. You can get these details from the [Application Settings](${manage_url}/#/applications/${account.clientId}/settings) section in the Auth0 dashboard.
-
-You need the following information:
-
-* **Domain**
-* **Client ID**
-* **Client Secret**
-
 ### Express Middleware
 There are three middlewares required to integrate Auth0 in your application.
 
 - `express.urlencoded` - parses urlencoded payloads sent back from the Auth0 server.
-- `express-session` - simple session middleware for Express (other session middleware can be used as a replacement).
-- `express-openid-connect` - an [OIDC](/protocols/oidc) compliant authentication middleware.
+- `express-session` - simple session middleware for Express (this can be swapped with any session middleware for express).
+- `express-openid-connect.auth` - an [OIDC](/protocols/oidc) compliant authentication middleware.
 
 ::: note
 To learn more about using middleware with Express view the [official documentation](https://expressjs.com/en/guide/using-middleware.html)
-:::
-
-::: note
-Secrets should never be checked into source control.  We recommend using environment variables in place of plain text.  You can use [Dotenv](https://github.com/motdotla/dotenv) to easily manage envrionment variables on your localhost.
 :::
 
 ```js
 const express = require('express');
 const session = require('express-session');
 const { auth } = require('express-openid-connect');
+
+// Example initialization of Express
+const app = express();
 
 // parses urlencoded payloads sent back from the Auth0 server
 app.use(express.urlencoded({
@@ -97,18 +71,12 @@ app.use(session({
   saveUninitialized: false
 }));
 
-// express-openid-connect.auth 
+// express-openid-connect.auth
 app.use(auth({
-  // required - forces authentication for all routes
   required: false,
-  // baseURL - There is where your /callback, /login & /logout URLs will be registered
   baseURL: 'http://localhost:3000',
-  // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
   issuerBaseURL: 'https://${account.namespace}',
-  // clientID - This is the Client ID found in your Auth0 Application settings
-  clientID: '${account.clientId}',
-  // clientSecret - This is your Auth0 Client Secret found in your Auth0 Application Settings
-  clientSecret: '${account.clientSecret}'
+  clientID: '${account.clientId}'
 }));
 ```
 
@@ -136,9 +104,6 @@ To log a user in you have 3 options:
 ::: note
 If you are testing your application locally, you can login by visiting `http://localhost:3000/login`
 :::
-
-### Logged in Users
-Once a user is logged in it is attached to the `request` via the `req.openid` object. You can access the full user object via `req.openid.user`
 
 ### Logout
 To log a user out, redirect them to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -24,7 +24,7 @@ github:
 Install the following dependencies
 
 - [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
-- [`express-session`](https://github.com/auth0/express-session) - simple session middleware for Express (any session middleware for express will work).
+- [`express-session`](https://github.com/auth0/express-session) - session middleware for Express.
 
 ```sh
 npm install express-openid-connect express-session
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -99,7 +99,7 @@ app.get('/profile', requiresAuth(), (req, res) => {
 To log a user out visit the `/logout` route provided by the library, or if you are testing your application locally [`localhost:3000/logout`](http://localhost:3000/)
 
 ## What's next?
-The Auth0 team has put together a few different examples of using [`express-openid-connect`](https://github.com/auth0/express-openid-connect)
+We put together a few examples of how to use [Express OpenID Connect](https://github.com/auth0/express-openid-connect) in more advanced use cases:
 * [Route Customization](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#2-route-customization)
 * [Using refresh tokens](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#4-using-refresh-tokens)
 * [Require auth for specific routes](https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md#3-require-auth-for-specific-routes)
--- a/articles/quickstart/webapp/nodejs-ea/01-login.md
+++ b/articles/quickstart/webapp/nodejs-ea/01-login.md
@@ -92,7 +92,7 @@ app.get('/', (req, res) => {
 A user can log into your application by visiting the `/login` route provided by the library. If you are testing your application locally, that link is [`localhost:3000/login`](http://localhost:3000/login).
 
 ## Profile
-To retrieve the user profile you will first want to protect the route where you display that users information.
+To display the user's profile, your application should provide a protected route.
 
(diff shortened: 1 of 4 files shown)


articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  - nodejs
+contentType: tutorial
+useCase: quickstart
+github:
+  path: 01-Login
+---
+
+<!-- markdownlint-disable MD002 -->
+## Getting Started
+In this quickstart you will learn how to implement authentication in an Express application using the OIDC Compliant library [Express OIDC](https://github.com/auth0/express-openid-connect) authored and maintained by Auth0.
```
joshcanhelp:
```suggestion
In this quickstart you will learn how to implement authentication in an Express application using the [Express OpenID Connect](https://github.com/auth0/express-openid-connect) library, authored and maintained by Auth0.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+::: warning
+Express OIDC is still in an [early access stage](/).
```
joshcanhelp: @albertoperdomo - Should we use generally the same messaging here as [on the repo](https://github.com/auth0/express-openid-connect#express-openid-connect)?

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+```sh
+# Using NPM
+npm install express-openid-connect express-session
+
+# Using Yarn
+yarn add express-openid-connect express-session
+```
+
+## Configure Auth0
```
joshcanhelp: I believe these should use includes like the [Passport one](https://raw.githubusercontent.com/auth0/docs/master/articles/quickstart/webapp/nodejs/01-login.md)

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+A logout URL is a URL in your application that Auth0 can return to after the user has been logged out of the authorization server. This is specified in the `returnTo` query parameter.
+
+The logout URL for your app must be whitelisted in the **Allowed Logout URLs** field in your [Application Settings](${manage_url}/#/applications/${account.clientId}/settings). If this field is not set, users will be unable to log out from the application and will get an error.
+
+::: note
+If you are testing your application locally, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `http://localhost:3000`.
+:::
+
+## Integrate Auth0
```
joshcanhelp, line 32: Same here, should probably use includes ... and should probably be above the last section

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+### Auth0 Application Keys
+
+You will need to retrieve some details for your application to communicate with Auth0. You can get these details from the [Application Settings](${manage_url}/#/applications/${account.clientId}/settings) section in the Auth0 dashboard.
+
+You need the following information:
+
+* **Domain**
+* **Client ID**
+* **Client Secret**
```
joshcanhelp: We want to leave this out for now ... more explanation below.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Express Middleware
+There are three middlewares required to integrate Auth0 in your application.
+
+- `express.urlencoded` - parses urlencoded payloads sent back from the Auth0 server.
+- `express-session` - simple session middleware for Express (other session middleware can be used as a replacement).
```
joshcanhelp: I don't think this is technically a middleware. Also, we probably just want to call out that there needs to be session handling of some kind. They can use `express-session` (stores in memory ... there was a conversation at some point internally about how this might not be ideal) or `cookie-session` or another session hander. We'll need to pick one for this QS (and be confident we're choosing the right one) but I think we should make it clear that either will work.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+- `express-openid-connect` - an [OIDC](/protocols/oidc) compliant authentication middleware.
```
joshcanhelp: Also not technically a middleware (it provides middlewares).

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+::: note
+Secrets should never be checked into source control.  We recommend using environment variables in place of plain text.  You can use [Dotenv](https://github.com/motdotla/dotenv) to easily manage envrionment variables on your localhost.
+:::
+
+```js
+const express = require('express');
+const session = require('express-session');
+const { auth } = require('express-openid-connect');
+
```
joshcanhelp, line 49: I would actually skip the whole "middleware" summary above and parse this chunk of code out into smaller snippets with explanation in between.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+- `express.urlencoded` - parses urlencoded payloads sent back from the Auth0 server.
+- `express-session` - simple session middleware for Express (other session middleware can be used as a replacement).
+- `express-openid-connect` - an [OIDC](/protocols/oidc) compliant authentication middleware.
+
+::: note
+To learn more about using middleware with Express view the [official documentation](https://expressjs.com/en/guide/using-middleware.html)
+:::
+
+::: note
```
joshcanhelp: This is a good note to have but ... we've got "secrets" in the code below. In other QSes, we say something like this but integrate our recommended library into the QS.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+You need the following information:
+
+* **Domain**
+* **Client ID**
+* **Client Secret**
+
+### Express Middleware
+There are three middlewares required to integrate Auth0 in your application.
+
+- `express.urlencoded` - parses urlencoded payloads sent back from the Auth0 server.
```
joshcanhelp: Might want to mention that `bodyParser.urlencoded` works the same and might already be in use in an application. It's also the way to get this working with a lower version of Express.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+```js
+const express = require('express');
+const session = require('express-session');
+const { auth } = require('express-openid-connect');
+
+// parses urlencoded payloads sent back from the Auth0 server
+app.use(express.urlencoded({
+  extended: false
+}));
+
```
joshcanhelp, line 60: Split

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+// session middleware required to handle user sessions
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+
```
joshcanhelp, line 73: Split

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+// express-openid-connect.auth 
+app.use(auth({
+  // required - forces authentication for all routes
+  required: false,
+  // baseURL - There is where your /callback, /login & /logout URLs will be registered
+  baseURL: 'http://localhost:3000',
+  // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
+  issuerBaseURL: '${account.namespace}',
```
joshcanhelp:
```suggestion
  issuerBaseURL: 'https://${account.namespace}',
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  // clientID - This is your Auth0 Client ID found in your Auth0 Application Settings
+  clientID: '${account.clientId}',
+  // clientSecret - This is your Auth0 Client Secret found in your Auth0 Application Settings
+  clientSecret: '${account.clientSecret}'
+}));
+```
+
```
joshcanhelp, line 82: This is probably a good place to call out that these values can be stored in `process.env` ([ref](https://github.com/auth0/express-openid-connect#getting-started)) and picked up automatically in the library. It would cut down on the amount of code that's here.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+// session middleware required to handle user sessions
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+
+// express-openid-connect.auth 
+app.use(auth({
+  // required - forces authentication for all routes
```
joshcanhelp: The current setting of `false` does the opposite of forcing auth

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  required: false,
+  // baseURL - There is where your /callback, /login & /logout URLs will be registered
+  baseURL: 'http://localhost:3000',
+  // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
```
joshcanhelp: It's actually the URL version of that ... domain pre-pended with `https:`

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+
+// express-openid-connect.auth 
+app.use(auth({
+  // required - forces authentication for all routes
+  required: false,
+  // baseURL - There is where your /callback, /login & /logout URLs will be registered
```
joshcanhelp: Confusing explanation here .... also, most of our quickstarts keep the code pretty clean and explain the values below. This makes it easy to copy/paste into an app without much modification.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  baseURL: 'http://localhost:3000',
+  // issuerBaseURL - This is your Auth0 Domain found in your Auth0 Application Settings
+  issuerBaseURL: '${account.namespace}',
+  // clientID - This is your Auth0 Client ID found in your Auth0 Application Settings
```
joshcanhelp:
```suggestion
  // clientID - This is the Client ID found in your Auth0 Application settings
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  clientID: '${account.clientId}',
+  // clientSecret - This is your Auth0 Client Secret found in your Auth0 Application Settings
+  clientSecret: '${account.clientSecret}'
```
joshcanhelp: Leave this out as well

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+}));
+```
+
+## Protect a Route
+Now that your application has the required middleware to integrate Auth0 in your application, you can start protecting routes.
+
+Add the `requiresAuth` middleware supplied from `express-openid-connect` for routes that need protection.  Any time a route using this middleware is hit, it will check for a valid user session, if none exists it will ask the user to log in.
```
joshcanhelp:
```suggestion
Add the `requiresAuth` middleware supplied from `express-openid-connect` for routes that require authentication.  Any route using this middleware will check for a valid user session and, if one does not exist, it will redirect the user to log in.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+```js
+const { requiresAuth } = require('express-openid-connect');
+
+app.get('/logged_in_user', requiresAuth(), (req, res) => {
```
joshcanhelp:
```suggestion
app.get('/profile', requiresAuth(), (req, res) => {
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  // when a user is logged in you can access the user object via `req.openid.user`
+  res.send(JSON.stringify(req.openid.user));
```
joshcanhelp: 😎

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+});
+```
+
+## Login and Logout
+### Login
+To log a user in you have 3 options:
+- Send a `GET` request to the `/login` route registered by the `express-oidc-connect.auth` middleware.
```
joshcanhelp:
```suggestion
- Redirect to the `/login` route registered by the `express-oidc-connect.auth` middleware.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+- Use the `express-oidc-connect.requiresAuth` middleware to protect a route.
+- When initializing the `express-oidc-connect.auth` middleware pass in `required: true` to force required authentication on all routes.
```
joshcanhelp:
```suggestion
- When initializing the `express-oidc-connect.auth` middleware, pass in `required: true` to force authentication on all routes.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+  // when a user is logged in you can access the user object via `req.openid.user`
+  res.send(JSON.stringify(req.openid.user));
+});
+```
+
+## Login and Logout
+### Login
+To log a user in you have 3 options:
+- Send a `GET` request to the `/login` route registered by the `express-oidc-connect.auth` middleware.
+- Use the `express-oidc-connect.requiresAuth` middleware to protect a route.
```
joshcanhelp: Maybe move this to the top of the list also.

```suggestion
- Use the `express-oidc-connect.requiresAuth` middleware to protect a route, as described above.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+- When initializing the `express-oidc-connect.auth` middleware pass in `required: true` to force required authentication on all routes.
+
+::: note
+If you are testing your application locally, you can login by visiting `http://localhost:3000/login`
```
joshcanhelp: This is handy but I would combine it with the bullet point above.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+:::
+
+### Logged in Users
+Once a user is logged in it is attached to the `request` via the `req.openid` object. You can access the full user object via `req.openid.user`
```
joshcanhelp: 🤔 ... I'd reword this a bit. Using `request` here is confusing. Maybe something like:

> The user's profile is always stored in `req.openid.user`

This makes me wonder ... should we add some kind of helper like `req.openid.isAuthenticated()` to the SDK that returns boolean based on if `user` is empty or not?

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Logout
+To log a user out send a `GET` request to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
```
joshcanhelp:
```suggestion
To log a user out, redirect them to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
```

joshcanhelp (changes requested):
Great work on this so far!!

There are a few things to address here but we're on the right path for sure. 

I mentioned removing the Client Secret ... in this case, the SDK is using the `form_post` response mode so no client secret is needed 👍

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: In our user research we discovered people skipping this section in search for the first code block or how to install to get started.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: Yes I noticed `cookie-session` was used in the internal doc, but used the more widely used library `express-session`.  We should pick one and stick with it.  Don't need to worry the user with choice here.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: what is the benefit of having smaller code snippets?

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: User research was showing that users wanted to handle secret management at a different stage and outside of the quickstart.  This is more of a PSA than anything.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: Passing `false` will make it not required, is that not clear?

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: We could reword this, but inline comments don't change the difficulty of copy/pasting the code.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: Do we have more documentation on that?  What version of express does it stop being required?

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: I wonder if this section was even needed, it is demonstrated how to access the user object in the protect a route section

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: They do if you don't want the comments in your final product :)

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: The purpose of putting it is in the note is to stay consistent with the Logged out user section, and we seem to use this format when informing the user of nice to knows or shortcuts.  If this note seems redundant, I think it makes more sense to remove it all together

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: Interesting! Well then, by golly, let's use the order that works!

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: The `urlencoded` always needs to be there but Express 4.16 added it to `express`. Before that version, you'd need to install `body-parser` and use that.

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: I agree ... but I want to make sure we're not *mandating* `express-session`. Cookie session works fine as well. Point being: there needs to be a session handler and the one we choose here is A or B. Just like env management in that sense.

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: It's more step-by-step, rather than a big chunk of explanation then a big chunk of code.

To be totally honest, though, maybe that's not what helps people the most.

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: As a setting, yes, but in this context, it sounds like you're saying that the value below forces auth, which it doesn't.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: The quickstart and the library call out that Express 4.16+ is required

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: You could be right. 

Also ... the method I mentioned already exists :)

https://github.com/auth0/express-openid-connect/blob/master/lib/context.js#L26-L28

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: > but I think we should make it clear that either will work.

the wording `other session middleware can be used as a replacement` is possibly poor, but this was the intent

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: I say we work it into smaller chunks.  Just wanted some clarification on the motive

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: As mentioned below, maybe comments in the code should be left for samples.

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: True but both work and they might already be using body parser. I don't know if it hurts to have both in there or not but it might be worth it just to say that both are acceptable. Or maybe thats TMI?

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: Interesting ... so we're thinking that the environment stuff should be left out, besides a note like this? 

I think I'm biased because the whole "environment storage" was such a catch point for me early on in my dev experience. The big downside of including it in the QS is, of course, that this is duplicated across all QSes in this platform.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: I was going back and forth on the amount of code versus too much magic.  The library offers these options,  but I was worrying that it introduces too much magic too soon with too little understanding.  For example, `issuerBaseURL` if this doesn't change for the user, what is the point of using `process.env` for it?  

I'm not against moving these to `process.env`, as it definitely cleans up the amount of code needed, but it introduces an additional dependency for this tutorial.

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: @joshcanhelp lets revisit this after I chunk out this section

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: The only reason for the note was because of sentiment shared internally that we need to signal to the user best practices in security, more specifically not to check secrets into the source code.  

Also discussed here https://github.com/auth0/docs/pull/8355/files/cd66335815abeb2f2bb399299c82f6408e5432cf#r333791381 possibly it is best to show the user how to manage secrets, but it does add an additional dependency and widens scope of what the user is doing in this tutorial

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: I was trying to avoid all the boilerplate that comes loaded in those partials.  But addressing the length/brevity of that content could be addressed separately.

joshcanhelp: Fair enough. As long as it was intentional 👍

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: Apologies, I did not explain that very clearly. 

The `express-oidc-connect` library can pick up it's required config values from the environment automatically. [README has instructions on that](https://github.com/auth0/express-openid-connect#getting-started). So this QS would not get a new dependency, but it would be a bit of black magic and start to get into env vars. 

Since we're not using a client secret, configuring in the code is maybe not ideal but also not disastrous. All of these values are public (appear in the URL). So maybe leave as-is for simplicity and add a note like:

> For other configuration options and methods see the [docs](https://github.com/auth0/express-openid-connect/blob/master/API.md)

articles/quickstart/webapp/nodejs-ea/01-login.md
joshcanhelp: I'm on the fence. The phrase:

> Redirect to the `/login` route registered by the `express-oidc-connect.auth` middleware.

... is accurate and clear to *me* but is that maybe too technical? I don't think it would hurt to add after that something like:

> You can login to this local application at [localhost:3000/login](http://localhost:3000/login).

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,116 @@
...
+// session middleware required to handle user sessions
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+
+// express-openid-connect.auth
+app.use(auth({
+  required: false,
```
joshcanhelp, line 75: I meant to ask earlier ... what do you think about this? The SDK is default "require auth for all routes." Do you think that's the most common case or do you think that applications would only need specific routes protected? If this default was switched, it feels like the DX is nicer as well:

```js
////
/// Auth required everywhere
//
const { auth } = require('express-openid-connect');

app.use(auth({ 
    required: true 
});

app.get('/profile', (req, res) => {
    // ...
});

////
/// Auth required on specific routes
//
const { auth, requiresAuth } = require('express-openid-connect');
app.use(auth());
app.get('/profile', requiresAuth(), (req, res) => {
    // ...
});
```

joshcanhelp: It's a fairly minor point but changing this is a pretty big breaking change and one I'd like to get in before EA if we decide to go that direction).

joshcanhelp: Going to punt this to beta. Have not had any complaints or comments about this, just something I was thinking about.

https://auth0team.atlassian.net/browse/SDK-1071

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+
+<%= include('../../../_includes/_new_app', { hideDownloadSample: true, hideDashboardScreenshot: true }) %>
+<%= include('../../../_includes/_callback_url') %>
+<%= include('../../../_includes/_logout_url' }) %>
+
+## Integrate Auth0
+### Install Dependencies
+Install the following dependencies
+
+- [`express-session`](https://github.com/auth0/express-session) - simple session middleware for Express (any session middleware for express will work).
```
joshcanhelp: Let's switch this to `cookie-session`. Simpler to use and less "gotchas." Similar philosophy to why we're going to skip secret management. A few instances below as well.

joshcanhelp: Also ... it looks like the only two options are `cookie-session` and `express-session` so maybe just call out the ability to use `express-session` but may need additional configuration considerations.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - an [OIDC](/protocols/oidc) compliant authentication library.
```
joshcanhelp:
```suggestion
- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+```sh
+# Using NPM
+npm install express-openid-connect express-session
+
+# Using Yarn
+yarn add express-openid-connect express-session
+```
+
+### Handling server responses
+Your application will need to parse url-encoded payloads sent back from the Auth0 server.  Express provides a middleware for this `express.urlencoded`. You might already be using `body-parser`, this will work too.
```
joshcanhelp:
```suggestion
Your application will need to parse URL-encoded data sent back from the Auth0 server.  Express provides a middleware for this called `express.urlencoded`. If you are integrating an existing application that uses `urlencoded ` from the `body-parser`module, that will work as well.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+const express = require('express');
+const app = express();
+
+app.use(express.urlencoded({
+  extended: false
+}));
+```
+
+### User sessions
+In order to keep the user logged in across requests you will need to use session middleware.  The session will allow you to recognize the user on concurrent requests. In this example, you will use `express-session` which stores the session in memory.
```
joshcanhelp: `cookie-session` update here as well.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+```
+
+### Authentication
+The last part of integrating Auth0 into your application will be using the `auth` middleware supplied to you from the Auth0 library `express-opeinid-connect`.  View the official [Api Documentation](https://github.com/auth0/express-openid-connect/blob/4374c0502d4aedf8b3975e7c5f26929b305b32f6/API.md) for configuration options and information.
```
joshcanhelp:
```suggestion
The last part of integrating Auth0 into your application will be using the `auth` middleware provided by Auth0's `express-openid-connect` installed above.  View the official [API Documentation](https://github.com/auth0/express-openid-connect/blob/master/API.md) for configuration options and information.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+
+::: note
+If you are testing your application locally, you can login by visiting `http://localhost:3000/login`
+:::
+
+### Logout
+To log a user out, redirect them to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
+
+::: note
+If you are testing your application locally, you can logout by visiting `http://localhost:3000/logout`
```
joshcanhelp:
```suggestion
If you are testing your application locally, you can logout by visiting [`localhost:3000/logout`](http://localhost:3000/logout)
```

I would combine this with the sentence above instead of calling it out (IMHO)

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+```js
+const { requiresAuth } = require('express-openid-connect');
+
+app.get('/profile', requiresAuth(), (req, res) => {
+  // when a user is logged in you can access the user object via `req.openid.user`
+  res.send(JSON.stringify(req.openid.user));
+});
+```
+
+## Login and Logout
```
joshcanhelp: Maybe instead of a main section header for both, Login and Logout get their own sections.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+:::
+
+### Logout
+To log a user out, redirect them to the `/logout` route registered by the `express-oidc-connect.auth` middleware.
+
+::: note
+If you are testing your application locally, you can logout by visiting `http://localhost:3000/logout`
+:::
+
+## What's next?
```
joshcanhelp, line 110: It feels like we should call out specific developer tasks here, like what is in EXAMPLES:

https://github.com/auth0/express-openid-connect/blob/master/EXAMPLES.md

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,110 @@
...
+### Login
+To log a user in you have 3 options:
+- Redirect to the `/login` route registered by the `express-oidc-connect.auth` middleware.
+- Use the `express-oidc-connect.requiresAuth` middleware to protect a route, as described above.
+- When initializing the `express-oidc-connect.auth` middleware, pass in `required: true` to force authentication on all routes.
+
+::: note
+If you are testing your application locally, you can login by visiting `http://localhost:3000/login`
+:::
+
```
joshcanhelp: I think we should add a **Profile** or **Get User Data** section here to call out `req.openid.user`. Maybe that's just a matter of renaming "Protect a Route" above and moving it here?

joshcanhelp (changes requested):
This is looking **so** clean. This QS is so short, it looks like things are missing but, shockingly, there aren't! This really _sells_ the SDK 😉

articles/quickstart/webapp/nodejs-ea/01-login.md
davidpatrick: pretty small sections, but we let's test it

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+  - logout
+  - nodejs
+contentType: tutorial
+useCase: quickstart
+github:
+  path: 01-Login
+---
+
+<%= include('../../../_includes/_new_app', { hideDownloadSample: true, hideDashboardScreenshot: true }) %>
+<%= include('../../../_includes/_callback_url') %>
```
joshcanhelp, line 20: We should tell them what callback URL they're using (`/callback`) or they won't know.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+<%= include('../../../_includes/_logout_url' }) %>
```
joshcanhelp, line 26: We should tell them what logout URL they're using (`/` I believe) or they won't know.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Integrate Auth0
+### Install Dependencies
+Install the following dependencies
+
+- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
+- [`express-session`](https://github.com/auth0/express-session) - simple session middleware for Express (any session middleware for express will work).
```
joshcanhelp:
```suggestion
- [`express-session`](https://github.com/auth0/express-session) - session middleware for Express.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+const express = require('express');
+const app = express();
+
+app.use(express.urlencoded({
+  extended: false
+}));
+```
+
+### User sessions
+In order to keep the user logged in across requests you will need to use session middleware.  The session will allow you to recognize the user on concurrent requests. In this example, you will use `express-session` which stores the session in memory.
```
joshcanhelp:
```suggestion
To keep the user logged in across multiple requests, your application will use the `express-session` middleware, which stores the user's data in memory. The user's session will allow your application to recognize them from request to request.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+```
+
+### Authentication
+The last part of integrating Auth0 into your application will be using the `auth` middleware provided by Auth0's `express-openid-connect` installed above.  View the official [API Documentation](https://github.com/auth0/express-openid-connect/blob/master/API.md) for configuration options and information.
```
joshcanhelp:
```suggestion
The last part of integrating Auth0 into your application will be using the `auth` middleware provided by the Express OpenID Connect library installed above. The [API documentation](https://github.com/auth0/express-openid-connect/blob/master/API.md) for this library outlines additional configuration options and methods.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+app.use(auth({
+  required: false,
+  auth0Logout: true,
+  baseURL: 'http://localhost:3000',
+  issuerBaseURL: 'https://${account.namespace}',
+  clientID: '${account.clientId}'
+}));
+```
+
+Now that your application has the middleware installed, you can use [various helpers and values](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the `express-openid-connect` library.  Here is an example of using `req.openid.isAuthenticated`
```
joshcanhelp:
```suggestion
Now that your application has the middleware installed, you can use [the helper functions and contextual data](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the Express OpenID Connect library.  Here is an example of how to use `req.openid.isAuthenticated()`:
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+```js
+app.get('/', (req, res) => {
+ res.send('You are logged ' + req.openid.isAuthenticated() ? 'In' : 'Out')
+});
+```
+
+## Login
+To log a user in visit the `/login` route provided by the library, or if you are testing your application locally [`localhost:3000/login`](http://localhost:3000/)
```
joshcanhelp:
```suggestion
A user can log into your application by visiting the `/login` route provided by the library. If you are testing your application locally, that link is [`localhost:3000/login`](http://localhost:3000/login).
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Profile
+To retrieve the user profile you will first want to protect the route where you display that users information.
```
joshcanhelp:
```suggestion
To display the user's profile, your application should provide a protected route.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+Add the `requiresAuth` middleware for routes that require authentication.  Any route using this middleware will check for a valid user session and, if one does not exist, it will redirect the user to log in.
+
+```js
+const { requiresAuth } = require('express-openid-connect');
+
+app.get('/profile', requiresAuth(), (req, res) => {
+  // when a user is logged in you can access the user object via `req.openid.user`
```
joshcanhelp:
```suggestion
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+  res.send(JSON.stringify(req.openid.user));
+});
+```
+
+## Logout
+To log a user out visit the `/logout` route provided by the library, or if you are testing your application locally [`localhost:3000/logout`](http://localhost:3000/)
```
joshcanhelp:
```suggestion
A user can log out of your application by visiting the `/logout` route provided by the library. If you are testing your application locally, that link is [`localhost:3000/logout`](http://localhost:3000/logout).
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+## What's next?
+The Auth0 team has put together a few different examples of using [`express-openid-connect`](https://github.com/auth0/express-openid-connect)
```
joshcanhelp:
```suggestion
We put together a few examples of how to use [Express OpenID Connect](https://github.com/auth0/express-openid-connect) in more advanced use cases:
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+  issuerBaseURL: 'https://${account.namespace}',
+  clientID: '${account.clientId}'
+}));
+```
+
+Now that your application has the middleware installed, you can use [various helpers and values](https://github.com/auth0/express-openid-connect/blob/master/API.md#session-and-context) provided by the `express-openid-connect` library.  Here is an example of using `req.openid.isAuthenticated`
+
+```js
+app.get('/', (req, res) => {
+ res.send('You are logged ' + req.openid.isAuthenticated() ? 'In' : 'Out')
```
joshcanhelp:
```suggestion
 res.send(req.isAuthenticated() ? 'Logged in' : 'Logged out'))
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Integrate Auth0
+### Install Dependencies
+Install the following dependencies
+
+- [`express-openid-connect`](https://github.com/auth0/express-openid-connect) - Auth0-maintained OIDC-compliant library for Express.
+- [`express-session`](https://github.com/auth0/express-session) - simple session middleware for Express (any session middleware for express will work).
+
+```sh
+npm install express-openid-connect express-session
```
joshcanhelp:
```suggestion
npm install express express-openid-connect express-session
```

Oddly, neither package installs `express` with this command 🤷‍♂

joshcanhelp (changes requested):
Getting really close, this is looking great. 

What is missing is directing the developer to open a file, put these code samples in, and run it with `node app.js` or whatever. We'll need some kind of direction for that. Maybe a note about putting the code in a JS file, then quick instructions on how to run a Node app locally at the end? 

Also needs:

```js
require('http').createServer(app).listen(3000, () => console.log('Listening on http://localhost:3000'));
```

Even with that code add, this is a 27 line Node application with authentication 🎉  phenomenal!

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,114 @@
...
+github:
+  path: 01-Login
+---
+
+<%= include('../../../_includes/_new_app', { hideDownloadSample: true, hideDashboardScreenshot: true }) %>
+
+<%= include('../../../_includes/_callback_url') %>
+
+:::note
+If you are running this project on your local machine, the callback URL you need to whitelist in the **Allowed Callback URLs** field is `http://localhost:3000/callback`.
```
joshcanhelp:
```suggestion
If you are running this project on your local machine, your application's callback URL is `http://localhost:3000/callback`
```

joshcanhelp: I don't think we need to partially duplicate the information above.

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,114 @@
...
+:::
+
+<%= include('../../../_includes/_logout_url' }) %>
+
+:::note
+If you are running this project on your local machine, the logout URL you need to whitelist in the **Allowed Logout URLs** field is `http://localhost:3000/logout`.
```
joshcanhelp:
```suggestion
If you are running this project on your local machine, your application's logout URL is `http://localhost:3000`
```

joshcanhelp: This is the URL that users come back to after logging out. Application setting field is a bit misleading ... same feedback as the above RE: duplicating steps

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,114 @@
...
+:::
+
+## Integrate Auth0
+### Install Dependencies
+Install the following dependencies
```
joshcanhelp:
```suggestion
Your application will need the following packages:
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,114 @@
...
+const express = require('express');
+const app = express();
+
+app.use(express.urlencoded({
+  extended: false
+}));
+```
+
+### User sessions
+To keep the user logged in across multiple requests, your application will use the `express-session` middleware, which stores the user's data in memory. The user's session will allow your application to recognize them from request to request.
```
joshcanhelp:
```suggestion
To keep the user logged in across multiple requests, your application will use the `express-session` middleware to store the user's data in memory. The user's session will allow your application to recognize them from request to request.
```

articles/quickstart/webapp/nodejs-ea/01-login.md
```diff
@@ -0,0 +1,114 @@
...
+
+app.use(session({
+  secret: 'this should be a secret',
+  resave: true,
+  saveUninitialized: false
+}));
+```
+
+### Authentication
+The last part of integrating Auth0 into your application will be using the `auth` middleware provided by the Express OpenID Connect library installed above. The [API documentation](https://github.com/auth0/express-openid-connect/blob/master/API.md) for this library outlines additional configuration options and methods.
```
joshcanhelp:
```suggestion
The last part of integrating Auth0 into your application will be using the `auth` middleware provided by the Express OpenID Connect library installed above. The [API documentation](https://github.com/auth0/express-openid-connect/blob/master/API.md) for this library outlines additional configuration options and methods that can be used when initializing this middleware.
```

joshcanhelp (changes requested):
I think this is it .... this is so clean ...

joshcanhelp (approved):

PR #8363, opened 2019-10-11
[IPS-752] Add troubleshooting guide for SIWA

josh-cain:


tamigoodall (approved):

articles/connections/apple-siwa/troubleshooting.md
```diff
@@ -0,0 +1,58 @@
...
+If your application successfully initiated the login flow with Auth0, the results will be reflected in the tenant logs.  Native social exchanges will use the `sens` and `fens` event types to indicate success and failure (respectively), while web flows will use the standard `s` and `f` event types.  All tenant logs interacting with the Apple IdP will use the connection value of `apple`.
+
+## Types of errors
+
+The following errors may be returned from the Apple IdP.  Auth0 will relay both status codes and error messages from Apple should a request fail.
+
+| Error | Status Code | Description |
+| - | - | - |
+| `invalid_request` | 400 | The request parameters were incomplete or incorrect |
+| `invalid_grant` | 400 | The authorization code or refresh token presented to the Apple IdP is not valid |
```
CriGoT, line 41: This makes it sound like we are sending the refresh token to Apple. I would probably remove the reference to refresh tokens completely and clarify that this is specific to the Native social login exchange

articles/connections/apple-siwa/troubleshooting.md
```diff
@@ -0,0 +1,58 @@
...
+| `invalid_client` | 400 | Apple was unable to successfully authenticate the client with the provided credentials |
```
CriGoT, line 42: I'm not sure about this one. Looking at the tests we return a `server_error` when  we get an error from apple (the apple error is `invalid_client`.

josh-cain: True for now, not once https://github.com/auth0/auth0-server/pull/6728 gets in... was writing to that state.

articles/connections/apple-siwa/troubleshooting.md
josh-cain: I see what you're saying.  It *is* what Apple would send us in the event that happened, but TMI here could definitely cause confusion.  I'll take it out (on a subsequent PR).

PR #8397, opened 2019-10-22
Add info re: current_user scope rate limits

khorne3:
https://docs-content-staging-pr-8397.herokuapp.com/docs/api/management/v2/get-access-tokens-for-spas

https://docs-content-staging-pr-8397.herokuapp.com/docs/policies/rate-limits

### DIFF
--- a/articles/api/management/v2/get-access-tokens-for-spas.md
+++ b/articles/api/management/v2/get-access-tokens-for-spas.md
@@ -39,7 +39,7 @@ With a Management API Token issued for a SPA, you can access the following scope
 | `delete:current_user_device_credentials` | [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id) |
 
 ::: note
-Please note that the above scopes (and therefore limits) are rate limited.
+Please note that the above scopes (and therefore limits) are [rate limited](/policies/rate-limits#access-tokens-for-spas).
 :::
 
 ## Using a Management API Token to call the Management API from a SPA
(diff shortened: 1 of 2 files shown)


articles/api/management/v2/get-access-tokens-for-spas.md
```diff
@@ -38,6 +38,16 @@ With a Management API Token issued for a SPA, you can access the following scope
 | `create:current_user_device_credentials` | [POST /api/v2/device-credentials](/api/management/v2#!/Device_Credentials/post_device_credentials) |
 | `delete:current_user_device_credentials` | [DELETE /api/v2/device-credentials/{id}](/api/management/v2#!/Device_Credentials/delete_device_credentials_by_id) |
 
+### Rate Limits
+
+Please note that the above scopes (and therefore limits) are rate limited. By default, the rate limit is **disabled** (that is, the controlling tenant-level flag `dedicated_rate_limit_for_current_user_scope` is set to `false`). When `dedicated_rate_limit_for_current_user_scope = false`, Auth0 makes a log entry whenever the rate limit is exceeded:
```
siacomuzzi: flag and logs are internal only

panga: +1 please delete any flag references, the default value will be true

michaelwallen:
@khorne3 Thanks you this all looks great:) 👍

siacomuzzi:
@michaelwallen why are we not saying that the new rate limit is "10 requests per minute per user"?
IMO it's the first thing a customer will ask, right?

michaelwallen:
@siacomuzzi Good point. @khorne3 Your thoughts on adding this?

khorne3:
@siacomuzzi and @michaelwallen In the past, we have avoided providing specific numbers, but I would be happy to include this information if you would like. What do you think?

siacomuzzi:
I think we should provide the numbers (we already are doing it with the other limits: https://docs-content-staging-pr-8397.herokuapp.com/docs/policies/rate-limits)

khorne3:
@siacomuzzi Added rate limit info. Thanks!

panga (approved):

PR #8400, opened 2019-10-23
Add file paths to code snippets

### LINKED ISSUE #4671: cordova tutorial no snippets location
(AUTH-4142)
[Cordova tutorial](https://auth0.com/docs/quickstart/native/cordova#set-up-url-redirects) snippets have no file path/name is placed in the first line.
image

joshcanhelp:
Closes #4671

### DIFF
--- a/articles/quickstart/native/cordova/01-login.md
+++ b/articles/quickstart/native/cordova/01-login.md
@@ -11,6 +11,8 @@ github:
 contentType: tutorial
 useCase: quickstart
 ---
+<!-- markdownlint-disable MD002-->
+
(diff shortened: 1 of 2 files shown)


stevehobbsdev (approved):
In general, I've seen mixed feedback about having the filename as the first line in a code snippet. The downsides being that it's included in a copy/paste and it doesn't work everywhere (JSON files, for example). Would be good to standardize.

Not a discussion for here though! 😄

joshcanhelp:
@stevehobbsdev - Good point on that. The main benefit of the comment is consistency but I think we can achieve the same with text above like "Working in `src/index.js`:" Or similar

PR #8402, opened 2019-10-23
Update install-deploy-cli.md

tamigoodall:
Updated based on user feedback


rkhoriander:
After a review, there seem to be serious issues with the content on this page:

* It looks like it's code copied from a README somewhere? Pre-reqs section mentions this, but looks like it means to be referencing the "installing the tool manually" section below.
* "installing the tool manually" just tells how to create the app, not actually install the tool. So seems like a circular reference from pre-reqs section. And the steps are ancient/don't match what our dashboard currently does.
* Prereqs don't seem to be pre-reqs, or else the whole page is not matching the doc title.

rkhoriander (changes requested):
Needs a rewrite. Seems to be a misinterpretation of the actual steps and needs updates to match our current dashboard.

rkhoriander (approved):

PR #8419, opened 2019-10-25
Update 01-authorization.md

joshcanhelp:
Update from a customer in QS feedback. Pointed to the sample here and mentioned it was out of date. I copy-pasted what was in there to here with minor changes, not sure if this is correct.


lbalmaceda (changes requested):
Seems like this code is already currently there, just above the snippet were you added. Let's have this conversation via DM so I can understand what is wrong. Blocking for now.

PR #8439, opened 2019-10-30
Angular qs fixcallback

kmaida:
Removed callback from Angular quick start in order to make spinning up a server easier for the developer (if they are not using the existing server to test). Fixed a TS Lint warning in the authentication service, and cleaned up an unnecessary line of code.


stevehobbsdev (approved):
LGTM! Thanks @kmaida

PR #8444, opened 2019-10-30
fixing doc about invitation flow after the security fix

pietrorosa77:
After this fix
https://auth0team.atlassian.net/browse/SMADMIN-541 addressed by this PR https://github.com/auth0/manage/pull/4595 users are now forced to sign up / login with the same email as the one the invitation was sent to


rkhoriander:
@pietrorosa77 , what happens if the user is already logged in with an Auth0 account, and it is *not* the same email address to which the invitation was sent?

rkhoriander (approved):

PR #8454, opened 2019-10-31
Native SIWA: Enable the connection for the application

ziluvatar:


josh-cain (approved):
Good catch!

ziluvatar:
Link to page: https://docs-content-staging-pr-8454.herokuapp.com/docs/connections/apple-siwa/add-siwa-to-native-app#configure-and-enable-the-connection-in-auth0

PR #8475, opened 2019-11-07
Update Identity Labs / SPA / exercise-01.md

joshcanhelp:


lbalmaceda (approved):
For native clients at least, we normally use the same callback for both login and logout. Check if that's correct 👍

PR #8479, opened 2019-11-08
Apply feedback from express research

davidpatrick:
Applying feedback from express qs research sessions


joshcanhelp (approved):

joshcanhelp:
@davidpatrick - Just FYI, I reverted the change that removed `urlencoded` so we could merge this before the next release.

PR #8487, opened 2019-11-12
Make singular verbs plural

jhdoak:
This PR fixes the plurality of two verbs in the authorization code examples


joshcanhelp (approved):
@jhdoak - Thank you for the keen eye and the contribution!

PR #8495, opened 2019-11-14
Professional services page re factor with gallery

PeterGFernandez:


rkhoriander:
Closing stale PR per Peter, so Heroku can auto-generate review apps again.

PR #8530, opened 2019-11-27
Fix Swift Quickstart issues [QS-410]

Widcket:
- Limited the JavaScript code snippet to the relevant lines that the user would have to change, instead of including the whole default script. 
- Fixed a typo.
- The `Credentials` object mentions were properly cased to match the Objective C Quickstart.


stevehobbsdev (approved):
Great that you applied the Rule link here too!

PR #8538, opened 2019-11-30
Fixing API and SPA documentation issues

omar-compres:


stevehobbsdev (approved):
Verified that `yarn test` passes before merge.

PR #8552, opened 2019-12-06
Add scope param to first config object.

joshcanhelp:
Based on [feedback from community](https://community.auth0.com/t/php-crash-issue-when-reloading-link-the-following-day/34739).


damieng:
Why are the other parameters removed - are they the defaults?

joshcanhelp:
@damieng - No, just unnecessary for the quickstart

damieng (approved):

PR #8559, opened 2019-12-09
Add urge to use UL with libraries

rkhoriander:


albertoperdomo:
Looks good, but should we include a message stating that the new recommended SDK for SPA is auth0-spa-js @rkhoriander @erjohnson ?

jeffreylees (approved):

PR #8564, opened 2019-12-10
[CF-53] Update API1 deprecation docs

santiagoaguiar:
Updated API1 deprecation docs to reflect current API2 status.

* Active Directory Connector Monitoring: supported on API2.
* application users: no special case, use user search.
* rules-config: supported on API2
* email-templates: supported on API2
* User search on PSaaS Appliance: can use search v2 / v3.
* API v1-v2 Migration guide

https://docs-content-staging-pr-8564.herokuapp.com/docs/migrations/guides/management-api-v1-v2
https://docs-content-staging-pr-8564.herokuapp.com/docs/api/management/v1/use-cases
https://docs-content-staging-pr-8564.herokuapp.com/docs/api/management/v2/changes

https://auth0team.atlassian.net/browse/CF-53


articles/api/management/v1/use-cases.md
```diff
@@ -8,40 +8,24 @@ contentType: reference
...
 * [Enterprise Users/Directory Searching](#enterprise-users/directory-searching)
-* [Rules Configuration](#rules-configuration)
-* [Searching via the PSaaS Appliance](#searching-via-the-auth0-appliance)
 
 ## Active Directory Connector Monitoring
 
-In Management API v1, there is a `GET` endpoint that allows you to monitor the status of your Active Directory Connector:
+The following endpoint allows you to monitor the status of your Active Directory Connector. It is not currently available on the [Management API v2](/api/v2).
 
 GET `/api/connections/{AUTH0_CONNECTION}/socket`
```
panga: I think we have the exact same feature in APIv2: https://github.com/auth0/api2/blob/master/api/connections/routes/get_connection_status.js

articles/api/management/v1/use-cases.md
```diff
@@ -8,40 +8,24 @@ contentType: reference
...
-
-[`/api/emails/{email-template-name}`](/api/v1#put--api-emails--email-template-name-)
-
 ## Enterprise Users/Directory Searching
 
-Management API v1 allows you to search directly for users authenticated using enterprise connections, such as Active Directory or Azure Active Directory.
+The following endpoints allows you to search directly for users authenticated using enterprise connections, such as Active Directory or Azure Active Directory. These endpoints are not currently available in the [Management API v2](/api/v2).
 
 * All users from a specific directory:
 [`/api/connections/{connection}/users`](/api/v1#get--api-connections--connection--users)
```
panga: Are we restricting which connection they can use to search?
e.g: include APIv1 enabled check inside the users search code to only allow enterprise connections (not social/database).

articles/api/management/v1/use-cases.md
santiagoaguiar: Yup! I added a JIRA task to track this thing: https://auth0team.atlassian.net/browse/CF-55

panga: socket wasn't public too, we can add this into migration docs regardless of adding it into APIv2 API docs.

santiagoaguiar: 👍 Added!

articles/api/management/v2/changes.md
```diff
@@ -87,6 +87,8 @@ For Private Cloud (search_engine:v2), use `q=identities.connection:"connection_n
 | [POST /api/rules](/api/v1#!#post--api-rules) | None. | [POST /api/v2/rules](/api/v2#!/Rules/post_rules-) |
 | [PUT /api/rules/{rule-name}](/api/v1#!#put--api-rules--rule-name-) | Uses `{id}` instead of `rule-name`. | [PATCH /api/v2/rules/{id}](/api/v2#!/Rules/patch_rules_by_id) |
 | [DELETE /api/rules/{rule-name}](/api/v1#!#delete--api-rules--rule-name-) | Uses `{id}` instead of `rule-name`. | [DELETE /api/v2/rules/{id}](/api/v2#!/Rules/delete_rules_by_id) |
+| [GET /api/rules-configs](/api/v1) | None. | [GET /api/v2/rules-configs](/api/v2#!/Rules_Configs/get_rules_configs) |
```
zamd, line 90: Seems like `/rules-config` api  wasn't documented earlier in V1 page?

articles/api/management/v1/use-cases.md
```diff
@@ -8,40 +8,17 @@ contentType: reference
 useCase: invoke-api
 ---
 
-# Management API v1 Use Cases
+# Features not affected by the Management API v1 Deprecation
```
zamd: This "heading" and following paragraph suggests (to me at least) that it's OK to use these endpoints even in new apps etc. ??

articles/api/management/v2/changes.md
santiagoaguiar: It was not, but it was mentioned on the 'use cases' doc. It is used by a few `us` customers.

articles/api/management/v1/use-cases.md
santiagoaguiar: Yes. I just started a discussion about this. Originally, we intended to leave these endpoints outside of the deprecation (and therefore available as they were before) since we have no replacement for them. 

However, they only are used by just a few tenants in cloud, so we might remove this doc completely and mark all for deprecation, and make an exception for those  tenants explicitly until we can provide an alternative.

mostekcm: I'm all for removing them.  If we don't want people to use them, we should remove the public facing docs on them :)

panga (dismissed):

articles/api/management/v1/use-cases.md
mmaddex: Sounds good to me. Though, we'll have 14 public cloud customers still using them for the immediate future and no idea how many PSaaS customers are using these at the moment. As long as it's a capability for some of our customers we need it documented somewhere. Could we extract this to the either the TF knowledge base or DSE's troubleshooting pages? As long as folks can easily find this info I'm happy to remove it from the public docs.

cc @mauricioschneider

panga (dismissed):

articles/api/management/v1/use-cases.md
santiagoaguiar: The docs for the endpoints will still be available here: https://auth0.com/docs/api/management/v1 (as they were), but we'll remove the 'use cases' docs and the endpoints once we reach EOL.

panga (dismissed):

panga (dismissed):

rkhoriander (approved):

PR #8565, opened 2019-12-10
Update 01-authorization.md

phildulacvooban:
TokenValidationParameters can only be set in the AddJwtBearer method. It's not available in AddAuthentication.


damieng (approved):

damieng:
Thanks!

PR #8636, opened 2020-01-13
URI.escape method is obsolete

jeremiemv:
Per RuboCop _Lint/UriEscapeUnescape_

`URI.escape` method is obsolete and should not be used. Instead, use `CGI.escape`, `URI.encode_www_form` or `URI.encode_www_form_component` depending on your specific use case.


joshcanhelp (approved):

joshcanhelp:
Thank you @jeremiemv!

PR #8654, opened 2020-01-17
Add Management API setup for Apple connections

rkhoriander:


rkhoriander:
Added this info into new PR since SIWA file structure had changed drastically.

PR #8673, opened 2020-01-22
Update Ionic4 Instructions

davidpatrick:
In line with https://github.com/auth0-samples/auth0-ionic4-samples/pull/6 these instructions fix a logout issue with Ionic 4 that doesn't close the in app browser after logout.

### DIFF
--- a/articles/quickstart/native/ionic4/01-login.md
+++ b/articles/quickstart/native/ionic4/01-login.md
@@ -320,37 +320,40 @@ export class AuthService {
   }
 
   logout() {
-    this.storage.remove('profile');
-    this.storage.remove('access_token');
-    this.storage.remove('expires_at');
-    this.accessToken = null;
-    this.user = null;
-    this.loggedIn = false;
+      this.accessToken = null;
+      this.user = null;
+      this.loggedIn = false;
+      this.safariViewController.isAvailable()
+        .then((available: boolean) => {
+          const auth0Domain = AUTH_CONFIG.domain;
+          const clientId = AUTH_CONFIG.clientId;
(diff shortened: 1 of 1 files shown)


articles/quickstart/native/ionic4/01-login.md
```diff
@@ -193,6 +196,7 @@ export class AppComponent {
 
       // Redirect back to app after authenticating
       (window as any).handleOpenURL = (url: string) => {
+        this.auth.safariViewController.hide();
```
lbalmaceda: I think the auth service should not expose the safari view controller directly. Instead, a new method `hideBrowser` or similar could be created and the property remain _private_. 

Also, after checking the sample code, wouldn't it be better to call `hide()` after you get an event from that browser you just opened? if that's something that you receive and can relate to this scenario..

See 

https://github.com/auth0-samples/auth0-ionic4-samples/pull/6/files#diff-fab356ab6389b09c66727fb44caa49ccL80-L87

lbalmaceda (changes requested):
I'm sure this works but I think it can be improved and encapsulate fields on their respective classes/files.

lbalmaceda:
On a separate note, why did codeowners not kick in for this PR? https://github.com/auth0/docs/blob/master/.github/CODEOWNERS#L1

articles/quickstart/native/ionic4/01-login.md
davidpatrick: I agree @lbalmaceda , I had toyed with that solution, but opted to default to how the Ionic 3 sample we had was handling it.  After getting a second pair of eyes on it, definitely think we should go the route you are suggesting

lbalmaceda (approved):

PR #8707, opened 2020-01-30
Specifying Where to Place the ExternalAPI.js File

jamesqquick:
In the current description, it says to "create a component". This makes me assume it should go in the "components" directory. However, when this file is imported later on in the tutorial, it is referenced from within the "views" directory. This caused an error in my app since I originally placed it inside of the "components" directory. I updated the docs to specify where exactly to put the file.


lbalmaceda (approved):
Thanks!

PR #8751, opened 2020-02-09
[IPS-986] Update docs with usage guidelines

asmclean:
Add a description for how to use a wildcard (`*`) in subdomains in application configuration.


articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,37 @@
...
+
+* The wildcard **MUST** be located in a subdomain within the hostname component. `https://*.com` will not work.
+
+* The wildcard **MUST** be located in the subdomain furthest from the root domain. `https://sub.*.example.com` will not work.
+
+* The URL **MUST NOT** contain more than one wildcard. `https://*.*.example.com` will not work.
+
+* A wildcard **MAY** be prefixed and/or suffixed with additional valid hostname characters. `https://prefix-*-suffix.example.com` will work.
+
+* A URL with a valid wildcard **WILL** match a URL with no subdomain at the same level as the wildcard. `https://*.example.com` will work with `https://example.com`.
```
CriGoT: This only applies to the first subdomain and with some additional conditions. I think we can probably remove it because it would cause confusion that `http://*.sub.domain.com` does not match `http://sub.domain.com`

articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,37 @@
...
+
+* The wildcard **MUST** be located within the hostname component of the URL. Star symbols (`*`) located outside of the hostname, such as in the path or query, will not be treated as wildcards. `https://example.com/*` will not be treated as having a wildcard.
+
+* The wildcard **MUST** be located in a subdomain within the hostname component. `https://*.com` will not work.
+
+* The wildcard **MUST** be located in the subdomain furthest from the root domain. `https://sub.*.example.com` will not work.
+
+* The URL **MUST NOT** contain more than one wildcard. `https://*.*.example.com` will not work.
+
+* A wildcard **MAY** be prefixed and/or suffixed with additional valid hostname characters. `https://prefix-*-suffix.example.com` will work.
```
CriGoT: Not sure if this is the right place but we can suggest that if you are using a cloud provider for development we strongly recommend always using a suffix or prefix instead of general wildcards. wdyt?

CriGoT (changes requested):

articles/applications/reference/wildcard-subdomains.md
asmclean: Removed.

articles/applications/reference/wildcard-subdomains.md
asmclean: I added it there.  I'm not entirely sure either, but since we provide guidance in our best practices to avoid wildcards, that seems as good of a place as any.

CriGoT (dismissed):

articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,35 @@
...
+---
+# Wildcard for Subdomains
+
+Several fields in [Application Settings](/dashboard/reference/settings-application) allow for use of wildcards for subdomains for convenient registration of URLs:
+ 
+* **Allowed Callback URLs**: Set of URLs to which Auth0 is allowed to redirect users after they authenticate.
+* **Allowed Logout URLs**: List of URLs to which you can redirect users after they log out from Auth0.
+* **Allowed Origins (CORS)**: Set of URLs that will be allowed to make requests from JavaScript to Auth0 API (typically used with CORS).
+
+See [Application Settings Best Practices](/best-practices/application-settings) for recommended settings for applications, which suggests to avoid using wildcards for subdomains in in application callbacks and allowed origins. If you choose to use them, you can use the star symbol (`*`) as a wildcard for subdomains, but it must be used in accordance with the following rules in order to properly function.
```
ghost:
```suggestion
See [Application Settings Best Practices](/best-practices/application-settings) for recommended settings for applications, which suggests to avoid using wildcards for subdomains in application callbacks and allowed origins. If you choose to use them, you can use the star symbol (`*`) as a wildcard for subdomains, but it must be used in accordance with the following rules in order to properly function.
```

articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,35 @@
...
+
+* The wildcard **MUST** be located within the hostname component of the URL. Star symbols (`*`) located outside of the hostname, such as in the path or query, will not be treated as wildcards. `https://example.com/*` will not be treated as having a wildcard.
+
+* The wildcard **MUST** be located in a subdomain within the hostname component. `https://*.com` will not work.
+
+* The wildcard **MUST** be located in the subdomain furthest from the root domain. `https://sub.*.example.com` will not work.
+
+* The URL **MUST NOT** contain more than one wildcard. `https://*.*.example.com` will not work.
+
+* A wildcard **MAY** be prefixed and/or suffixed with additional valid hostname characters. `https://prefix-*-suffix.example.com` will work.  A wildcard for subdomains of cloud provider **SHOULD** always include at least a prefix or a suffix, such as `https://my-app-prefix-*.heroku.com`.
```
ghost:
```suggestion
* A wildcard **MAY** be prefixed and/or suffixed with additional valid hostname characters. `https://prefix-*-suffix.example.com` will work.  A wildcard for subdomains of cloud provider **SHOULD** always include at least a prefix or a suffix, such as `https://my-app-prefix-*.herokuapp.com`.
```

ziluvatar (dismissed):

tamigoodall (dismissed):

articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,33 @@
...
+
+* The wildcard **must** be located within the hostname component of the URL. Star symbols (`*`) located outside of the hostname, such as in the path or query, will not be treated as wildcards. `https://example.com/*` will not be treated as having a wildcard.
+
+* The wildcard **must** be located in a subdomain within the hostname component. `https://*.com` will not work.
+
+* The wildcard **MUST** be located in the subdomain furthest from the root domain. `https://sub.*.example.com` will not work.
+
+* The URL **must not** contain more than one wildcard. `https://*.*.example.com` will not work.
+
+* A wildcard **may** be prefixed and/or suffixed with additional valid hostname characters. `https://prefix-*-suffix.example.com` will work.  A wildcard for subdomains of cloud provider **should** always include at least a prefix or a suffix, such as `https://my-app-prefix-*.herokuapp.com`.
```
luuuis: A prefix on something like a `herokuapp.com` only gives an illusion of added security, but it's no more than that as anyone can create the `my-app-prefix-but-actually-an-evil-app` Heroku app.

For that reason I don't think we should endorse or suggest this by using it as an example and would rather see us use a fictitious domain name in our examples. Customers can still use this on Heroku but let them put 2 and 2 together...

luuuis (commented):
In `Wildcards for Subdomains` I think it's our responsibility to inform customers that using subdomain wildcards on domains where the public can register a new subdomain (e.g. Heroku) is fundamentally unsafe because customers won't be able to prevent others from registering a new subdomain using that same prefix.

Perhaps we can point to one of the better resources on "Subdomain takeover" that are already available?

articles/applications/reference/wildcard-subdomains.md
asmclean: We provide this guidance, which is linked in the beginning of this document: https://auth0.com/docs/best-practices/application-settings#do-not-use-wildcards-or-localhost-in-callbacks-or-origins-fields  Do you think we should include something like that directly in this document as well?

I'll remove the cloud provider example for now.

articles/applications/reference/wildcard-subdomains.md
```diff
@@ -0,0 +1,33 @@
...
+---
+# Wildcards for Subdomains
+
+You can use wildcards for subdomain URL registration in your application configuration in the Dashboard with these fields:
+
+* **Allowed Callback URLs**: Set of URLs to which Auth0 is allowed to redirect users after they authenticate.
+* **Allowed Logout URLs**: List of URLs to which you can redirect users after they log out from Auth0.
+* **Allowed Origins (CORS)**: Set of URLs that will be allowed to make requests from JavaScript to Auth0 API (typically used with CORS).
+
+See [Application Settings Best Practices](/best-practices/application-settings) for recommended settings for applications, which suggests to avoid using wildcards for subdomains in application callbacks and allowed origins. If you choose to use them, you can use the star symbol (`*`) as a wildcard for subdomains, but it must be used in accordance with the following rules in order to properly function:
```
luuuis: Can we make this sentence pop out as a warning/notice? (yellow background or something of the sort)

> Avoid using wildcards for subdomains in application callbacks and allowed origins as it can make your application vulnerable to attacks. See [Application Settings Best Practices](/best-practices/application-settings) for this and other recommended settings.

And then continue the regular flow with

> To set up subdomain wildcards you can use the star symbol (`*`) as a wildcard for subdomains ...

WDYT?

articles/applications/reference/wildcard-subdomains.md
luuuis: Good idea, perhaps  a note towards the top of the page (see my other comment)?

articles/applications/reference/wildcard-subdomains.md
asmclean: Screen Shot 2020-02-21 at 11 55 01 AM

luuuis (dismissed):

jeffreylees (approved):

PR #8785, opened 2020-02-25
To add an explanation of SendGrid settings (for Japanese career mail)

luckpoint:
To add an explanation of SendGrid settings about increasing deliverability for Japanese mobile career users.

(This information came from the SendGrid Japan team.)


rkhoriander:
Hi, @luckpoint . I don't know if you saw my comment in Slack, but we can't include this content since there isn't an English translation we can link to directly. If you would like to work with us to create similar content within Auth0, please let me know. Thanks!

PR #8795, opened 2020-02-27
Removed reference to implicit flow in favor of new SPA recommendation…

MatthewBacalakis:
…: Auth Code w/ PKCE.


rkhoriander:
Hey, @MatthewBacalakis   I will need to review this a little later. We have another PR in review that should go out by Monday at the latest. It updates all of our guidance about Implicit and Auth Code w/PKCE, so this may already be contained in it. As soon as it goes, I will review this PR again and see if this update is still needed.

rkhoriander (approved):

PR #8826, opened 2020-03-10
Made user fillable

rlewkowicz:
Following the original documentation results in the sub field receiving a null and then complaining about not being nullable.

### DIFF
--- a/articles/quickstart/webapp/laravel/01-login.md
+++ b/articles/quickstart/webapp/laravel/01-login.md
@@ -297,6 +297,27 @@ Save this file and run the artisan command to create this table:
 php artisan migrate
 ```
 
+From here, we need to make sub fillable. Edit the app/User.php file to remove the `password` entry and replace it with `sub`  
+
+```php
+// app/User.php
+// ...
+class User extends Authenticatable
+{
+    use Notifiable;
+
+    /**
+     * The attributes that are mass assignable.
+     *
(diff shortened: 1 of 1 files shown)


joshcanhelp:
@rlewkowicz - Appreciate the PR here! Can you explain a little more what's happening here and why this is necessary? I'm wondering if this is maybe happening in a newer version of Laravel as I have not come across this issue. 

Also, if this is required, we need to add a few more fields based on the scopes that we're using (which control what we get back as the user identity). The default scopes will populate the following fields, in addition to the ones you added:

- `nickname`
- `email_verified`
- `picture`
- `updated_at`

rlewkowicz:
@joshcanhelp 

I was running into an issue where it just wouldn't add the sub id to the database. So I'd login with a user, and it would throw an error that the field was not nullable. If you add a default value it would work once and then fail consecutive since we made it unique. 

Now it's been a few years since I've built an app, but I thought that there was a specific mass fillable exception, but it seemed to be failing silently. 

I actually took this https://auth0.com/blog/build-a-laravel-6-app-with-authentication/ and added in the custom user handling.

Making the field fillable fixed my issues

evansims:
Hey Ryan 👋  Thank you for your PR! This appears to have been rolled into a separate commit at some point and is now live in the docs, so I'll be closing this. Take care!

PR #8839, opened 2020-03-16
Add notice on AD/LDAP health extension for API v1 deprecation

santiagoaguiar:


articles/migrations/guides/management-api-v1-v2.md
```diff
@@ -65,9 +65,15 @@ By migrating your requests to API v2 and disabling API v1 as soon as possible, y
...
 :::
- 
+
 If you need help with the migration, contact us using the [Support Center](https://support.auth0.com/) or our [Community Site](https://community.auth0.com/c/auth0-community/Migrations).
 
+## Auth0 AD/LDAP Connector Health Monitor extension
+
+The [Auth0 AD/LDAP Connector Health Monitor](/extensions/adldap-connector) extension v1 uses the API v1 `GET /api/connections/{connection-name}` and `GET /api/connections/{connection-name}/socket` endpoints.
+
+Please upgrade to the latest version of the extension before disabling API v1 support.
```
mmaddex, line 75: Looks good thanks. I think my only feedback is if some customers are likely to run into an issue using the upgrade button, we could let them know here that they might need to remove and re-install the extension. I think there was some discussion about this in Slack, but I don't recall if this was still an issue or not.

santiagoaguiar: Discussed on slack. Customers could ran into this issue still, but I don't think we should put it in the doc. We can treat it on a case-by-case basis if necessary.

tamigoodall (approved):

PR #8840, opened 2020-03-17
Update auth0.android index page

erjohnson:
parity with auth0.android readme


lbalmaceda (commented):
@erjohnson 👋 We definitely don't want to have a mirror of the repo's readme file as the site's main article for the android library. 
See https://auth0.com/docs/libraries/auth0-android. Note on the sidebar we have different sections, each on a separate file and focused on showing a given objective. That would be the goal. 

Can you share more context on where this initiative came from?

PR #8854, opened 2020-03-23
Update 01-login.md

iliaselmatani:
"authlib.flask.client" is deprecated we need to use "authlib.integrations.flask_client" instead


lbalmaceda:
Correct, but this sample app was not hardcoding any version fo the `authlib` package.

image

Anyway, I enforced a min of `0.14.1` on the sample and made the same line change on the imports. Sample PR https://github.com/auth0-samples/auth0-python-web-app/pull/57

lbalmaceda (approved):

PR #8873, opened 2020-04-01
Adding Node 12 URLs to delegated admin extension docs.

seejee:
The introduction of Node 12 means that we will be using new webtask URLs for tenants who have opted into Node 12. 

This PR adds the new URLs to use when setting up the delegated admin extension.


rkhoriander:
These have already been deployed and fixed in other PRs, so closing. :)

PR #8874, opened 2020-04-01
Update Spring 5 API Quickstart to Spring Boot 2.2.6

jimmyjames:
Related to the change to the sample to use Spring Boot 2.2.6: https://github.com/auth0-samples/auth0-spring-security5-api-sample/pull/6

### DIFF
--- a/articles/quickstart/backend/java-spring-security5/01-authorization.md
+++ b/articles/quickstart/backend/java-spring-security5/01-authorization.md
@@ -55,7 +55,7 @@ If you are using Gradle, you can add the required dependencies using the [Spring
 // build.gradle
 
 plugins {
-    id 'org.springframework.boot' version '2.2.4.RELEASE'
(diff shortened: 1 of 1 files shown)


lbalmaceda (approved):

lbalmaceda:
Unless there's a security vulnerability involved, I'd not do this for every new patch/minor that goes out. As of today, this is not easy to keep up to date with the sample unless you manually open the 2 PRs.

jimmyjames:
> Unless there's a security vulnerability involved, I'd not do this for every new patch/minor that goes out. As of today, this is not easy to keep up to date with the sample unless you manually open the 2 PRs.

I agree. I wanted to try using the new DynamicPropertySource in Boot 2.2.6, so after I upgraded locally figured I'd just update the sample. But in general yes, it's not worth chasing every minor/patch unless there are security vulnerabilities or new features related to auth/security.

PR #8880, opened 2020-04-03
Fix salt in example

alex-broad:
The existing version does not validate against the schema or be accepted by the service. This is the fixed version that validates against the schema and works against the service.


alex-broad:
Not sure about this. Think the schema has a bug actually. Closing until I can figure it out.

PR #8881, opened 2020-04-03
Amend salt required parameters

alex-broad:
'hash' does not exist on the salt parameter. Suggest this should be 'position'.


ian-speers (approved):
Good catch, thanks!

PR #8901, opened 2020-04-09
Reformat the action table for /authorize parameters

leeomara:
Make it easier to compare the behaviour of `/authorize` arguments when there is and isn't an existing session.

The original version didn't cover what action would happen when no session exists but both `prompt=login&screen_hint=signup` is used.


erjohnson (approved):
💯

PR #8905, opened 2020-04-10
Add Product Release Stages

rkhoriander:
* Added intro to Product Lifecycle index page.
* Added Product Release Stages page + add EOL stage
* Edit Deprecations/EOL page
* Update Product Lifecycle side nav

https://docs-content-staging-pr-8905.herokuapp.com/docs/product-lifecycle


articles/product-lifecycle/product-release-stages.md
```diff
@@ -0,0 +1,71 @@
...
+
+## General Availability
+
+General Availability (GA) releases are fully functional and available to all subscribers (limited by pricing tier) for production use. If a new release replaces an existing feature, we provide a period of backward compatibility in accordance with our deprecation policy and inform customers so they have time to adopt the new release.
+
+## Deprecation
+
+Deprecated features are not supported for use by new subscribers, are not actively being enhanced, and are being only minimally maintained. <dfn data-key="tenant">Tenants</dfn> using the feature at the time of deprecation will continue to have access. 
+
+Deprecations begin with a <dfn data-key="breaking-change">breaking change</dfn> and end when the referenced feature or behavior moves into the End of Life product release stage. During Deprecation, customers should engage in a <dfn data-key="migration">migration</dfn> to move away from the deprecated feature or behavior. To learn more, see [Migration Process](/product-lifecycle/migration-process).
```
mmaddex: Kind of a nitpick, but should we mention something here about how the breaking change would only be experienced as a breaking change at the End Of Life? If I read this as literally as possible it sounds like 1. we break functionality, 2. customers migrate to the new functionality. 

I'm probably overthinking this line, maybe no one else would read it that way... and I'm not sure how to better phrase this. May something like 
> Deprecation beings with the introduction of new behavior that would be experienced as a breaking change if no action is taken to migrate away before the End Of Life...
?

Anyway, just a thought.

mmaddex: Oh actually... just reading the migrations-process page. That breaks it down bit more how I think about it in my head
> When we must introduce a breaking change, we first deprecate the affected feature or behavior and announce the Deprecation to our customers.

So 1. feature is deprecated 2. migration performed 3. breaking change mitigated 🥳

mmaddex:
Looks good 👍 - love the new format. Just the one misc thought above.

articles/product-lifecycle/product-release-stages.md
rkhoriander: Heroku app is rebuilding, but see what you think about my change. I believe it sidesteps the problem and keeps it simple, while leaving questions about the full process on the migration process page.

randynasson (approved):

PR #8926, opened 2020-04-15
Add steps for RSA JWT value calculations

### LINKED ISSUE #8731: Incorrect description of signature validation in case of RSA256 encryption.
### Description

Description of signature validation in case of RSA256 encryption is not correct - 
[Validate JSON Web Tokens](https://auth0.com/docs/tokens/guides/validate-jwts)

> Remember that the Signature is created using the Header and Payload segments, a signing algorithm, and a secret or public key (depending on the chosen signing algorithm).

Signature in RSA256 created from message hash by using private key.
 
> To verify that the signature is correct, you need to generate a new Base64url-encoded signature using the public key (RS256) or secret (HS256) and verify that it matches the original Signature included with the JWT:
> 
> Take the original Base64url-encoded Header and original Base64url-encoded Payload segments (Base64url-encoded Header + "." + Base64url-encoded Payload), and hash them with SHA-256.
> Encrypt using either HMAC or RSA (depending on your selected signing algorithm) and the appropriate key.

In case of RSA following steps performed.

1. Calculate current Hash Value.
In the first step, a hash-value of the signed message is calculated. For this calculation, the same hashing algorithm is used as was used during the signing process. The obtained hash-value is called the current hash-value because it is calculated from the current state of the message.

2. Calculate the Original Hash-Value.
In the second step of the digital signature verification process, the digital signature is decrypted with the same encryption algorithm that was used during the signing process. The decryption is done by the public key that corresponds to the private key used during the signing of the message. As a result, we obtain the original hash-value that was calculated from the original message during the first step of the signing process (the original message digests).

3. Compare the Current and the Original Hash-Values.
In the third step, we compare the current hash-value obtained in the first step with the original hash-value obtained in the second step. If the two values are identical, the verification if successful and proves that the message has been signed with the private key that corresponds to the public key used in the verification process. If the two values differ from onr another, this means that the digital signature is invalid and the verification is unsuccessful.

tamigoodall:
Add steps for RSA JWT value calculations

### DIFF
--- a/articles/tokens/guides/validate-jwts.md
+++ b/articles/tokens/guides/validate-jwts.md
@@ -81,6 +81,7 @@ To verify the signature, you will need to:
     1. Retrieve the `alg` property from the decoded Header.
     2. Ensure that it is an allowed algorithm. Specifically, to avoid certain attacks, make sure you disallow `none`.
     3. Check that it matches the algorithm you selected when you [registered your Application](/getting-started/set-up-app) or [API](/getting-started/set-up-api) with Auth0.
+
 2. Confirm that the token is correctly signed using the proper key. Check the Signature to verify that the sender of the JWT is who it says it is and that the message wasn't changed along the way.
 
     Remember that the Signature is created using the Header and Payload segments, a [signing algorithm](/tokens/concepts/signing-algorithms), and a secret or public key (depending on the chosen signing algorithm).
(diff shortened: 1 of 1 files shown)


rkhoriander:
Still working on this one. Don't think the provided info is being presented quite correctly. Doing some research.

tamigoodall:
Updated in Contentful

PR #8927, opened 2020-04-15
Zendesk logout URL setting

### LINKED ISSUE #8353: Missing Tenant configuration step for SSO redirectTo support
### Description

The external SSO [integration configuration documentation](https://github.com/auth0/docs/blob/master/articles/integrations/sso/_template.md) (I was using it for Zendesk) does not discuss how to properly configure the tenant to support the `returnTo` param defined in Remote Logout URL.

### Reproduction

If the existing documentation is followed and the Auth0 generated Remote Logout URL is used within Zendesk, an Auth0 error page will be shown when the user attempts to logout.

The issue is that for this generated `returnTo` to work, the _Tenant > Settings > Advanced > Allowed Logout URLs_ also needs to be updated to support the logout URL. It took a bit of troubleshooting and discussion with support for us to figure out this was the root cause of the error.

_Note: Zendesk was adding their own additional params including `return_to` which caused our debugging to be a lot longer._

### Environment

Current Auth0 tenant version (created 3 months ago)

tamigoodall:
Zendesk logout URL setting

### DIFF
--- a/articles/integrations/sso/_template.md
+++ b/articles/integrations/sso/_template.md
@@ -164,6 +164,8 @@ Zendesk **requires** that all users have an email address. When enabling Enterpr
 that can be sent to Zendesk.
 :::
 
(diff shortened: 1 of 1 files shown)


tamigoodall:
Updated in Contentful

PR #8931, opened 2020-04-16
refresh-token-rotation.md with partial attack

HunderlineK:
Demonstrates the usefulness of 'Automatic reuse detection' by detailing how it can mitigate a partially successful attack. The current documentation for Automatic reuse detection has an example of how it completely prevents an attack, but in doing so it fails to note that there is still a window of attack.

I suspect the original example was written only with replay attacks in mind.

### DIFF
--- a/articles/tokens/concepts/refresh-token-rotation.md
+++ b/articles/tokens/concepts/refresh-token-rotation.md
@@ -44,11 +44,11 @@ Without enforcing sender-constraint, it’s impossible for the authorization ser
 
 For example, consider the following scenario: 
 1. Legitimate Client has **Refresh Token 1**, and it is leaked to or stolen by Malicious Client. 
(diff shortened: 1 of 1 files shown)


HunderlineK:
I just realized the associated diagram on the web page needs to be updated as well; unfortunately, I don't have access to the tools for creating that.

HunderlineK:
There is also another vector of attack with even a more significant impact: if the legitimate client never uses the refresh token e.g. user leaves their session before the expiry of their first token, then the automatic invalidation will never be invoked, and the malicious client can keep using the token.

Compared to a cookie-based approach, the rotating refresh token leaves open a rathe glaringly vector of attack open.

Perhaps it could be mitigated with required PKCE for every request? Or maybe that's what implemented under the hood if SDK is used and it is just not documented?

PR #8937, opened 2020-04-17
Prefer double-hyphen option leader in command line examples for impor…

ericksoen:
Using the double-hyphen option leader, e.g., `--config_file` instead of `-c` should reduce cognitive load for new users reading the documentation.


seejee (approved):
Looks great. Thanks!

rkhoriander (approved):

PR #8944, opened 2020-04-22
Removing C# support for rules

rkhoriander:
RIP

https://docs-content-staging-pr-8944.herokuapp.com/docs/rules/guides/csharp
https://docs-content-staging-pr-8944.herokuapp.com/docs/rules/current/csharp
should both redirect to:
https://docs-content-staging-pr-8944.herokuapp.com/docs/rules


config/redirects.js
```diff
@@ -2236,6 +2236,7 @@ module.exports = [
   { from: '/rules/current/redirect', to: '/rules/guides/redirect' },
   { from: '/rules/current/metadata-in-rules', to: '/rules/guides/metadata' },
   { from: '/rules/current/csharp', to: '/rules/guides/csharp' },
+  { from: '/rules/guides/csharp', to: '/rules' },
```
seejee, line 2239: Awesome, didn't know we had this capability! 👍

seejee (dismissed):

config/redirects.js
seejee: And I assume the redirect rule above this line will chain through to this one? So `'/rules/current/csharp'` will end up at `'/rules'`?

rkhoriander: Yes. Although, I guess technically I could just make it go directly to /rules without the jump.

rkhoriander: In fact, that's a better choice. Fixed it.

seejee (approved):

tamigoodall (approved):

PR #8946, opened 2020-04-23
Update express quickstart

davidpatrick:
We are removing the secure localhost setup.  There were concerns with generating a root local authority and the complexity this adds to the quickstart.

The reason why we have the secure local guidance there is because of same site flags coming from Chrome for localhost.  However, they have announced that they will delay the roll out of same site here https://www.chromium.org/updates/same-site

With the delay in rollout, this gives us time to improve the secure localhost guidance, or change the quickstart to use authcode+pkce to avoid having to use secure localhost for a quickstart.


adamjmcgrath (changes requested):
You'll also need to change the `baseURL` to `http`

davidpatrick:
Thanks @adamjmcgrath 😅

adamjmcgrath (approved):

pdillon:
Late nit comment: should usage of `appSessionSecret` be changed to `appSession.secret` or something similar to reflect the config change?

PR #8948, opened 2020-04-23
Account Linking / email_verified changes

jeffreylees:


articles/best-practices/rules.md
```diff
@@ -253,14 +253,13 @@ There may be occasions when it becomes necessary to persist, say, [updates to me
...
+Failure to call the function will result in a stall of pipeline execution, and ultimately in an error condition being returned. Each rule then must call the `callback` function exactly once;. Calling it once prevents the stall of the pipeline, but more could cause unpredictable results or errors.
 
 ```js
-    // Roles should only be set to verified users.
-    if (!user.email || !user.email_verified) {
+    if (!user.email) {
     } else {
       getRoles(user.email, (err, roles) => {
```
aaguiarz: Let's use `user.user_id` instead email, so we make the email thing a non-issue here

articles/best-practices/rules.md
```diff
@@ -288,7 +287,7 @@ As authentication has already occurred, any early exit of the pipeline with an (
...
 :::
 
-The example provided above also demonstrates best practice use of both [early exit](/best-practices/performance#exit-early) as well as [email address verification](#check-if-an-email-is-verified), as described in [Performance Best Practices](/best-practices/performance) and [Security Best Practices](/best-practices/custom-db-connections/security). Note: the `getRoles` function used is implemented elsewhere within the rule, as a wrapper function to a third-party API.
+The example provided above also demonstrates best practice use of [early exit](/best-practices/performance#exit-early) as described in [Performance Best Practices](/best-practices/performance). Note: the `getRoles` function used is implemented elsewhere within the rule, as a wrapper function to a third-party API.
```
aaguiarz: If we remove the if (user) then we need to remove the paragraph about 'early exit'

articles/best-practices/rules.md
```diff
@@ -333,6 +332,8 @@ function (user, context, callback) {
 }
 ```
 
+However, if you want to execute different code depending on the company the user belongs to, you **should not rely on the email domain**, but instead on data that can link the user to the identity provider they authenticated with (connection/identity provider specific fields such as Azure’s **tenant id**).
```
aaguiarz:
```suggestion
However, if you want to execute different code depending on the company the user belongs to, you **should not rely on the email domain**, but instead on data that can link the user to the identity provider they authenticated with (the connection or identity-provider -specific fields such as Azure’s **tenant id**).
```

articles/security/bulletins/2019-01-10_rules.md
```diff
@@ -159,9 +166,12 @@ The above logic would return true given emails such as these:
 - `"user@domain.com"@not-domain.com` (quotes included)
 
 ### Am I affected?
+
 Only customers using conditional logic in rules as described above, are affected.
 
 ### Mitigation steps
+
+
```
aaguiarz: Do we need two blank lines?

articles/security/bulletins/2019-01-10_rules.md
```diff
@@ -253,9 +271,11 @@ To mitigate this risk, consider one or more of the following actions:
 - Ensure that the rule only gets triggered for an authorized subset of users, or under other appropriate conditions. For example, you may wish to add logic that checks if a user has a particular email domain, <dfn data-key="role">role</dfn>/group, or subscription level before triggering the call to the paid service.
 
 ### Will this update impact my users?
+
 The impact to end users depends on which of the options described above you choose to mitigate the issue.
 
 ## Authorization based on unverified email address
```
aaguiarz:
```suggestion
## Authorization based on email address
```

articles/security/bulletins/2019-01-10_rules.md
```diff
@@ -253,9 +271,11 @@ To mitigate this risk, consider one or more of the following actions:
...
+
 If you have rule logic that requires access control based on a particular email or email domain, but you are failing to check whether that user has verified their email, an adversary could potentially gain additional privileges by using a second connection type. For example:
```
aaguiarz:
```suggestion
If you have rule logic that requires access control based on a particular email or email domain, but you are failing to check whether that user has verified their email, an adversary could potentially gain additional privileges by using a second connection type. For example:
```
```suggestion
If you have rule logic that requires access control based on a particular email or email domain, an adversary could potentially gain additional privileges by using a second connection type. For example:
```

articles/security/bulletins/2019-01-10_rules.md
```diff
@@ -266,11 +286,15 @@ If you have rule logic that requires access control based on a particular email
 
 The above logic would return **true** if an adversary could create an account using a different connection type (such as _social_) with an email address present in the allow-list. This happens because the same email can exist in different connection types.
 
+If you want to execute different code depending on the company the user belongs to, you **should not rely on the email domain**, but instead on data that can link the user to the identity provider they authenticated with (connection/identity provider specific fields such as Azure’s **tenant id**).
```
aaguiarz:
```suggestion
Checking if the email is verified before using it for granting authorization can give you some sense of security but is not enough. You can read more about how to use email verification in this [document](... link to Eric's doc).
 
If you want to execute different code depending on the company the user belongs to, you **should not rely on the email domain**, but instead on data that can link the user to the identity provider they authenticated with (the connection/identity-provider-specific fields such as Azure’s **tenant id**).
```

aaguiarz (changes requested):

aaguiarz (changes requested):
Let's remove this doc

articles/security/bulletins/2019-01-10_rules.md
jeffreylees:
```suggestion
```

articles/best-practices/rules.md
```diff
@@ -233,7 +233,7 @@ Updating a user via use of the `auth0` object ultimately results in a call to th
...
 
-There are occasions when `primaryUser` must be updated as the primary user may change&mdash;i.e., the user returned to the Auth0 engine will be different from the user on rule pipeline entry; [automatic account linking](/users/concepts/overview-user-account-linking#automatic-account-linking) is the primary use case. On such occasions, a rule must update `primaryUser` to reflect the new primary user identifier. Note that this change *will not* affect any subsequent rule executed in the current instance of the pipeline; the `user` object will remain unchanged.
+There are occasions when `primaryUser` must be updated as the primary user may change&mdash;i.e., the user returned to the Auth0 engine will be different from the user on rule pipeline entry. On such occasions, a rule must update `primaryUser` to reflect the new primary user identifier. Note that this change *will not* affect any subsequent rule executed in the current instance of the pipeline; the `user` object will remain unchanged.
```
esarafianou, line 236: @aaguiarz what are other cases where the user returned to the Auth0 engine will be different from the user on rule pipeline entry? Does this sentence still make sense if we remove the account linking?

aaguiarz: @esarafianou No, this only applies to account linking. However, we are not removing account linking, we are changing the guidance on how to trigger it.

articles/best-practices/rules.md
```diff
@@ -253,17 +253,16 @@ There may be occasions when it becomes necessary to persist, say, [updates to me
...
+The `callback` function supplied to a rule effectively acts as a signal to indicate completion of the rule. A rule should complete immediately following a call to the callback function, either implicitly or by explicitly executing a (JavaScript) `return` statement, and should refrain from any other operation.
 
-Failure to call the function will result in a stall of pipeline execution, and ultimately in an error condition being returned. Each rule then must call the `callback` function exactly once; at least once in order to prevent stall of the pipeline, but not more than once&mdash;otherwise unpredictable results and/or errors will occur:
+Failure to call the function will result in a stall of pipeline execution, and ultimately in an error condition being returned. Each rule then must call the `callback` function exactly once;. Calling it once prevents the stall of the pipeline, but more could cause unpredictable results or errors.
 
 ```js
-    // Roles should only be set to verified users.
-    if (!user.email || !user.email_verified) {
+    if (!user.user_id) {
```
aaguiarz: @jeffreylees this does not make sense as the user_id always exists for a user

articles/users/concepts/overview-user-account-linking.md
```diff
@@ -168,25 +168,15 @@ The [Auth0 Node.js SDK for APIv2](https://github.com/auth0/node-auth0/tree/v2) i
...
 ## Scenarios
 
-Here are some scenarios that implement account linking:
-* [Automatic account linking](#automatic-account-linking): automatically link accounts with the same email address
+Here are two scenarios that implement account linking:
+
 * [User-initiated account linking](#user-initiated-account-linking): allow your users to link their accounts using an admin screen in your app
 * [Suggested account linking](#suggested-account-linking): identify accounts with the same email address and prompt the user in your app to link them
 
 ::: warning
```
aaguiarz: I think this warning is not required, as with user-initiated account linking, you don't need verified emails

aaguiarz (approved):

PR #8986, opened 2020-04-30
fixing bad link

jeffreylees:
Resolves DR-556


krismb (approved):

krismb (dismissed):
This page has a larger update in the works which will remove the changed line, but I am approving and merging to correct the broken link.

jeffreylees:
Let's not try to fix the build error and do the link fix here - I'll add it to Andres PR instead

PR #8992, opened 2020-05-05
Improved references to account linking samples

aaguiarz:


articles/users/references/link-accounts-client-side-scenario.md
```diff
@@ -55,15 +55,15 @@ You can find the full source of this sample application [on GitHub](https://gith
 
   if (!email_verified) {
     throw new Error(
-      `Account linking is only allowed to a verified account. Please verify your email ${email}.`
+      `Account linking is only allowed to a verified account. Please verify your email <%= "${email}" %>.`
     );
   }
 
-  await fetch(`https://${config.domain}/api/v2/users/${sub}/identities`, {
+  await fetch(`https://${config.domain}/api/v2/users/<%= "${sub}" %>/identities`, {
```
jeffreylees: Is `${config.domain}` intended to be an actually processed variable? If so it should be this: `${account.namespace}` right?

jeffreylees (approved):

PR #8996, opened 2020-05-05
APIv1 PSaaS Migration 

mmaddex:
Add notes for PSaaS


mmaddex:
@tehsis Could you update the "Am I affected by the migration" section for PSaaS customers. Is there a release after which we disabled API v1 by default for new tenants etc?  Sync with the Core Foundations team if necessary.

tehsis (dismissed):

rkhoriander (approved):

PR #9096, opened 2020-05-29
Add custom scope key to Node (Express) API: Authorization

kdesimini:
In Auth0, the "scope" are actually permissions. The library express-jwt-authz, uses the actual "scope" key from the token by default and to avoid confusion, should include the custom scope key by default.

[This](https://auth0.com/docs/quickstart/backend/nodejs/01-authorization#protect-api-endpoints) is the url to the section of the requested change.

### DIFF
--- a/articles/quickstart/backend/nodejs/01-authorization.md
+++ b/articles/quickstart/backend/nodejs/01-authorization.md
@@ -97,7 +97,7 @@ Pass the `checkJwt` and `checkScopes` middlewares to the route you want to prote
 ```js
 // server.js
 
-const checkScopes = jwtAuthz([ 'read:messages' ]);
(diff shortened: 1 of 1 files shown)


kdesimini:
@davidpatrick Is this still relevant? If not, I can close this PR.

davidpatrick:
@kdesimini I will take a look at this next week, thanks for pinging me

davidpatrick:
`customScopeKey` is to change the location of where the scopes are stored in the JWT.  By default those are stored under the `scope` key in a JWT, but `customScopeKey` allows you to change checking the default location of the token. We don't change the default location of the scope key in the JWT, so this change is not needed.

PR #9097, opened 2020-05-30
Use npm instead of yarn

thameera:
The rest of the tutorial uses `npm`. Also, we haven't given any specific instructions to install `yarn`. Using `npm start` is a good default.


stevehobbsdev (approved):
Thanks! I agree, let's go with `npm` here.

PR #9101, opened 2020-06-01
Fixed broken link in Use Refresh Token Rotation

tamigoodall:
Fixed broken link


tamigoodall:
Updated in Contentful - however, the link will have to be changed once the /libraries content has been migrated.

jeffreylees (approved):

PR #9134, opened 2020-06-08
Update index.md - MyLock.Theme

sdotlock:
@style/MyLock.Theme does not exist. Replacing with @style/Lock.Theme solves the error.


lbalmaceda (approved):
Thanks!

PR #9205, opened 2020-06-28
Update 02-calling-an-api.md erroneous link to localhost

nopestack:
Docs linked to `localhost:3000` as opposed to the intended location, furthermore, the URL did not point to a valid location when replacing `localhost:3000` with `https://auth0.com/` but the proposed URL does point to the section I believe was the intention on the author to point the reader towards.


stevehobbsdev (approved):
Good catch! Thanks!

PR #9215, opened 2020-06-30
Update tenant-create.png

tamigoodall:
Update tenant-create.png to use .us.auth0.com


rkhoriander:
Updated in Contentful.

rkhoriander (approved):

PR #9234, opened 2020-07-07
Update Logout Example button text

taylorcjohnson:
In the `auth0-react` library documentation it looks like the `LogoutButton` text (of the [Logout Section](https://auth0.com/docs/libraries/auth0-react#logout)) is currently 'Log in' - should this read 'Log out' instead?


jeffreylees (approved):
Thanks!

PR #9244, opened 2020-07-13
fix variable name auth0_api

abbaspour:
$auth0_auth_api => $auth0_api


tamigoodall:
Hi @abbaspour 
We are currently in a content freeze while we migrate all docs content to Contentful CMS. I have created a ticket in our Jira backlog to make this change in Contentful after our freeze is over. https://auth0team.atlassian.net/browse/DOCS-1413

erjohnson:
updated in contentful, should be live shortly

PR #9246, opened 2020-07-14
Split Java and Android instructions for generating code verifier

edwinyoung:
At the moment, the Java instructions listed for generating a code verifier for the PKCE flow are actually Android instructions (since the `java.util.Base64` class in Java has no constants, unlike `android.util.Base64`).

These changes fix that problem by separating the instructions for Java and Android.


tamigoodall:
We are currently in a content freeze. We will address this change after the freeze is complete.

tamigoodall:
Change made in contentful content.

PR #9265, opened 2020-07-22
Update IdP-Initiated SSO documentation to reference the enable step

CriGoT:
Updates screenshots and texts to include the new option used to enable/disable IdP-Initiated SSO flows.


luuuis (dismissed):

articles/protocols/saml/saml-configuration/troubleshoot/common-saml-errors.md
```diff
@@ -33,6 +33,31 @@ This message indicates that the Application doesn't have an active Connection as
...
 
+## Error: IdP-Initiated Login Is Not Enabled
+
+```json
+{
+  "invalid_request": "IdP-Initiated login is not enabled for connection 'CONNECTION_NAME'."
+}
+```
+
+### How to Fix
```
ziluvatar: 🤔  This way to fix it is risky to tell like this, I mean, if customer is really doing an IdP Initiated Login, then yes, this is the way to fix it, however, as it explains below this is not the solution for SP Initiated.

CriGoT: I'm not sure I follow. This fixes the error, the steps are clear that you are enabling IdP-Initiated and it has a note about running into the error when using SP-initiated. What am I missing?

BTW, this is the way that this document is structured with the panel after the guide

ziluvatar: Yeah, I saw it is the same after 🤷 ,  If it's clear for you let's leave it as is, for me with the "developer troubleshooting hat":
> I developed an integration with an IdP using usual SP integration, I start seeing this error, I come here I read the error, I read the how to fix, I do it, problem solved, probably I won't even read the panel below and will not know I suddenly open my system to IdP Initiated logins (I mean I read those words but... I don't know what that means probably).

My suggestion is to invert the position of panel vs steps. So the safe solution for SP initiated is shown first.
This can apply to the next item in the guide too.

ziluvatar (approved):

tamigoodall:
Hi  @CriGoT - We have migrated to another Documentation content management system (Contentful) and I have made your changes there. You can see these as drafts in these link:

Configure SAML Identity Provider-Initiated Single Sign-On: https://auth0-docs-staging.herokuapp.com/docs/contentful/5DtrDKCUrGZMmm5QfKaW3T

Configure OneLogin: https://auth0-docs-staging.herokuapp.com/docs/contentful/goKGoqcik465QDrcgULI3

Troubleshoot SAML Configurations: https://auth0-docs-staging.herokuapp.com/docs/contentful/7IE3fmXv22zPjYYpx9EZh3

PR #9267, opened 2020-07-23
typo in the LoginButton

daisuke-awaji:
There is a typo in the part exported by the LoginButton


stevehobbsdev (approved):
Good spot - thanks!

PR #9271, opened 2020-07-24
Update customize-with-hooks.md

ColCoutts:
Removed 'beta' tag as hooks are now in GA :)


tamigoodall:
No need for this. @rkhoriander has it covered.

PR #9274, opened 2020-07-27
Add references in the refresh token rotation doc

radekk:
Adding the two new links to provide a better guidance to a user. Otherwise, we leave that to a user to search our docs and try to understand where are the log events listed as well as not being able to tell directly that log streaming could be used to detect suspicious activity.

- New link to log event types
- New link to log streaming
- Added a few words of explanation of why to use log streaming


tamigoodall:
These updates are being made to the docs that have been migrated to Contentful. They will not be made in the GH content.

PR #9315, opened 2020-08-03
Update the-basics.md

K-Sato1995:
Found a type and changed it to `US` for coherence.


jeffreylees:
Thanks! I have made the change in our CMS platform, which will go live with on Monday, August 10. We're no longer updating most of our docs via GitHub. Closing this PR!

PR #9316, opened 2020-08-03
Update manage.md

tswallen:
DELETE should be POST as the method for /api/v2/users/{id}/recovery-code-regeneration


jeffreylees:
Thanks! The change has been made in Contentful; it will go live with the Contentful launch on Monday, August 10. Closing this PR!

PR #9326, opened 2020-08-03
Make version agnostic and fix typos

imliam:
Makes the documentation version agnostic. Currently, it points to Laravel 7 for everything, however, the package and instructions will work throughout versions 5, 6, 7 and 8 which will be released next month, so there is no point of hardcoding it to a version that'll always make it seem like the wrong version for some people.

Also fixes a typo later on where the wrong variable name is referenced, and a cast is made unnecessarily.

### DIFF
--- a/articles/quickstart/webapp/laravel/01-login.md
+++ b/articles/quickstart/webapp/laravel/01-login.md
@@ -20,15 +20,15 @@ github:
 
 <%= include('../../../_includes/_logout_url', { returnTo: 'http://localhost:3000' }) %>
 
-## Install and Configure Laravel 6
+## Install and Configure Laravel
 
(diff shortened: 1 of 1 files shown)


evansims:
Hey Liam! 👋  Really appreciate the PR, thank you! I've gone ahead and updated our documentation to reference V8 and fixed that typo in separate PRs. Didn't mean to steal your thunder on this PR I just didn't see it beforehand, my bad!

I certainly respect the reasoning of your suggestion in removing specific version suggestions from an upkeep standpoint, and you're quite right we don't require the latest and greatest version of Laravel for our SDK to work. However, I do believe there's value in the clarity that recommending specific versions brings to quickstarts. We do point out that we support a variety of versions in the SDK documentation itself, but as far as quickstarts go, being a little more opinionated on versions for the sake of clarity probably isn't such a bad thing.

Thank you again for the PR and your suggestions!

PR #9345, opened 2020-08-06
Update Android external dev docs links

rkhoriander:
Update Android docs links that link checker showed were causing a redirection loop.


lbalmaceda (approved):
Both previous and updated links work fine for me.

PR #9367, opened 2020-08-12
Update redirect.md

abbaspour:
fix redirect url


erjohnson:
updated in contentful, should be live soon

PR #9400, opened 2020-08-27
Update redirects.js

tamigoodall:
Added missing redirects

### DIFF
--- a/config/redirects.js
+++ b/config/redirects.js
@@ -1641,7 +1641,7 @@ module.exports = [
     to: '/flows/client-credentials-flow'
   },
   {
-    from: ['/flows/concepts/token-exchange','/flows/guides/auth-code/add-login-auth-code','/flows/guides/auth-code/includes/authorize-user-add-login','/flows/guides/auth-code/includes/sample-use-cases-add-login','/flows/guides/auth-code/includes/refresh-tokens','/flows/guides/auth-code/includes/request-tokens','/flows/guides/regular-web-app-login-flow/add-login-using-regular-web-app-login-flow','/oauth-web-protocol', '/protocols/oauth-web-protocol', '/protocols/oauth2/oauth-web-protocol','/application-auth/current/server-side-web','/client-auth/server-side-web','/application-auth/legacy/server-side-web'],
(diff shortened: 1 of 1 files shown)


config/redirects.js
```diff
@@ -2066,6 +2066,10 @@ module.exports = [
     from: ['/sso/current/integrations/office-365', '/integrations/sso/office-365','/integrations/office-365'],
     to: '/integrations/sso-integrations/office-365'
   },
+  {
+    from: ['/integrations/office-365-custom-provisioning'],
+    to: '/integrations/sso-integrations/office-365-custom-provisioning'
```
rkhoriander: This doesn't exist. Looks like the from URL is the correct one.

config/redirects.js
```diff
@@ -1641,7 +1641,7 @@ module.exports = [
...
   {
-    from: ['/flows/guides/auth-code/add-login-auth-code','/flows/guides/auth-code/includes/authorize-user-add-login','/flows/guides/auth-code/includes/sample-use-cases-add-login','/flows/guides/auth-code/includes/refresh-tokens','/flows/guides/auth-code/includes/request-tokens','/flows/guides/regular-web-app-login-flow/add-login-using-regular-web-app-login-flow','/oauth-web-protocol', '/protocols/oauth-web-protocol', '/protocols/oauth2/oauth-web-protocol','/application-auth/current/server-side-web','/client-auth/server-side-web','/application-auth/legacy/server-side-web'],
+    from: ['/flows/concepts/token-exchange','/flows/guides/auth-code/add-login-auth-code','/flows/guides/auth-code/includes/authorize-user-add-login','/flows/guides/auth-code/includes/sample-use-cases-add-login','/flows/guides/auth-code/includes/refresh-tokens','/flows/guides/auth-code/includes/request-tokens','/flows/guides/regular-web-app-login-flow/add-login-using-regular-web-app-login-flow','/oauth-web-protocol', '/protocols/oauth-web-protocol', '/protocols/oauth2/oauth-web-protocol','/application-auth/current/server-side-web','/client-auth/server-side-web','/application-auth/legacy/server-side-web'],
```
rkhoriander: The Token Exchange flow is a completely different flow than Auth Code. I wrote the docs for it last fall, but the project has been on hold since then. I believe they're picking it up some time next year again. But it also looks the like the part of the Dashboard UI that is directing people there is hidden unless your tenant has a specific feature flag set.

I'll defer to you since you know this Dashboard request, but in my opinion, this link should 404 since the docs don't exist. Redirecting to Auth Code seems like it might confuse some people?

tamigoodall: That makes total sense. Thanks for pointing that out.

config/redirects.js
tamigoodall: You are correct

tamigoodall:
All links accounted for.

PR #9461, opened 2020-09-24
Deploy CLI Docs Review

faroceann:
- Fix typo in `::note`
- Make docs more clear about keyword /environment variable replace mapping and their utility
- Make docs less specific about where the deploy CLI can be used (not just "existing CI/CD pipelines") 
- Avoid overwhelming readers with info on export methods in intro paragraph and leave it to be covered later. At this point in the doc, they are still gathering an understanding of what the tool does to begin with.


tamigoodall:
Changes made to docs in Contentful.

PR #9480, opened 2020-10-09
[SDK-2013] Update PHP SDK version reference to current version

evansims:
Quickstart fix: the PHP SDK version we are referencing is out of date, should be bumped to the current V7 release.

### DIFF
--- a/snippets/server-platforms/php/dependencies.md
+++ b/snippets/server-platforms/php/dependencies.md
@@ -1,5 +1,5 @@
 To install dependencies, run the following
 
 ```bash
-composer require auth0/auth0-php:"~5.0"
+composer require auth0/auth0-php:"~7.0"
 ```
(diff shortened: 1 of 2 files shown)


snippets/server-platforms/php/dependencies.md
```diff
@@ -1,5 +1,5 @@
 To install dependencies, run the following
 
 ```bash
-composer require auth0/auth0-php:"~5.0"
+composer require auth0/auth0-php:"~7.0"
```
frederikprijck, line 4: Should we also update the requirements for the quickstart considering the [PHP support ](https://www.php.net/supported-versions.php)?

image

evansims: Good catch, thanks! I'll update to fix that as well. That PHP version needs bumped to 7.1 as well as that's the minimum supported by our PHP SDK. I'm on it.

frederikprijck (approved):

PR #9513, opened 2020-11-13
Removed confusing paragraph

tamigoodall:
DR-793 from Alejandro Jurado Walls


avanscoy (approved):
I can see where that paragraph could be confusing.

PR #9553, opened 2021-01-05
Updated old New Universal Login screen with new screenshot provided b…

avanscoy:
…y Andres


misterpantz (approved):
binary so I can't see :)

misterpantz (approved):
binary so I can't see. :)

PR #9574, opened 2021-01-13
add rails > 5.2 notes about change in credentials

Datise:
The newer rails projects will use a different credential storage system then is referenced by the Quickstart. After recently setting up a new Rails 6 app, this was the only discrepancy between versions that I experienced.

 Given as it's a security sensitive step, I felt it was probably better to be as explicit as seemed necessary. We don't want to accidentally lead a new dev down a misinformed path. 

Should I also update the requirements to list be > Rails 5.2 ?

### DIFF
--- a/articles/quickstart/webapp/rails/01-login.md
+++ b/articles/quickstart/webapp/rails/01-login.md
@@ -82,6 +82,33 @@ end
 This tutorial uses omniauth-auth0, a custom [OmniAuth strategy](https://github.com/intridea/omniauth#omniauth-standardized-multi-provider-authentication).
 :::
 
+#### For Rails 5.2 and greater
+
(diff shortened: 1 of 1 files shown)


davidpatrick:
Hey @Datise thanks for the work on this.  My concern with this approach for Rails 5.2+ is it makes the quickstart a bit confusing.  Have you looked at using the custom configuration API https://guides.rubyonrails.org/configuring.html#custom-configuration which was added in Rails 4.2.0?  This will allow us to have consistent instructions on the sample and quickstart for both 4.2.0+ and 5.2.0+

davidpatrick (changes requested):
I'll be helping contribute to this. Marking as reviewed for now.

davidpatrick:
Will open new PR

PR #9575, opened 2021-01-14
g suite --> google workspace

misterpantz:


misterpantz:
Closing pull request because I discovered it's in Contentful after all

PR #9623, opened 2021-02-12
Correct Information for 3-Day Verification Window

dapperdandev:
The documentation currently states that you have a 3-day window for Auth0 to verify CNAME entries AFTER the entries are added. 

The 3-day window starts as soon as "Add Domain" is clicked. Auth0 would not have a way to know when the CNAME entries were added.


tamigoodall:
This change is a known issue that is being updated in our source files which are now sourced in Contentful.

PR #9639, opened 2021-03-09
Update dependency array for React Quickstart (Calling an API)

CustomCodeIT:
When running through the documentation on a brand new react project, I noticed the user metadata wasn't ever displaying because the useEffect was firing a single time before the user gets populated. By adding "getAccessTokenSilently" and "user?.sub" to the useEffect dependency array, I was able to get this code working reliably.

You can see in the console logs from the catch block that the first fire fails as the user has not yet been defined. When the user is updated, useEffect fires again which causes the metadata to populate properly.


stevehobbsdev (approved):
Thanks for the contribution!

PR #9654, opened 2021-03-17
Remove the passport QS

adamjmcgrath:
The Passport QS is obsolete, we recommend the `express-openid-connect` one now.


config/redirects.js
```diff
@@ -43,7 +43,7 @@ module.exports = [
   {
     from: ['/design','/design/web'],
     to: '/'
-  },     
+  },
```
adamjmcgrath: My IDE deleted all the trailing whitespace - view the diff with `&w=1`

stevehobbsdev (approved):

PR #9656, opened 2021-03-17
Grammer suggestion

klequis:
Added the word "to" to "You also have the means require the credentials". Not having it stopped the flow of reading.


tamigoodall:
Hi @klequis - Thanks for the update. I have made the update to our source files in Contentful. They should be live shortly.

PR #9660, opened 2021-03-20
replace env local with the actual values

woloski:
I was watching a twitch livestream of someone integrating nextjs live and realized that the quickstart is not replacing the values for client id, secret and domain. That lead to a user error when trying to replace the domain in issuerBaseUrl they didn't put the https and that took a while to realize


Widcket (approved):

Widcket:
Thanks!

PR #9714, opened 2021-04-30
Add redirect for ping federate saml idp config doc

erjohnson:


avanscoy (approved):
Tested the redirect and it works :D

PR #9740, opened 2021-06-03
Base 64-compatible characters

Javarome:
To make the function `randomString()` function more convenient, replaced special characters `.-_` (which are not allowed by base 64) by `+/` (which are allowed).
See https://base64.guru/learn/base64-characters


rkhoriander:
Made change to code sample in new CMS. Closing this PR since the issue has been addressed. Thanks!

PR #9757, opened 2021-06-16
Add entity limit redirects

rkhoriander:
Change redirect from Authz Core to entity limits (combine pages).


avanscoy (commented):
Tested redirect link. Went to the correct page. Approved.

avanscoy (approved):

PR #9783, opened 2021-07-02
Remove "$" from bash snippets

lbalmaceda:
These snippets should not include the "$" character at the beginning. When that's the case and the text is copied and ran in a terminal, the command fails.


cocojoe (approved):
👍

rkhoriander (approved):

PR #9785, opened 2021-07-06
Updated Guzzle Link

thepersonthatexists:
Current Guzzle link seems to be outdated, updated to https://docs.guzzlephp.org/en/stable/


evansims:
Hey @danielh00 👋 Thanks for your PR! The link to Guzzle's site is accurate, and is reflected as the URL listed on [their GitHub repo](https://github.com/guzzle/guzzle). I understand your concern, as it does currently redirect to the URL you mentioned, but I'm more comfortable pointing to the same as their GitHub repo, should they decide to change redirects at some point.

PR #9838, opened 2021-08-05
Update 01-authorization.md

olegykz:
By using `JWT::JWK.import` you can verify not only RSA keys (`kty: 'RSA'`) but also ECDSA (`kty: 'EC'`)


Widcket:
cc @evansims as you were looking into the Rails QS

Widcket:
cc @stevehobbsdev

PR #9853, opened 2021-08-18
Quickstart redirects for new nav

tamigoodall:
Quickstart redirects for new nav


rkhoriander:
Opened a new PR with correct changes.

PR #9906, opened 2021-10-22
Fix common errors table

y13i:
Removed an extra column in the head/body separator line so it should be rendered properly.
Screen Shot 2021-10-22 at 10 04 57


y13i:
It is fixed now.

PR #9914, opened 2021-11-03
Update AD/LDAP warning panel

avanscoy:
docs-question request to update warning panel language.


avanscoy:
After discussion with Nico, this panel should not change or define what the support team should do.

PR #9919, opened 2021-11-08
[DXDOCS-129] Set up script for cleaning chained redirects

pclavier92:
Use `jscodeshift` to manipulate the `configs/redirects.js` file and merge chained redirects that affect SEO.

(At the moment the only changes introduced to the `configs/redirects.js` file were running `prettier`)

This will allow to clean up the redirects file using the command `yarn run clean-redirects`.

This will also introduce a GitHub Action for running this script once a week.


config/redirects.js
```diff
@@ -1,6843 +1,7403 @@
...
-  'php',
-  'python',
-  'rails',
-  'ruby',
-  'salesforce-sandbox',
-  'salesforce',
-  'sap-odata',
-  'wcf-service',
-  'webapi-owin'
+  "aspnet-webapi",
```
pdillon: should configure a prettierrc file for single quotes
```
{
  "singleQuote": true,
  "arrowParens": "always",
  "printWidth": 120,
  "trailingComma": "es5"
}
```

pdillon: unless this makes the ast transform harder

pdillon (approved):
👏 Nice work!

PR #9931, opened 2021-11-11
Update 01-login.md

Janik-DEDON:
Typo fix. As a verb, 'set up' is an open compound (two unhyphenated words).


stevehobbsdev (approved):
Thanks for the contribution! 🎉

PR #9937, opened 2021-11-18
fix: minor note change to the webapp golang quickstart[1].

rene00:
https://auth0.com/docs/quickstart/webapp/golang


sergiught (approved):
Thanks a lot for fixing this!:)

PR #9961, opened 2022-01-05
Fix Rails login normalized user profile

jusleg:
Given that `session[:userinfo]` already maps to raw info, we can no longer show the full user profile by calling `session[:userinfo][:extra][:raw_info]` as it leads to a `NoMethodError`. Now, we display the normalized information by using `@user` directly. This is in line with the changes in the sample repository https://github.com/auth0-samples/auth0-rubyonrails-sample/commit/c0da5862af2e548186ec00771e15e9de1aa75c4a#diff-6c10f94ab4d3763545c6c4868dcbdb21e18335b7dcbd7390bcf5a74c1ce350f7R9


Widcket:
cc @evansims as you were looking into the Rails QS

PR #9972, opened 2022-01-14
Update 01-login.md

MattMorden:
Updated typo for class name when accessing AuthenticationScheme constant value in Login method.


MattMorden:
Constants does not exist, but Auth0Constants does as seen in the image here. This also matches the value seen in the Logout method. 
image

stevehobbsdev:
@frederikprijck

Widcket (approved):
`Auth0Constants` is mentioned in the **Login** section above
> To add the `Login`, call `ChallengeAsync` and pass "Auth0" (`Auth0Constants.AuthenticationScheme`) as the authentication scheme.

PR #9983, opened 2022-01-19
Update _getting_started.md - PHP callback URL

ghost:
Apologies if this is not the correct file to edit.

I believe there is an issue in the document 'https://auth0.com/docs/quickstart/webapp/php#configure-callback-urls'.

I've followed the tutorial through and upon testing the application received the error:

image

The document states to enter the callback URL in Auth0 as 'http://127.0.0.1:3000/', however the tutorial states to configure the PHP application with 'http://127.0.0.1:3000/callback'. Entering 'http://127.0.0.1:3000/callback' into the Auth0 callback URL works.

### DIFF
--- a/articles/quickstart/webapp/_includes/_getting_started.md
+++ b/articles/quickstart/webapp/_includes/_getting_started.md
@@ -9,5 +9,5 @@
 <%= include('../../../_includes/_callback_url') %>
 
 ::: note
-If you are following along with the sample project you downloaded from the top of this page, the callback URL you need to add to the **Allowed Callback URLs** field is `${callback}`.
+If you are following along with the sample project you downloaded from the top of this page, the callback URL you need to add to the **Allowed Callback URLs** field is `${callback}`/callback.
 :::


Ben-S-218i:
Apologies if this is not the correct file to edit.

I believe there is an issue in the document 'https://auth0.com/docs/quickstart/webapp/php#configure-callback-urls'.

I've followed the tutorial through and upon testing the application received the error:

image

It appears that the docs state to enter the callback URL in Auth0 as 'http://127.0.0.1:3000/' however the tutorial states to configure the PHP application with 'http://127.0.0.1:3000/callback'. Entering 'http://127.0.0.1:3000/callback' into the Auth0 callback URL works.

evansims:
Hey @Ben-S-218i 👋 Thanks for your suggestion on this, you bring up a good point. That's the correct file but, unfortunately, this is a shared template across all our quickstarts, so altering this would affect that guidance across all our SDK's guides. I'm not quite sure the best course of action on addressing this one yet, but I'll see what I can figure out!

PR #10008, opened 2022-02-23
Correct spring oauth server maven dependency

poovamraj:
Spring boot oauth2 resource server dependency was pointed wrong in the maven code snippet. This has been fixed in this PR


jimmyjames (approved):
👍

PR #10033, opened 2022-03-17
New Python API QS migrated to new design

adamjmcgrath:
I updated the Python QS to simplify it using `Authlib` - and took the opportunity to migrate it to the new design at the same time.

image

### DIFF
--- a/articles/quickstart/backend/python/index.yml
+++ b/articles/quickstart/backend/python/index.yml
@@ -36,9 +36,8 @@ hidden_articles:
   - interactive
 requirements:
   - Python 3.6 and up
-  - Authlib 1.0.0
   - Flask 2.0.3
-  - Flask-Cors 3.0.10
+  - Authlib 1.0.0
 next_steps:
   - path: 01-authorization
     list:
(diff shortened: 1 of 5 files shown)


articles/quickstart/backend/python/index.yml
```diff
@@ -32,12 +32,12 @@ sdk:
...
+hidden_articles:
+  - interactive
 requirements:
-  - python 2.7, 3.3 and up
-  - flask 0.11.1
-  - python-jose-cryptodome 1.3.2
-  - flask-cors 3.0.2
-  - six 1.10.0
+  - Python 3.6 and up
+  - Flask 2.0.3
```
evansims: I think most readers would know better, but should we consider being slightly less specific about the version listed as required, to avoid confusion that a very specific version is needed for some reason? For example, when Flask 2.0.4 or 2.1 inevitably comes down the road, might our QS saying 2.0.3 specifically be confusing? Maybe just saying "Flask 2.0" and "Authlib 1.0" would be enough here 🤔 Up to you!

evansims (approved):
Just left a note worth consideration, but I leave that to you to decide. LGTM otherwise!

articles/quickstart/backend/python/index.yml
adamjmcgrath: Yeah - good point 👍

PR #10095, opened 2022-05-26
Updates token audience validation for Spring 2.7

ebrooks42:


Widcket:
cc @jimmyjames @poovamraj

jimmyjames:
Nice! And FYI we are planning to update the quickstart and sample to use Spring Boot 3 and Spring Security 6, there may be additional improvements to the DX for authN/authZ with that release.

jimmyjames (approved):

PR #10150, opened 2022-10-25
Fix links impacted by readme redesign

ewanharris:
This fixes links broken by the readme redesign, I can't fully remember if the content outside of quickstarts is synced with the site or if that's in contentful so let me know if I should revert that


Widcket:
> if the content outside of quickstarts is synced with the site or if that's in contentful

IIRC the content outside of quickstarts is all in Contentful now, and it's not synced.

Widcket (approved):

PR #10198, opened 2023-01-09
ULP-3961: update custom-texts docs

sebadoom:
We have been receiving reports from customers that are having trouble using custom texts due to the fact the text-ids are wrong. This PR brings the custom-texts docs up to date. 

## Reference
- Jira: https://auth0team.atlassian.net/browse/ULP-3961


kegashe (approved):
LGTM

PR #10199, opened 2023-01-11
SES-1601 API documentation for /oidc/logout

kbrimble:


articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
-- If the `client_id` parameter is included, the `returnTo` URL must be listed in the `Allowed Logout URLs` set at the application level (see [Setting Allowed Logout URLs at the App Level](/logout#set-the-allowed-logout-urls-at-the-application-level)).
-- If the `client_id` parameter is NOT included, the `returnTo` URL must be listed in the `Allowed Logout URLs` set at the tenant level (see [Setting Allowed Logout URLs at the Tenant Level](/logout#set-the-allowed-logout-urls-at-the-tenant-level)).
-- If the `client_id` parameter is included and the `returnTo` URL is NOT set, the server returns the user to the first Allowed Logout URLs set in the Dashboard (see [Setting Allowed Logout URLs at the Tenant Level](/logout#set-the-allowed-logout-urls-at-the-tenant-level)).
+Use this endpoint to logout a user. If you want to navigate the user to a specific URL after the logout, set that URL at the `post_logout_redirect_uri` parameter. The URL should be included in any the appropriate `Allowed Logout URLs` list:
```
avanscoy: In the sentence: "The URL should be included in any" is a little awkward. Should it be: "The URL should be included in the appropriate `Allowed Logout URLs` list" or "The URL should be included in any of the the appropriate `Allowed Logout URLs` lists"?

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
+
+- If the `client_id` parameter is included, the `post_logout_redirect_uri` URL must be listed in the `Allowed Logout URLs` set at the application level (see [Setting Allowed Logout URLs at the App Level](https://auth0.com/docs/logout#set-the-allowed-logout-urls-at-the-application-level)).
```
avanscoy: URLs in this repo only use the ending slug, so what's after /docs: [Setting Allowed Logout URLs at the App Level](/logout#set-the-allowed-logout-urls-at-the-application-level)

avanscoy: Same for the URLs below.

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
+| `id_token_hint` (Recommended)         | Previously issued ID Token for the user. This is used to indicate which user to log out                                                                         |
+| `logout_hint` (Optional)              | Optional `sub` (user ID) or `sid` (session ID) value to indicate which user to log out.                                                                         |
+| `post_logout_redirect_uri` (Optional) | URL to redirect the user after the logout.                                                                                                                      |
+| `client_id` (Optional)                | The `client_id` of your application.                                                                                                                            |
+| `federated` (Optional)                | Add this query string parameter to the logout URL, to log the user out of their identity provider, as well: `https://YOUR_DOMAIN/oidc/logout?federated`.        |
```
avanscoy: Add this query string parameter to log the user out of their identity provider: `https://YOUR_DOMAIN/oidc/logout?federated`.

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
+| `state` (Optional)                    | An opaque value the applications adds to the initial request that the authorization server includes when redirecting the back to the`post_logout_redirect_uri`. |
+| `ui_locales` (Optional)               | Space-delimited list of locales used to constrain the language list for the request. The first locale on the list must match the enabled locale in your tenant  |
 
 
 ### Test with Authentication API Debugger
 
 <%= include('../../_includes/_test-this-endpoint') %>
 
-1. At the *Configuration* tab, set the fields **Application** (select the application you want to use for the test) and **Connection** (the name of the social connection to use).
+1. At the Configuration tab, set the fields **Application** (select the application you want to use for the test) and **Connection** (the name of the social connection to use).
```
avanscoy: *Configuration* should be italicized to be consistent with the rest of the API docs: https://auth0.com/docs/api/authentication#idp-initiated-single-sign-on-sso-flow

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
-1. At the *Other Flows* tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
+1. At the _Other Flows_ tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
 
 
 ### Remarks
 
 - Logging the user out of their identity provider is not common practice, so think about the user experience before you use the `federated` query string parameter.
-- The sample auth0.js script uses the library version 8. If you are using auth0.js version 7, please see this [reference guide](/libraries/auth0js/v7).
+- If providing both `id_token_hint` and `logout_hint`, the `logout_hint` value must match either the `sub` or `sid` claim from the `id_token_hint`.
+- If `id_token_hint` is not provided then the user will be prompted for consent unless a `logout_hint` that matches either the user's ID or session ID is provided.
```
avanscoy: Comma with if/then statements: If `id_token_hint` is not provided, then

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
+- The `POST` HTTP method is also supported for this request. When using `POST`, the request parameters should be provided in the request body as form parameters instead of the query string.
+  - As the `federated` parameter requires a boolean value, it should be sent with a value of `true` or `false`.
```
avanscoy: This line has an extra space ^
 The `federated` parameter requires a value of `true` or `false`.

articles/api/authentication/_logout.md
```diff
@@ -1,16 +1,24 @@
 # Logout
 
 ```http
-GET https://${account.namespace}/v2/logout?
+GET https://${account.namespace}/oidc/logout?
```
julioramirez: This call can be made in many ways, but there are two expected:

1. using id_token_hint alone (it contains all the needed info) e.g. GET https://${account.namespace}/oidc/logout?
  post_logout_redirect_uri=LOGOUT_URL&
  id_token_hint=ID_TOKEN_HINT

2. using logout_hint, client_id ....
e.g. GET https://${account.namespace}/oidc/logout?
  client_id=${account.clientId}&
  post_logout_redirect_uri=LOGOUT_URL&
  logout_hint=LOGOUT_HINT

The recommendation is to use (1). The reason why there are two are use cases where the IDT can't be stored in the client (SPAs without a backend). Then, they have to fallback to logout_hint

So, lets place an example of each or, if that is not possible, an example with the id_token_hint

articles/api/authentication/_logout.md
julioramirez: when sending an `id_token_hint` alone, the service will use the `aud` claim to pick the 'Allowed Logout URLs'

when both `id_token_hint` and `client_id` are used, they must match (so the same as above applies)

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
 
-1. At the *Other Flows* tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
+1. At the _Other Flows_ tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
 
 
 ### Remarks
 
 - Logging the user out of their identity provider is not common practice, so think about the user experience before you use the `federated` query string parameter.
-- The sample auth0.js script uses the library version 8. If you are using auth0.js version 7, please see this [reference guide](/libraries/auth0js/v7).
+- If providing both `id_token_hint` and `logout_hint`, the `logout_hint` value must match either the `sub` or `sid` claim from the `id_token_hint`.
```
julioramirez: @avanscoy do you think we should bring the 'Custom Client Implementations' text in the public doc, to this API doc instead? alternativelly I think the API doc can be less verbose and link to the public docs.

Ref. https://docs.google.com/document/d/1irPMWMcpCkIXfl9YfkrppsnGyH1qpZE43iDGmq1udhs/edit#

julioramirez: - If providing both `id_token_hint` and `logout_hint`, the `logout_hint` value must match either the `sub` or `sid` claim from the `id_token_hint`.
- If providing both `id_token_hint` and `client_id`, the `client_id` value must match either the `aud` claim from the `id_token_hint`.

articles/api/authentication/_logout.md
```diff
@@ -24,50 +32,62 @@ curl --request GET \
...
+- If `id_token_hint` is not provided then the user will be prompted for consent unless a `logout_hint` that matches either the user's ID or session ID is provided.
+- The `POST` HTTP method is also supported for this request. When using `POST`, the request parameters should be provided in the request body as form parameters instead of the query string.
+  - As the `federated` parameter requires a boolean value, it should be sent with a value of `true` or `false`.
+- This conforms to the [OIDC RP-initiated Logout Specification](https://openid.net/specs/openid-connect-rpinitiated-1_0.html).
 
 ### More Information
 
 - [Logout](/logout)
```
julioramirez, line 167: @avanscoy we should add a link to the public documentation. We'd need a placeholder by now. wdyt?

articles/api/authentication/_logout.md
kbrimble: I've changed this section a bit to try and cover these different cases. I fear that it's getting quite complicated and that there isn't a clear understanding, at least in these docs, what overrides what. Looking for suggestions to make this better!

articles/api/authentication/_logout.md
avanscoy: I feel like the custom client details aren't necessary here and we can link to the public docs.

rgrace-puck:
Is there a public issue tracking the progress for the endpoint discovery UI toggle?

articles/api/authentication/_logout.md
kbrimble: Added now that we have it

articles/api/authentication/_logout.md
```diff
@@ -1,16 +1,22 @@
...
-  --url 'https://${account.namespace}/v2/logout' \
+  --url 'https://${account.namespace}/oidc/logout' \
   --header 'content-type: application/json' \
-  --data '{"client_id":"${account.clientId}", "returnTo":"LOGOUT_URL"}'
+  --data-raw '
+  { 
+    "client_id":"${account.clientId}", 
+    "post_logout_redirect_uri":"LOGOUT_URL", 
+    "id_token_hint":"ID_TOKEN_HINT",
+    "logout_hint":"LOGOUT_HINT"
```
julioramirez: Even though it is permitted(*), there is no use case for "id_token_hint" & "logout_hint" to be included in the same request. A user needs one (normally "id_token_hint") or the other ("logout_hint" if the app cannot store IDTs)

I suggest we make two examples, each with one of the claims. Or just one with the "id_token_hint" and add a comment to the alternate "logout_hint"

(*) If they are both included their sid values must match (and would hence be redundant)

articles/api/authentication/_logout.md
```diff
@@ -24,32 +30,42 @@ curl --request GET \
   });
   
   webAuth.logout({
-    returnTo: 'YOUR_LOGOUT_URL',
-    client_id: '${account.clientId}'
+    post_logout_redirect_uri: 'YOUR_LOGOUT_URL',
+    id_token_hint: 'YOUR_ID_TOKEN_HINT',
+    logout_hint: 'YOUR_LOGOUT_HINT'
```
julioramirez: Same comment than `shell` example about id_token and logout hints.

articles/api/authentication/_logout.md
```diff
@@ -24,32 +30,42 @@ curl --request GET \
...
+  - When the `client_id` parameter is NOT included, the server uses the URL from the `aud` claim in the `id_token_hint` to select which of the `Allowed Logout URLs` at the tenant level to use.
+- If the `id_token_hint` parameter is not included:
+  - If the `client_id` parameter is included, the `post_logout_redirect_uri` URL must be listed in the `Allowed Logout URLs` set at the application level (see [Setting Allowed Logout URLs at the App Level](/docs/logout#set-the-allowed-logout-urls-at-the-application-level)).
+  - If the `client_id` parameter is NOT included, the `post_logout_redirect_uri` URL must be listed in the `Allowed Logout URLs` set at the tenant level (see [Setting Allowed Logout URLs at the Tenant Level](/docs/logout#set-the-allowed-logout-urls-at-the-tenant-level)).
+  - If the `client_id` parameter is included and the `post_logout_redirect_uri` URL is NOT set, the server returns the user to the first `Allowed Logout URLs` set in the Dashboard (see [Setting Allowed Logout URLs at the Tenant Level](/docs/logout#set-the-allowed-logout-urls-at-the-tenant-level)).
```
julioramirez, line 130: If the id_token_hint is not available, include the `sid` value in the `logout_hint`

articles/api/authentication/_logout.md
```diff
@@ -24,32 +30,42 @@ curl --request GET \
...
-|:-----------------|:------------|
-| `returnTo `      | URL to redirect the user after the logout. |
-| `client_id`      | The `client_id` of your application. |
-| `federated`      | Add this query string parameter to the logout URL, to log the user out of their identity provider, as well: `https://${account.namespace}/v2/logout?federated`. |
+| Parameter                             | Description                                                                                                                                                     |
+| :------------------------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `id_token_hint` (Recommended)         | Previously issued ID Token for the user. This is used to indicate which user to log out                                                                         |
+| `logout_hint` (Optional)              | Optional `sub` (user ID) or `sid` (session ID) value to indicate which user to log out.                                                                         |
```
julioramirez: Remove: `sub` (user ID) or
note: we are changing that

articles/api/authentication/_logout.md
```diff
@@ -24,32 +30,42 @@ curl --request GET \
...
+| `post_logout_redirect_uri` (Optional) | URL to redirect the user after the logout.                                                                                                                      |
+| `client_id` (Optional)                | The `client_id` of your application.                                                                                                                            |
+| `federated` (Optional)                | Add this query string parameter to log the user out of their identity provider: `https://YOUR_DOMAIN/oidc/logout?federated`.        |
+| `state` (Optional)                    | An opaque value the applications adds to the initial request that the authorization server includes when redirecting the back to the`post_logout_redirect_uri`. |
+| `ui_locales` (Optional)               | Space-delimited list of locales used to constrain the language list for the request. The first locale on the list must match the enabled locale in your tenant  |
```
julioramirez, line 142: I don't think we are honoring this parameter. We need to confirm with Udo

articles/api/authentication/_logout.md
```diff
@@ -58,16 +74,22 @@ Use this endpoint to logout a user. If you want to navigate the user to a specif
...
-1. Copy the <dfn data-key="callback">**Callback URL**</dfn> and set it as part of the **Allowed Logout URLs** of your [Application Settings](${manage_url}/#/applications).
+1. Copy the **_Callback URL_** and set it as part of the **Allowed Logout URLs** of your Application Settings.
 
-1. At the *Other Flows* tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
+1. At the _Other Flows_ tab, click **Logout**, or **Logout (Federated)** to log the user out of the identity provider as well.
 
 
 ### Remarks
 
 - Logging the user out of their identity provider is not common practice, so think about the user experience before you use the `federated` query string parameter.
```
julioramirez: I am hesitant to add this. This comment about not doing federated logout applies to social login but not so to enterprise federations. We should add that we refer to social login feds instead.

articles/api/authentication/_logout.md
```diff
@@ -58,16 +74,22 @@ Use this endpoint to logout a user. If you want to navigate the user to a specif
...
-- The sample auth0.js script uses the library version 8. If you are using auth0.js version 7, please see this [reference guide](/libraries/auth0js/v7).
+- If providing both `id_token_hint` and `logout_hint`, the `logout_hint` value must match either the `sub` or `sid` claim from the id_token_hint.
```
julioramirez: We don't allow the  `sub` claim anymore

articles/api/authentication/_logout.md
```diff
@@ -58,16 +74,22 @@ Use this endpoint to logout a user. If you want to navigate the user to a specif
...
+- If providing both `id_token_hint` and `client_id`, the `client_id` value must match the `aud` claim from the `id_token_hint`.
+- If `id_token_hint` is not provided, then the user will be prompted for consent unless a `logout_hint` that matches either the user's ID or session ID is provided.
```
julioramirez: remove "either the user's ID or", uid does not apply anymore

articles/api/authentication/_logout.md
```diff
@@ -58,16 +74,22 @@ Use this endpoint to logout a user. If you want to navigate the user to a specif
...
+- The `POST` HTTP method is also supported for this request. When using `POST`, the request parameters should be provided in the request body as form parameters instead of the query string. The federated parameter requires a value of `true` or `false`.
```
julioramirez, line 162: I suggest we add an example

kbrimble: Unfortunately, there are 3 available example types which are all provided at the top of the file. Only 1 HTTP example can be provided and we are providing the GET one.

articles/api/authentication/_logout.md
kbrimble: I've added the word "social":

> Logging the user out of their social identity provider is not common practice, so think about the user experience before you use the `federated` query string parameter with social identity providers.

articles/api/authentication/_logout.md
kbrimble: I've updated the "Request Parameters" table with this information as all of these points relate to redirect URLs

articles/api/authentication/_logout.md
kbrimble: I've removed the `logout_hint` from the examples. Unfortunately, there isn't really a way of adding in extra examples on this page. Perhaps it can go on the actual documentation page?

articles/api/authentication/_logout.md
```diff
@@ -1,16 +1,21 @@
 # Logout
```
avanscoy, line 2: Okay. `oidc/logout` is taking the place of `v2/logout` but is there going to be another entry for `/v2/` logout?

articles/api/authentication/_logout.md
kbrimble: Checked and we will be for GA 👍

avanscoy (approved):

PR #10216, opened 2023-01-28
Update 02-calling-an-api.md

kastl:
user_metadata field didn't exist. I just changed it to email which was one of the fields returned.


kastl:
Apparently there is a way to get the field to show if it is added in the Auth0 settings.

PR #10243, opened 2023-03-31
Guidance to use custom scheme for Expo

poovamraj:
Updating Expo Quickstarts to use custom scheme mandatorily. This is to avoid Auth0 using same scheme as Expo to redirect after authentication.

The following PRs update the samples and README -
https://github.com/auth0/react-native-auth0/pull/610
https://github.com/auth0-samples/auth0-react-native-sample/pull/113

### DIFF
--- a/articles/quickstart/native/react-native-expo/00-login.md
+++ b/articles/quickstart/native/react-native-expo/00-login.md
@@ -112,7 +112,7 @@ These values are found in the Expo config file at `app.json` or `app.config.js`.
 
 <%= include('../../../_includes/_callback_url') %>
 
-#### Callback URL and logout URL
+#### iOS callback URL
 
 ```text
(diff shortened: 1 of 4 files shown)


stevehobbsdev (dismissed):
LGTM, if @Widcket could please take a look as well.

articles/quickstart/native/react-native-expo/00-login.md
```diff
@@ -96,10 +101,10 @@ These values are found in the Expo config file at `app.json` or `app.config.js`.
 {
     ...
     "android": {
-      "package": "com.auth0samples"
+      "package": "YOUR_ANDROID_PACKAGE_NAME_IS_FOUND_HERE"
```
Widcket, line 104: Given that above (L93 and L95) we use `com.auth0samples` as the example value, should we keep it here? Or change these above? Having two different ways to refer to the same thing in a single tutorial can be confusing, as a developer might think these refer to different things.

Widcket (approved):

PR #10244, opened 2023-04-06
Update custom database postgresql templates

XanderPSON:
This PR updates the PostgreSQL templates to use the current Client syntax. Currently the templates throw an error message as `postgres.connect` has been deprecated since postgres v7.0.0.


XanderPSON:
@joshcanhelp

PR #10247, opened 2023-04-14
Docs 3080 private key jwt updates

avanscoy:

### DIFF
--- a/articles/api/authentication/_multifactor-authentication.md
+++ b/articles/api/authentication/_multifactor-authentication.md
@@ -103,6 +103,8 @@ If OTP is supported by the user and you don't want to request a different factor
 | Parameter        | Description |
 |:-----------------|:------------|
 | `mfa_token` <br/><span class="label label-danger">Required</span> | The token received from `mfa_required` error. |
+| `client_assertion`| A JWT containing a signed assertion with your application credentials. Required when Private Key JWT is your application authentication method.|
+|`client_assertion_type`| The value is `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.  Required when Private Key JWT is the application authentication method.|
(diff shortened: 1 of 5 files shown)


articles/api/authentication/_multifactor-authentication.md
```diff
@@ -103,6 +103,8 @@ If OTP is supported by the user and you don't want to request a different factor
 | Parameter        | Description |
 |:-----------------|:------------|
 | `mfa_token` <br/><span class="label label-danger">Required</span> | The token received from `mfa_required` error. |
+| `client_assertion`| A JWT containing a signed assertion with your application credentials. Required when Private Key JWT is your application authentication method.|
+|`client_assertion_type`| The value is `urn:ietf:params:oauth:client-assertion-type:jwt-bearer`.  Required when Private Key JWT is the application authentication method.|
```
CharlesRea: Should we bold "Required" and "Application Authentication Method" to be consistent with `client_secret` below? Or remove the bold text from them?

Can we switch the ordering so these below `client_id` in the table? I think it feels most natural to have `client_id` first (very minor point though!). And the same for the similar exmples below.

CharlesRea (commented):
Looks good!

articles/api/authentication/_multifactor-authentication.md
avanscoy: Sure! Is the preference bold or not? I feel like that could be too many bold statements for a short list, but lmk what you think.

I can switch the `client_id` param, no problem.

CharlesRea: I'd agree, feels like too much bold. Should we get rid of it altogether?

pdillon (approved):

PR #10248, opened 2023-04-14
Remove small duplications in 01-login.md

nathburg:
Removes duplicate import in views.py code block. 
Removes duplicate line about configuring Authlib to handle application's authentication.


frederikprijck (approved):
Thanks!

Issue #10254, opened 2023-04-19
Add faq "User is not logged in after page refresh" to the "01-login.md" for the people following the quick start guide.


decipher-cs:
### Description

When following "Show User Profile Information" in the Auth0 SPA quickstart guide (see image), user might encounter the problem ["isAuthenticated and user are null after login #93"](https://github.com/auth0/auth0-react/issues/93) if using safari or brave browser.

### Solution

Add a link to the faq to the quickstart guide.

![auth0 show user profile information](https://user-images.githubusercontent.com/130742861/233001586-bdcc2ffd-d3df-4282-9369-4c2c21002028.png)

PR #10255, opened 2023-04-19
Update 01-login.md

decipher-cs:
Added a link to the Auth0 FAQ in the Quickstart Guide to provide users with additional information and troubleshooting tips, including what to do if the user is not logged in after a page refresh.


decipher-cs:
#10254

PR #10260, opened 2023-05-16
Update redirects.js

avanscoy:
Added a redirect for enable-sso-for-applications


avanscoy:
The docs for this PR need to be removed according to our internal process. I'm closing this PR so we can create a new one without issues.

PR #10263, opened 2023-05-25
Update 02-calling-an-api.md

nranderson:
These changes are required to work in the latest version of Angular.
* Add the '?' behind user to fix the ts(18049) error.
* Add the type and change to dot notation for the map() to fix the ts(7053) error.


frederikprijck (approved):
Thanks!

PR #10266, opened 2023-06-02
Update redirects.js

avanscoy:
Redirect from mistaken publication. The Nav didn't get published via Contentful so this page became an orphan. The orphan link is showing in Google search.


pavokta (approved):
Will not break anything.

pdillon (approved):

PR #10345, opened 2023-12-11
Added redirect to prepare archive of "Sanitize HTTP Traces" article

kegashe:
https://auth0team.atlassian.net/browse/DOCS-3542


avanscoy (approved):
Approved! Thanks so much.

PR #10350, opened 2023-12-21
Clarify port for spring boot login quickstart

jimmyjames:
Clarifies that the Spring Boot quickstart/sample uses port 3000 instead of the standard java port 8080.


frederikprijck (approved):

deepu105 (approved):
lgtm

PR #10365, opened 2024-02-02
DOCS-3574-Edick

rle28:
- Updated existing redirect for Tenant Hostname Migration
- Added 6 new Past Migration redirects


rle28:
Redirects will be re-added in a different PR

PR #10373, opened 2024-02-20
Update redirects.js

lrzhou25:
https://auth0team.atlassian.net/browse/DOCS-3696


avanscoy (approved):
Approved but with the caveat it might fail due to `/` so we might need to redirect to /articles instead

PR #10382, opened 2024-03-08
Added mTLS authentication as an authentication method

lrzhou25:


articles/api/authentication/_introduction.md
milan-khan, line 1: Can we say that mTLS Authentication is for confidential applications also? Since mTLS clients will need a way to protect their private key, so it's not really for public clients.

articles/api/authentication/_introduction.md
```diff
@@ -45,6 +46,24 @@ Send the Client ID. For public applications (applications that cannot hold crede
...
+
+Generate a certificate, either [self-signed](https://auth0.com/docs/get-started/applications/configure-mtls/configure-mtls-for-a-client#self-signed-certificates) or [certificate authority signed](https://auth0.com/docs/get-started/applications/configure-mtls/configure-mtls-for-a-client#certificate-authority-signed-certificates). Then, [set up the customer edge network](https://auth0.com/docs/get-started/applications/configure-mtls/set-up-the-customer-edge) that performs the mTLS handshake. 
+
+Once your edge network verifies the certificate, forward the request to the Auth0 edge network with the following headers:
+
+- The Custom Domain API key as the `cname-api-key` header.
+- The client certificate as the `client-certificate` header.
+- The client certificate CA verification status as the `client-certificate-ca-verified` header. For more information, see [Forward the Request](https://auth0.com/docs/get-started/applications/configure-mtls/set-up-the-customer-edge#forward-the-request-). 
+
+The following Authentication API endpoints support mTLS authentication for configured applications:
```
dlozlla: The below list is not exhaustive (some MFA endpoints are missing). Basically Daniel confirmed it is available to all authenticated endpoints in the API, so I think we don't need to say anything specific about this for mTLS either.

articles/api/authentication/_introduction.md
avanscoy: Can we also add client assertions signed with RSA key pairs with Private Key JWT?

lrzhou25: hi @avanscoy, i created a ticket for this [issue](https://auth0team.atlassian.net/browse/DOCS-3792?atlOrigin=eyJpIjoiOTc5NWQ3OGI5NDdiNGM5ZWIzMGY5ZGE5ZDZhYjVlOTUiLCJwIjoiamlyYS1zbGFjay1pbnQifQ). let's discuss more through slack. i can create a separate PR with those updates

kegashe (approved):

PR #10395, opened 2024-03-29
Added redirect to Manage Support Center Users page

lrzhou25:
Removed Manage Support Center Users page and added a redirect to Dashboard Access By Role. See https://auth0team.atlassian.net/browse/DOCS-3853.


meyerpdx (approved):
Looks good!

PR #10423, opened 2024-05-20
Update 01-login.md

Mehr-creates:


Mehr-creates (commented):
You can obtain a list of all the claims contained in the ID Token by iterating through the `Claims` collection:

PR #10429, opened 2024-06-03
Fix Maven dependency group ID

xiao2shiqi:
Updated the group ID for the Okta Spring Boot Starter dependency in the Maven POM file from 'com.okta' to 'com.okta.spring'.

This change corrects the group ID to ensure the dependency is resolved correctly.


frederikprijck (approved):
Thanks!

PR #10516, opened 2025-03-07
Updated non-interactive nextJS guide with SDK 4 instructions.

BcnCarlos:


tusharpandey13 (approved):
lgtm

frederikprijck (changes requested):
This shouldn't have been approved, as this is not correct.

The quickstart doesnt compile in its current state and still has content for v3.

PR #10518, opened 2025-03-13
updating python backend quickstart to use pyjwt instead of jose package

wdaimee:
This PR updates the python backend quickstart by replacing the deprecated python-jose package with the PyJWT package.


BcnCarlos (approved):
LGTM

